PR #198, opened 2022-11-17
refactor: pass vertices instead of transaction list to State Manager

dhedey:
Working with @malexandru-rdx 

This is still a work in progress :)

### DIFF
--- a/core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/PrepareRequest.java
+++ b/core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/PrepareRequest.java
@@ -69,7 +69,9 @@ import com.radixdlt.sbor.codec.CodecMap;
 import com.radixdlt.sbor.codec.StructCodec;
 import com.radixdlt.transactions.RawNotarizedTransaction;
 import com.radixdlt.utils.UInt64;
+import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
 
 public record PrepareRequest(
     byte[] parentAccumulator,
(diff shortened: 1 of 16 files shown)


core-rust/state-manager/src/jni/state_computer.rs
```diff
@@ -263,9 +264,10 @@ impl From<JavaCommitRequest> for CommitRequest {
     }
 }
 
-#[derive(Debug, Decode, Encode, Categorize)]
+#[derive(Debug, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
```
dhedey, line 267: Out of interest, why did you have to change this to `ScryptoX`? Is there something that is only in the Scrypto model embedded in these types?

core/src/main/java/com/radixdlt/ledger/StateComputerLedger.java
```diff
@@ -228,24 +226,56 @@ private Optional<ExecutedVertex> prepareInternal(
                 timeSupplier.currentTime()));
       }
 
-      final var executedTransactionsOptional =
-          this.verifier.verifyAndGetExtension(
+      // It's possibly that this function is called with a list of vertices which starts with some
```
dhedey, line 229: It's possible*

dhedey (commented):
Looks good, thanks @malexandru-rdx  - would appreciate a look over from @LukasGasior1  before merging 👍

core/src/main/java/com/radixdlt/ledger/StateComputerLedger.java
```diff
@@ -228,24 +226,56 @@ private Optional<ExecutedVertex> prepareInternal(
...
+        if (committedAccumulatorHasMatchedStartOfAVertex) {
+          verticesInExtension.add(previousVertex);
+        }
+      }
+
+      if (!committedAccumulatorHasMatchedStartOfAVertex) {
+        // This could trigger if the "previous" vertices don't line up with the committed state.
+        // In other words, they don't provide a valid partial path from the committed state to the
+        // start of the proposal.
+        return Optional.empty();
```
LukasGasior1, line 258: might be useful to add logs and/or metrics alongside Optional.empty() results

LukasGasior1 (approved):
just some renaming suggestions :)

core-rust/state-manager/src/jni/state_computer.rs
malexandru-rdx: `PreviousVertex` was using `ScryptoCategorize` and hence forced `PrepareRequest` to have the same.
It's kind of confusing what exactly each of the versions does so I just went with the flow :).

I've changed them to use `Decode`/`Encode`.

core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/PrepareRequest.java
```diff
@@ -67,13 +67,15 @@
...
 import com.radixdlt.transactions.RawNotarizedTransaction;
 import com.radixdlt.utils.UInt64;
+import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
 
 public record PrepareRequest(
-    List<RawLedgerTransaction> previous,
+    byte[] parentAccumulator,
+    List<PreviousVertex> previous,
```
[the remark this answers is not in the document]
malexandru-rdx: As of right now `prepare` function should support just a partial path up to root as long as there was a previous `prepare` that was up to the root from one of the vertices. I think it should work with no previous vertices as long as the proper `parentAccumulatorHash` is provided.

I would keep it as `previousVertices`.

core-rust/state-manager/src/jni/state_computer.rs
```diff
@@ -263,9 +264,10 @@ impl From<JavaCommitRequest> for CommitRequest {
...
 pub struct JavaPrepareRequest {
-    pub already_prepared: Vec<JavaRawTransaction>,
+    pub parent_accumulator: Vec<u8>,
+    pub prepared_vertices: Vec<JavaPreviousVertex>,
```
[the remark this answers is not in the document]
malexandru-rdx: Done

core-rust/state-manager/src/jni/mempool.rs
```diff
@@ -202,27 +202,27 @@ fn do_get_transactions_to_relay(
 // DTO Models + Mapping
 //
 
-/// Corresponds to the payload_hash
+/// Corresponds to HashCode from Java
 #[derive(Debug, PartialEq, Eq, Categorize, Encode, Decode)]
-pub struct JavaPayloadHash(Vec<u8>);
+pub struct JavaHashCode(pub Vec<u8>);
```
malexandru-rdx, line 207: Is this fine? In Java it's represented as `HashCode`, it feels better to have the same naming.

PR #222, opened 2022-12-14
radixdlt-scrypto lib update

talekhinezh:
Update `radixdlt-scrypto` lib to latest commit on `develop` branch


core/src/main/java/com/radixdlt/rev2/REv2StateComputer.java
```diff
@@ -158,7 +158,7 @@ public StateComputerLedger.StateComputerResult prepare(
             proposedTransactions,
             UInt64.fromNonNegativeLong(roundDetails.epoch()),
             UInt64.fromNonNegativeLong(roundDetails.roundNumber()),
-            UInt64.fromNonNegativeLong(roundDetails.proposerTimestampMs()));
+            roundDetails.proposerTimestampMs());
```
dhedey, line 161: Argh sorry I resolved this a different way on `betanet` / `main` - continued using a UInt64 on the boundary and casted inside the state manager.

I'm happy for us to resolve it either way around

core-rust/state-manager/src/transaction/validator_transaction.rs
```diff
@@ -64,35 +67,35 @@ impl ValidatorTransaction {
...
     hash: Hash,
-    manifest: TransactionManifest,
+    instructions: Vec<Instruction>,
+    blobs: Vec<Vec<u8>>,
 }
 
 impl PreparedValidatorTransaction {
     pub fn get_executable(&self) -> Executable {
-        let transaction_hash = Hash([0u8; Hash::LENGTH]);
+        let transaction_hash = Hash([0u8; Hash::LENGTH]); // TODO: Is this okay?
```
dhedey, line 86: Hmm shouldn't we just use `self.hash` here?

core-rust/state-manager/src/transaction/validation.rs
```diff
@@ -74,7 +76,13 @@ impl LedgerTransactionValidator {
         let validator = NotarizedTransactionValidator::new(self.validation_config);
         match prepared_transaction {
             PreparedLedgerTransaction::User(notarized_transaction) => {
-                validator.validate(notarized_transaction, &self.intent_hash_manager)
+                // TODO: Remove
```
dhedey, line 79: Can you elaborate - how should it be removed in your opinion? Does this need changes in `radixdlt-scrypto`?
Or do you intend for the payload size to be calculated further up the stack and passed in here? In which case, can we do it as part of this PR?

core-rust/state-manager/src/store/db.rs
```diff
@@ -130,7 +132,8 @@ impl StateManagerDatabase {
             debug!("Running genesis on the engine...");
             let mut db_txn = state_manager_db.create_db_transaction();
 
-            let genesis_receipt = bootstrap(&mut db_txn).expect("Genesis wasn't run");
+            let interpreter = ScryptoInterpreter::<DefaultWasmEngine>::default();
```
dhedey, line 135: This uses a different interepreter (and eg different metering config) compared with the state manager.

Now I'm not sure _why_ we overwrite the metering config - this should probably be changed to just be the default? In case genesis ends up deploying / running packages - which it might do to be fair?

We should probably create the ScryptoInterpreter in state manager init and then pass the same one into DB and State Manager set up?

core-rust/state-manager/src/state_manager.rs
```diff
@@ -562,7 +561,7 @@ where
 
             let validate_result = self
                 .user_transaction_validator
-                .validate_and_create_executable(&parsed);
+                .validate_and_create_executable(&parsed, proposed_payload.len());
```
dhedey, line 564: I'm very worried that it's not clear what this proposed payload length is and that we might end up being inconsistent in whether we use the notarized payload length or the ledger payload length(!!).

To mitigate this, could we perhaps have `validate_and_create_executable` only take some struct which is created by `UserTransactionValidator and ensures we use the same payload length for both?

Which should probably be the notarized payload length.

However, decoding the LedgerPayload and then encoding the Notarized Payload just to subtract a constant from the length feels a little silly - so maybe we can try to be a little more clever and calculate that constant, and have some tests which verify that constant is accurate?

dhedey (approved):
Generally good, and approving so you can merge after mark-ups, but a couple of things to flag:

* You'll need to resolve merge conflicts on whether we support i64 / negative timestamps in the Core API or State Manager interface - I'd suggest no (ie sticking to unsigned on these interfaces) is probably best, but up to you which way you'd like to resolve it.
* I'm uncomfortable with lots of TODOs around payload length - I view that as rather risky - eg a differential on what we interpret as the payload (ledger or notarized transaction) could come back to bite us with a liveness break... We're at a point now where I'd rather we remove TODOs than add them, so ideally I'd like to hook all these up properly on the node side.

core-rust/state-manager/src/transaction/validator_transaction.rs
talekhinezh: Yup, I think that should work.

core-rust/state-manager/src/transaction/validation.rs
talekhinezh: Yes, I think the fact that payload size is being passed around is a bit of a code smell suggesting that we have some weird abstractions. I'd prefer to keep this PR to just updating scrypto lib and working on another PR to fix the broken abstraction.

core-rust/state-manager/src/store/db.rs
talekhinezh: This is cleaned up in the next PR which focuses on cleaning up genesis

core-rust/state-manager/src/state_manager.rs
talekhinezh: Yup, totally agree here. As mentioned earlier though I think this will require a bit of a refactor and would prefer to have this PR focused on just integration. Will work on the refactor right after this though.

PR #235, opened 2022-12-20
Rust debug symbols

mozi-rdx:
Temporary pull request to trigger docker image build, so I can start running some test on ephemeral clusters.


mozi-rdx:
on develop: `/usr/lib/jni/libcorerust.so: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=046bb15bc65892fc7497bdac068c1e6836850ac6, with debug_info, not stripped`

PR #243, opened 2023-01-05
Fix BFT current round recovery so that it takes a timeout certificate into account

LukasGasior1:
Fixes current round recovery when the previous quorum formed a timeout certificate (rather than a typical QC).
The initial round is now `max(highestQc.round, highestTc.round)`. Prior to the fix, it could happen that the node recovered to an earlier round (latest QC) and, if it had already sent a vote for its previously latest round, was unable to send a vote (including a timeout vote) for its current round (because of the safety rules). This can lead to a liveness break if it happens to a few nodes at once.

This PR also includes a new metric that measures the time consensus events spend in a queue.

### DIFF
--- a/common/src/main/java/com/radixdlt/monitoring/Metrics.java
+++ b/common/src/main/java/com/radixdlt/monitoring/Metrics.java
@@ -163,6 +163,7 @@ public record Metrics(
       Counter noVotesSent,
       Counter voteQuorums,
       Counter timeoutQuorums,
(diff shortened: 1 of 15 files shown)


dhedey (approved):
Nice spot / fix!

As discussed, let's review all our uses of highQc.highestQc() to check we're not missing any other cases, in a follow-up PR.

dhedey (approved):
Nice test! 👍

PR #245, opened 2023-01-09
Introduce State's hash tree

jakrawcz-rdx:
This is a draft-only PR for demonstrating expected change surface after introducing State hash tree to Node.


core-rust/state-manager/src/state_manager.rs
```diff
@@ -178,6 +174,74 @@ impl<S> StateManager<S> {
...
+                }
+            },
+            TransactionResult::Reject(reject_result) => {
+                Err(format!("Validator txn rejected: {:?}", reject_result))
+            }
+        }
+    }
+
+    // DEV: question - should this, instead, become a new behavior of
+    // radix_engine::transaction::execute_and_commit_transaction() ?
```
malexandru-rdx, line 214: From the node perspective: `execute_and_commit_transaction` (I'd argue all `execute_*`) should not be used directly anyway, with some exceptions (i.e. `preview_call`) since you want to hit the execution cache first.
From the perspective of engine/scrypto: unless the hash root is actually used inside of it I don't see why it needs to know about it.

jakrawcz-rdx: I had an impression that `execute_and_commit_transaction()` is only a convenience method, while the "raw business" is the `execute_transaction()` itself.

For (unrelated) reasons, we want the JMT impl to live under the Scrypto repo (even though, technically, it could be fully contained within Node). And in this context, I assumed that the convenience `execute_and_commit_transaction()` could take the tree update responsibility into account as well - but that's just a thought. I'll probably only do it if it simplifies some code (of Node or of Scrypto or even of tests), I guess that's what "convenience method" means :)

PR #246, opened 2023-01-09
Cleanup HighQc/TimeoutCertificate use in the vertex store

LukasGasior1:


jakrawcz-rdx (commented):
The PR looks fine from a "coding" perspective 👍 but please don't treat my review as conclusive (I think some BFT expert still needs to see it; or I'd have to see a test actually demonstrating the change in behaviour)

core/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
```diff
@@ -437,12 +449,32 @@ private void rebuildAndSyncQC(SyncState syncState) {
       syncState.fetched.sort(Comparator.comparing(v -> v.vertex().getRound()));
       ImmutableList<VertexWithHash> nonRootVertices =
           syncState.fetched.stream().skip(1).collect(ImmutableList.toImmutableList());
+
+      final var syncStateHighestCommittedQc = syncState.highQC().highestCommittedQC();
+      final var syncStateHighestTc = syncState.highQC.highestTC();
+      final var currentHighestTc = vertexStore.highQC().highestTC();
+      final Optional<TimeoutCertificate> highestKnownTc;
+      /* highestKnownTc = max(syncStateHighestTc, currentHighestTc) but must handle Optionals */
```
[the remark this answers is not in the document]
LukasGasior1: nice, much cleaner :)

core/src/main/java/com/radixdlt/consensus/bft/VertexStoreAdapter.java
```diff
@@ -109,29 +109,26 @@ public boolean tryRebuild(VertexStoreState vertexStoreState) {
...
-            .onPresent(
-                committedUpdate ->
-                    this.bftCommittedDispatcher.dispatch(
-                        new BFTCommittedUpdate(
-                            committedUpdate.committedVertices(), inserted.vertexStoreState())));
-        yield true;
+  public VertexStore.InsertQcResult insertQc(QuorumCertificate qc) {
+    final var result = vertexStore.insertQc(qc);
+    if (result instanceof VertexStore.InsertQcResult.Inserted inserted) {
+      // TODO: why is this if statement needed?
```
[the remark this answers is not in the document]
dhedey: I feel like dispatching a HighQC event makes sense, although I'm surprised we don't just always do it?

Like why if we get a commit do we not need to also replace HighQC? Or why does a vertex missing mean we don't still store HighQC?

:')

core/src/main/java/com/radixdlt/consensus/bft/VertexStoreAdapter.java
```diff
@@ -109,29 +109,26 @@ public boolean tryRebuild(VertexStoreState vertexStoreState) {
     return result.isPresent();
   }
 
-  public void insertTimeoutCertificate(TimeoutCertificate timeoutCertificate) {
-    vertexStore.insertTimeoutCertificate(timeoutCertificate);
+  public boolean insertTimeoutCertificate(TimeoutCertificate timeoutCertificate) {
```
dhedey, line 112: Can we copy the java doc to the adaptor to explain what the boolean means?

By "ignored" I guess it means "ignored because it's not the highest"?

core/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
```diff
@@ -215,13 +215,13 @@ public EventProcessor<RoundQuorumReached> roundQuorumReachedEventProcessor() {
       final var highQC =
           switch (roundQuorumReached.votingResult()) {
             case FormedQC formedQc -> HighQC.from(
```
dhedey, line 217: Does it make sense for HighQC to have a `withHigherQC` / `withHigherTC` method on it which creates a new high qc?

core/src/main/java/com/radixdlt/consensus/bft/VertexStoreJavaImpl.java
```diff
@@ -183,6 +183,7 @@ public Option<VertexStoreState> tryRebuild(VertexStoreState vertexStoreState) {
     this.rootVertex = vertexStoreState.getRoot();
     this.highestCommittedQC = vertexStoreState.getHighQC().highestCommittedQC();
     this.highestQC = vertexStoreState.getHighQC().highestQC();
+    this.highestTC = vertexStoreState.getHighQC().highestTC();
```
dhedey, line 186: Does it make sense to just store `highQC` on this class? instead of these separate values?

core/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
```diff
@@ -442,12 +455,20 @@ private void rebuildAndSyncQC(SyncState syncState) {
...
+      final var currentHighestTc = vertexStore.highQC().highestTC();
+      final var highestKnownTc =
+          Stream.of(currentHighestTc, syncStateHighestTc)
+              .flatMap(Optional::stream)
+              .max(Comparator.comparing(TimeoutCertificate::getRound));
+
       var vertexStoreState =
           VertexStoreState.create(
-              HighQC.from(syncState.highQC().highestCommittedQC()),
+              HighQC.from(syncStateHighestCommittedQc, syncStateHighestCommittedQc, highestKnownTc),
```
dhedey, line 469: Oh dang. Nice catch here.

TBH this method `HighQC.from(qc)` feels wrong - what do you think about removing this constructor for `HighQC` and make people be explicit about committed / timeout?

core/src/main/java/com/radixdlt/rev1/modules/REv1LedgerRecoveryModule.java
```diff
@@ -153,7 +153,7 @@ private static VertexStoreState genesisEpochProofToGenesisVertexStore(
             lastEpochProof.consensusParentRoundTimestamp(),
             lastEpochProof.proposerTimestamp());
     var genesisQC = QuorumCertificate.createInitialEpochQC(genesisVertex, nextLedgerHeader);
-    return VertexStoreState.create(HighQC.from(genesisQC), genesisVertex, Optional.empty(), hasher);
+    return VertexStoreState.create(HighQC.from(genesisQC), genesisVertex, hasher);
```
dhedey, line 156: Slightly left over from Josh's PR - could we rename `genesisQC` in these places as `initialEpochQC`? (and similarly grep other uses of genesis and change those?)

core/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
dhedey: Happy to have an explicit method `HighQC.ofInitialEpochQc` or something though

dhedey (approved):
Really nice work, thanks.

A few suggestions for some further renames to clarify things.

Separately - what do you think about renaming `HighQC` to be eg `HighCerts` or something? As confusingly it's not just a single QC?

core/src/main/java/com/radixdlt/consensus/bft/VertexStoreAdapter.java
LukasGasior1: I'd have to copy all the docs, for consistency. Not sure if it makes sense? It's quite clear that the adapter forwards it to the main impl.

Extended the comment with ignore reason :+1:

core/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
LukasGasior1: how about `withHighestTC` and `withHighestQC`, without any additional assumptions whether the qc/tc is indeed higher than the current?

dhedey: Sure, sounds good!

Perhaps withPossiblyHigherTC / withPossiblyHigherQC might be even more useful, but I'll let you decide 👍 :)

core/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
LukasGasior1: :+1: renamed to `ofInitialEpochQc`

core/src/main/java/com/radixdlt/consensus/bft/VertexStoreAdapter.java
jakrawcz-rdx: > It's quite clear that the adapter forwards it to the main impl.

@LukasGasior1 I now had a quick look at this entire "adapter" and it does not do much adapting TBH.
OFC I have no idea on the motivation (or dev history of this class), but ATM it looks more like a decorator - namely, an `EventDispatchingVertexStore`, doesn't it?

Such decorator would just need a class-level javadoc (e.g. "A decorator dispatching events according to the results of the underlying `VertexStore`").

Then, the methods' javadocs (which David wants) would be auto-propagated from the interface.
(of course we would first have to move them from the `VertexStoreJavaImpl` to the `VertexStore` interface, which is the right move anyway)

(the only odd, non-decorator-ish method left would be `hasCommittedVertexOrRootAtOrAboveRound()`, which looks more like a convenience method, which is easiest to maintain as a `default` within the interface itself).

I find this mechanical change worthwhile, but easily delegated to another PR.

core/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
LukasGasior1: (and added a genesis check)

core/src/main/java/com/radixdlt/consensus/bft/VertexStoreAdapter.java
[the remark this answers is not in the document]
LukasGasior1: I was surprised too, that's why I've added this comment :smile:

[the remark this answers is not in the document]
LukasGasior1: it might just be some tests that wrongly rely on receiving only one of those events

core/src/main/java/com/radixdlt/consensus/bft/VertexStoreAdapter.java
LukasGasior1: yup, sounds good. The original reason to introduce the adapter was the handle the java-side events once vertex store is in rust (which hasn't happened yet)

LukasGasior1: > Then, the methods' javadocs (which David wants) would be auto-propagated from the interface.

Not sure how that works? The Adapter itself doesn't implement VertexStore

jakrawcz-rdx: > The Adapter itself doesn't implement VertexStore

Yeah, and in https://github.com/radixdlt/babylon-node/pull/246#discussion_r1069367993 I postulate that it should, since it is more of an "event-dispatching VertexStore decorator" than adapter.

Since my Java is less ambiguous than my English, I imagined something like this:
```
public class EventDispatchingVertexStore implements VertexStore {
  private final VertexStore underlying;

  @Override
  public boolean tryRebuild(VertexStoreState vertexStoreState) {
    var result = this.underlying.tryRebuild();
    this.bftRebuildDispatcher.dispatch(whatever);
    return result;
  }
}
```

I agree that at this point the only thing it gives us is "contract becomes more obvious + javadoc gets propagated + less apis exist". Functionally it isn't a "must-have", just a minor refactor that could happen.

LukasGasior1: Ah, gotcha. But currently VertexStoreAdapter's interface is slightly different. VertexStore was modified to make Java->Rust transition easier, while VertexStoreAdapter remained unchanged (i.e. it has the original VertexStore interface) to support existing clients. TBH, I wouldn't spend much time refining it in the current form, given its uncertain future.

core/src/main/java/com/radixdlt/consensus/bft/VertexStoreAdapter.java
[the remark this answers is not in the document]
dhedey: Might be worth us trying to remove the if statement and see if any tests fail?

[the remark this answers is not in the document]
LukasGasior1: yup, I've already gave it a whirl in a separate PR

core/src/test/java/com/radixdlt/recovery/REv2LedgerRecoveryTest.java
```diff
@@ -99,7 +99,7 @@ public static Collection<Object[]> parameters() {
         new Object[] {90},
         new Object[] {100},
         new Object[] {500},
-        new Object[] {100});
+        new Object[] {1000});
```
LukasGasior1, line 102: I believe 1000 was meant here (100 was a duplicate)

core/src/test/java/com/radixdlt/recovery/REv2LedgerRecoveryTest.java
```diff
@@ -123,7 +123,7 @@ private DeterministicTest createTest() {
                     StateComputerConfig.rev2(
                         Network.INTEGRATIONTESTNET.getId(),
                         TransactionBuilder.createGenesisWithNumValidators(
-                            1, UInt64.fromNonNegativeLong(10)),
+                            1, UInt64.fromNonNegativeLong(Long.MAX_VALUE)),
```
LukasGasior1, line 126: There are still epoch changes in rev2 state computer, even though this test uses a "no epochs" node module, which messes up safety store recovery (`NoEpochsSafetyRecoveryModule` is stripped of the epochs logic that `EpochsSafetyRecoveryModule` has). Not sure if this is the cleanest solution, but seems reasonable(?)

core/src/main/java/com/radixdlt/rev2/REv2StateComputer.java
```diff
@@ -225,59 +225,67 @@ public StateComputerLedger.StateComputerResult prepare(
...
+              proof.getAccumulatorState(),
+              proof.consensusParentRoundTimestamp(),
+              proof.proposerTimestamp());
+      final var initialEpochQC =
+          QuorumCertificate.createInitialEpochQC(initialEpochVertex, nextLedgerHeader);
+
+      // Post commit vertex store is a fresh one
+      postCommitVertexStoreState =
+          VertexStoreState.create(
+              HighQC.ofInitialEpochQc(initialEpochQC), initialEpochVertex, hasher);
```
dhedey, line 251: I appreciate this code is just mostly a move, but it might be clearer if we can encapsulate this logic into a new function, eg `VertexStoreState.createNewForNextEpoch(nextEpoch)` ?

core/src/test/java/com/radixdlt/recovery/REv2LedgerRecoveryTest.java
dhedey: Haha good spot.

Hmm tbh I'm not sure if we even need `NoEpoch` tests?

Like it might be nice if we could just make no epoch = long epochs? Would certainly decrease the number of supported combinations of the system and make it less likely we overlook something.

What do you think @LukasGasior1 ? @talekhinezh ?

dhedey (approved):
Thanks - I like clarifying that the vertex store state is pre/post commit

core/src/main/java/com/radixdlt/rev2/REv2StateComputer.java
LukasGasior1: :+1: done

PR #248, opened 2023-01-11
Add timer metrics to mempool/ledger

LukasGasior1:


malexandru-rdx (approved):

core-rust-bridge/src/main/java/com/radixdlt/mempool/RustMempool.java
```diff
@@ -80,7 +82,10 @@
 
 public class RustMempool implements MempoolReader<RawNotarizedTransaction> {
 
-  public RustMempool(StateManager stateManager) {
+  private final Metrics metrics;
+
+  public RustMempool(Metrics metrics, StateManager stateManager) {
+    this.metrics = Objects.requireNonNull(metrics);
```
jakrawcz-rdx, line 88: (minor; barely related) I think that here (and in many other applicable places) we can drop the null-checks of Guice-injected services.
As explained in https://github.com/google/guice/wiki/UseNullable#guice-forbids-null-by-default we will never actually get `null`, unless you annotate the injection point with `@Nullable` (in which case you don't want the null-check anyways!).

core/src/main/java/com/radixdlt/ledger/StateComputerLedger.java
```diff
@@ -280,12 +294,20 @@ public EventProcessor<BFTCommittedUpdate> bftCommittedUpdateEventProcessor() {
...
+      this.commitWithTimer(transactionsWithProof, committedUpdate.vertexStoreState());
     };
   }
 
   public EventProcessor<CommittedTransactionsWithProof> syncEventProcessor() {
-    return p -> this.commit(p, null);
+    return p -> this.commitWithTimer(p, null);
+  }
+
+  private void commitWithTimer(
```
jakrawcz-rdx, line 305: The `prepareWithTimer()` and `commitWithTimer()` methods start to look like a recurring pattern.
What do you think about adding a convenience `.measure(Runnable/Callable<T>)` method to our `Timer` class, instead?
(the impl will simply forward to the existing underlying `Summary.time(Runnable/Callable<T>)`)

jakrawcz-rdx (approved):
Very cool, please just see some minor improvement ideas 👍

dhedey (commented):
I like Jakub's suggestions.

Actually this strongly relates to discussions with @mozi-rdx on timing our lock contention...

I feel like it would be nice to have a more comprehensive monitoring of calls into state manager or time waiting for locks.

We could either:
* (A) basically add this in the Java layer around all of our calls (it could even be a part of `NativeCalls.Func1` etc taking a metric reference if we wanted to force the point)
* (B) be in the rust layer when we take locks

Do you have any thoughts on this @LukasGasior1 ?

Possibly (A) is easier to implement and is not blocked on a refactor where the metrics need to be taken outside of the state manager lock (for multiple reasons) - but I'm happy for either (A) or (B)

core-rust-bridge/src/main/java/com/radixdlt/mempool/RustMempool.java
LukasGasior1: Sounds good. I've removed the check from this PR, but we can leave cleaning up the remaining classes to a separate PR, I guess? Note that there are some that can be either created by guice or via a constructor (e.g. SafetyRules), where we'd probably want to keep the validation.

PR #255, opened 2023-01-16
Feature: Core API - January tweaks 1

dhedey:
The main external change is vastly improved Core API clarity for rejections (status and submit).

In particular:
* The `/transaction/status` endpoint can now mark an intent as "PermanentlyRejected" if at least one of its payload rejections includes a permanent rejection reason applicable to the intent itself
* The `/transaction/status` endpoint includes a text-based description of the intent status returned, to help users understand what the returned status means.
* The `/transaction/submit` endpoint now returns a structured error message, including details in some cases. In particular, in the case of a transaction rejection, lots of information about the rejection is returned, such as whether the rejection is permanent for the intent, and/or the payload; whether the returned status is calculated fresh or cached, and when the node will consider calculating/executing it again. The latter might well change in the future.

It also includes a few refactors to enable `details` to be added onto error responses.

Further changes as part of this PR include:
* Make use of the inheritance-based "oneOf" discriminator tooling for all our discriminated unions, instead of anyOf/oneOf - https://redocly.com/docs/resources/discriminator/ - this isn't even discussed in the Open API documentation, but appears to be supported well in the generators.
* Return the ISO datetime as well as the unix timestamp for instants


core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -926,12 +926,27 @@ components:
...
       maxLength: 64
       description: The hex-encoded ledger-wrapped transaction hash. This is known as the Ledger Hash. This hash is `SHA256(SHA256(ledger_transaction_bytes))`
+    ErrorResponseType:
+      type: string
+      enum:
+        - Basic
+        - TransactionSubmit
     ErrorResponse:
       type: object
+      discriminator:
```
KrzysztofLabus-RDX, line 936: OK, there is some problem here. For reasons unknown to me yet C# OAS generator generates `public BasicErrorResponse(ErrorResponseType errorType = "BasicErrorResponse", int code = ...`, i.e. it tries to use string rather enum.

KrzysztofLabus-RDX (changes requested):

core-rust/core-api-server/core-api-schema.yaml
KrzysztofLabus-RDX: OK, I think I know what the problem is. Current model tries to effectively serialize error responses to:
```
{"error_type":"Basic", "code":123, "message":"bleh"}
{"error_type":"TransactionSubmit", "code":321, "message":"bleh bleh", "some_extra_prop":321}
```
While this is clearly better and more logical structure I'd suggest to go back to what we have with Olympia, i.e.
```
{"code":123, "message":"bleh", "details":{"error_type":"Basic"}}
{"code":321, "message":"bleh bleh", "details": {"error_type":"TransactionSubmit", "some_extra_prop":321}}
```

At least this will work with OAS generator.

core-rust/core-api-server/src/core_api/handlers/transaction_submit.rs
```diff
@@ -22,13 +26,41 @@ pub(crate) async fn handle_transaction_submit(
...
+            StatusCode::BAD_REQUEST,
+            "Transaction was rejected",
+            TransactionSubmitErrorDetails::TransactionSubmitRejectedErrorDetails {
+                error_message: format!("{}", rejection.reason),
+                is_fresh: !rejection.was_cached,
+                is_payload_rejection_permanent: rejection.is_permanent_for_payload(),
+                is_intent_rejection_permanent: rejection.is_permanent_for_intent(),
+                is_rejected_because_intent_already_committed: rejection
+                    .is_rejected_because_intent_already_committed(),
+                recalculation_due: match rejection.recalculation_due {
```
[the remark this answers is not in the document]
dhedey: Fixed (or, well, I return an `Instant` which includes both the unix timestamp and the date time)

core-rust/core-api-server/core-api-schema.yaml
dhedey: Have agreed to proceed with the inheritance model - and will fix up the bugs in the C# open api generator on the C# end.

PR #256, opened 2023-01-16
Architecture: Remove use of BFTNode from P2P Layer

talekhinezh:
* Use `NodeId` instead of `BFTNode` for most P2P interactions
* This in preparation for the upcoming `Validator Components` PR which will need a separate abstraction between `BFTNode` and `NodeId`


core/src/main/java/com/radixdlt/messaging/MessagingModule.java
```diff
@@ -237,39 +262,54 @@ RxRemoteEnvironment rxRemoteEnvironment(
...
-          return messageCentralLedgerSync.ledgerStatusUpdates().map(m -> (RemoteEvent<T>) m);
-        } else if (remoteEventClass == Ping.class) {
-          return messageCentralPeerLiveness.pings().map(m -> (RemoteEvent<T>) m);
-        } else if (remoteEventClass == Pong.class) {
-          return messageCentralPeerLiveness.pongs().map(m -> (RemoteEvent<T>) m);
-        } else if (remoteEventClass == GetPeers.class) {
-          return messageCentralPeerDiscovery.getPeersEvents().map(m -> (RemoteEvent<T>) m);
-        } else if (remoteEventClass == PeersResponse.class) {
-          return messageCentralPeerDiscovery.peersResponses().map(m -> (RemoteEvent<T>) m);
+      public <N, T> Flowable<RemoteEvent<N, T>> remoteEvents(
```
dhedey, line 265: I'm concerned about this if statement. I may be misunderstanding when this is called, but I think this means that every message we receive, we are allocating 20 MessageTransportType classes?

I imagine we'd want to neaten this up?

core/src/main/java/com/radixdlt/messaging/MessagingModule.java
```diff
@@ -126,102 +129,124 @@ protected void configure() {
...
+        messageCentralLedgerSync.ledgerStatusUpdateDispatcher());
   }
 
   @ProvidesIntoSet
-  private RxRemoteDispatcher<?> ledgerStatusUpdateDispatcher(
+  private RxRemoteDispatcher<?, ?> bftLedgerStatusUpdateDispatcher(
       MessageCentralLedgerSync messageCentralLedgerSync) {
     return RxRemoteDispatcher.create(
-        LedgerStatusUpdate.class, messageCentralLedgerSync.ledgerStatusUpdateDispatcher());
+        BFTValidatorId.class,
```
dhedey, line 244: This appears to be the only place we register with `N = BFTValidatorId`, but it appears to duplicate the above which has `N = NodeId`. So it looks like each remote message can be registered with both?

I'm really not sure I understand the model here, maybe worth a discussion on slack?

core/src/main/java/com/radixdlt/environment/RemoteEventProcessor.java
```diff
@@ -64,18 +64,17 @@
...
 import com.radixdlt.environment.rx.RemoteEvent;
 
 /**
  * Processes events from a remote node
  *
  * @param <T> the event class
  */
-public interface RemoteEventProcessor<T> {
-  void process(BFTNode sender, T t);
+public interface RemoteEventProcessor<N, T> {
```
dhedey, line 74: I'd really like us to explain what `N` is here and on the MessageTransportType interface, and explain what it means for the messages and their receivers.

dhedey (commented):
A few worries, and more higher-level questions - which I've shared on slack. Happy to approve once these are resolved.

dhedey (approved):

PR #257, opened 2023-01-17
fix(build): use ubuntu:22.04 for builder base image

mozi-rdx:
There was a linking issue when babylon-node was built on a Debian-based
image, since the OpenSSL was pretty old there, and was not found on a
Ubuntu-based image. (v1.1.1 vs v3.0.0)

### DIFF
--- a/docker/rust-builder/Dockerfile
+++ b/docker/rust-builder/Dockerfile
@@ -1,20 +1,18 @@
-FROM rust:1.65.0-slim as base
+FROM ubuntu:22.04 as base
 WORKDIR /app
 
-RUN apt update; apt upgrade -y && \
+RUN apt-get update; apt-get upgrade -y && \
   apt install -y \
   build-essential \
   curl \
   g++-aarch64-linux-gnu \
   g++-x86-64-linux-gnu \
   libc6-dev-arm64-cross \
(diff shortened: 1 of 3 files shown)


docker/rust-builder/Dockerfile
```diff
@@ -1,20 +1,18 @@
...
-RUN apt update; apt upgrade -y && \
+RUN apt-get update; apt-get upgrade -y && \
   apt install -y \
   build-essential \
   curl \
   g++-aarch64-linux-gnu \
   g++-x86-64-linux-gnu \
   libc6-dev-arm64-cross \
-  libclang-dev
+  libclang-dev \
```
shambupujar, line 11: Wondering do we need to pin these versions?

mozi-rdx: I wouldn't pin the package versions at this point (in my experience, nobody (including me) cared enough to keep that tidy in a long run), I would rather create a new lifecycle for the builder/runtime images (i.e. moving them to a separate repository and CI), and occasionally tag and snapshot those images.

KrzysztofLabus-RDX (approved):

shambupujar (approved):

PR #259, opened 2023-01-18
Use RocksDB column families; some overall improvements

LukasGasior1:
Some notable changes:
* using RocksDB column families instead of custom prefixes
* txns are now primarily indexed by state version, not a hash
* txns, receipts and identifiers (accumulator hash) are stored separately: not a particularly strong opinion for this one, but I think it's worth it (to be able to serve sync responses without the need to scrypto_decode each individual transaction)

### DIFF
--- a/core-rust/state-manager/src/store/rocks_db.rs
+++ b/core-rust/state-manager/src/store/rocks_db.rs
@@ -63,12 +63,13 @@
  */
 
 use crate::types::UserPayloadHash;
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
+use std::fmt;
 
 use crate::store::traits::*;
 use crate::{
-    CommittedTransactionIdentifiers, HasIntentHash, IntentHash, LedgerPayloadHash,
-    LedgerTransactionReceipt,
+    AccumulatorHash, CommittedTransactionIdentifiers, HasIntentHash, HasLedgerPayloadHash,
+    HasUserPayloadHash, IntentHash, LedgerPayloadHash, LedgerTransactionReceipt,
 };
 use radix_engine::ledger::{
     OutputValue, QueryableSubstateStore, ReadableSubstateStore, WriteableSubstateStore,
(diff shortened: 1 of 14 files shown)


core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -78,217 +79,286 @@ use radix_engine::types::{
...
-        // it indicates a bug.
-        let transaction_already_exists = self
-            .db_txn
-            .get(transaction_key)
-            .expect("RocksDB: failure to check for transaction presence")
-            .is_some();
-        if transaction_already_exists {
-            panic!("Attempt to persist a duplicate transaction payload - which isn't supported at the moment in our database structure: {:?}", transaction)
-        }
+
```
LukasGasior1, line 235: Not sure if we still want to check for ledger payload hash duplicates? I haven't added it yet in this PR as it seems unnecessary (user txns are covered by the intent hash check and there should be no more duplicates in validator/system txns [unless there's a bug, but I'm not convinced that keeping an extra query here is worth it])

dhedey (approved):
Really nice work, thanks Lukasz!

Just to understand - when you say "read all via ledger sync" took 400ms (ish), was that all 100 batches of 1000 transactions(?), so about 4ms per batch?

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -78,217 +79,286 @@ use radix_engine::types::{
...
 impl RocksDBStore {
     pub fn new(root: PathBuf) -> RocksDBStore {
-        let db = TransactionDB::open_default(root.as_path()).unwrap();
+        let mut db_opts = Options::default();
+        db_opts.create_if_missing(true);
+        db_opts.create_missing_column_families(true);
+
+        let transactional_db_opts = TransactionDBOptions::default();
+
+        let column_families: Vec<ColumnFamilyDescriptor> = ALL_COLUMN_FAMILIES
```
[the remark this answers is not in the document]
LukasGasior1: correct

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -78,217 +79,286 @@ use radix_engine::types::{
...
-            .put(
-                db_key!(Substates, &scrypto_encode(&substate_id).unwrap()),
-                scrypto_encode(&substate).unwrap(),
-            )
-            .expect("RockDB: put_substate unexpected error");
+        let column_families = ALL_COLUMN_FAMILIES
+            .iter()
+            .map(|cf| (cf.clone(), self.cf_handle(cf)))
+            .collect();
+        RocksDBCommitTransaction {
```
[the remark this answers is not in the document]
LukasGasior1: Yes, I still want to give it a try, in a follow up PR (if either option turns out worthwhile).

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -78,217 +79,286 @@ use radix_engine::types::{
...
+            .2
+            .state_version;
+        let max_state_version_in_db = self.max_state_version();
+        if first_txn_state_version != max_state_version_in_db + 1 {
+            panic!("Attempted to commit a txn batch that starts with state version {} but the latest state version in DB is {}",
+                first_txn_state_version, max_state_version_in_db);
+        }
+
+        for committed_txn_bundle in committed_transaction_bundles {
+            self.insert_transaction(committed_txn_bundle);
```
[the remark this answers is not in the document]
LukasGasior1: hmh, I don't think there'll be huge benefit (if any), but I can still give it a try

LukasGasior1:
> Just to understand - when you say "read all via ledger sync" took 400ms (ish), was that all 50 batches of 1000 transactions(?), so about 8ms per batch?

it's a total time it took to read the whole ledger, but in different batches (1000 txns * 50 was just on the commit side).

* core API was called with a limit of 10000 txns (sent 7 requests [last resp empty])
* ledger sync used a 100Kb response limit (made 144 calls)

dhedey:
Gotchya. Do you have a feeling for what a sensible size of `MAX_TXN_BYTES_FOR_A_SINGLE_RESPONSE` is?

It'll certainly need to be bigger than a block size I guess.

core-rust/state-manager/src/store/rocks_db.rs
LukasGasior1: > Why do we actually need an index on user_payload_hash or ledger_payload_hash?

Hah, good question. I guess they're both not really needed, but might be useful to have them around for e.g. debugging?

> Is "no duplicates for ledger payload hash" an invariant we actually need?

Another good question :thinking: intuition tells me "yes", but I can't really come up with an example to justify it. I don't think we rely on this invariant anywhere in the node

dhedey: If it's easy to re-run the benchmark, I'd be curious: if we remove maintaining these indices (ie remove insertions on create for now), do we see a noticeable commit speedup?

LukasGasior1: There's almost no difference

LukasGasior1: I guess it makes sense then to keep the index and add an extra ledger payload duplicate check to ensure correctness

LukasGasior1:
A value that matches `OVERRIDE_MAX_PAYLOAD_SIZE` and `MAX_PACKET_LENGTH` seems reasonable. I've only changed it to run the test. Not sure about `MAX_TXNS_FOR_A_SINGLE_RESPONSE` anymore though. I guess this limit should only apply after the first proof (i.e. if there are more txns [but still within the bytes size limit] under the first proof, then it's OK)

PR #260, opened 2023-01-19
add execution cache

malexandru-rdx:
Update scrypto-lib and implement execution cache

## Description

Please see previous PRs/discussions: [#203](https://github.com/radixdlt/babylon-node/pull/203) & [#231](https://github.com/radixdlt/babylon-node/pull/231)


core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -2415,6 +2438,21 @@ components:
           type: string
           description: |
             The hex-encoded raw bytes of the mutadata data of the NF.
+    ValidatorSubstate:
+      allOf:
+        - $ref: '#/components/schemas/SubstateBase'
```
dhedey, line 2443: We have changed how we do disciminators in Core API after a PR last week - and got rid of the "Base" classes - this should instead reference Substate here

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -2125,6 +2127,14 @@ components:
         - KeyValueStoreEntry
         - NonFungibleStoreEntry
         - Vault
+        - Validator
+    SubstateBase:
```
dhedey, line 2131: This shouldn't exist any more :)

core-rust/core-api-server/src/core_api/conversions/receipt.rs
```diff
@@ -132,19 +136,32 @@ pub fn to_api_deleted_substate(
     })
 }
 
-#[tracing::instrument(skip_all)]
+// #[tracing::instrument(skip_all)]
```
dhedey, line 139: These comments here and below should probably be removed?

core-rust/core-api-server/src/core_api/conversions/substate.rs
```diff
@@ -690,37 +694,23 @@ pub fn to_api_package_royalty_accumulator_substate(
...
+        round,
+        epoch,
     } = substate;
 
     Ok(models::Substate::EpochManagerSubstate {
+        address: bech32_encoder.encode_component_address_to_string(address),
         epoch: to_api_epoch(*epoch)?,
-        round: to_api_round(*round)?,
-        rounds_per_epoch: to_api_round(*rounds_per_epoch)?,
+        round: (*round) as i64,
```
dhedey, line 712: Why are we not using `to_api_round` here and below?

core-rust/core-api-server/src/core_api/conversions/substate.rs
```diff
@@ -891,3 +884,42 @@ fn to_api_data_struct(
...
+    } = validator_set;
+
+    Ok(models::Substate::ValidatorSetSubstate {
+        validator_set: validator_set
+            .iter()
+            .map(|validator| {
+                to_api_active_validator(bech32_encoder, validator.address, validator.key)
+            })
+            .collect::<Result<_, _>>()?,
+        epoch: (*epoch) as i64,
```
dhedey, line 923: This should use `to_api_epoch`

core-rust/state-manager/src/state_manager.rs
```diff
@@ -209,13 +217,52 @@ where
...
+        payload_hash: &LedgerPayloadHash,
+    ) -> (AccumulatorHash, &TransactionReceipt) {
+        let new_accumulator_hash = parent_accumulator_hash.accumulate(payload_hash);
+
+        match self.execution_cache.get(&new_accumulator_hash) {
+            None => {
+                let parent_key = self
+                    .execution_cache
+                    .get(parent_accumulator_hash)
+                    .expect("Parent AccumulatorHash not found in cache. This should not happen!");
```
dhedey, line 241: When could this happen? I guess if the node tries to prepare a vertex without giving the transactions in the path from the committed state?

dhedey: Could we be a little clearer in the error message about why we believe this could never happen? (eg because of specific logic in StateComputerLedger.java or something?)

dhedey (commented):
Nice work. A few small tweaks and then it'll be ready to merge 👍

PR #263, opened 2023-01-22
Refactor: Use mapping context abstraction for Core API

dhedey:
Increases consistency in the Core API mappers, and allows for adding further tweaks to API response formats (eg different SBOR response types) with ease


malexandru-rdx:
Nice refactor.
Wouldn't it better with only 1 type of `Context` and maybe construct only one under `lazy_static` since it's passed as immutable ref everywhere? Otherwise, if only 2 types of contexts are kept (encoder/decoder) I think it's fine too.

dhedey:
> Nice refactor. Wouldn't it better with only 1 type of Context and maybe construct only one under lazy_static since it's passed as immutable ref everywhere? Otherwise, if only 2 types of contexts are kept (encoder/decoder) I think it's fine too.

In its current form, yes. But eventually this context will include parameters selected by the client. (eg "what form of SBOR do I want?"). Perhaps we could have a lazy static for the bech32encoder/decoder, but (assuming we need a state manager lock) it's very cheap to create, so I'm not sure it'd add much 👍. We can definitely revisit this in future though if we can use it to avoid state manager locks.

malexandru-rdx (approved):

PR #266, opened 2023-01-24
Write state manager commit changes in a single (atomic) batch, instead of using DB transaction

LukasGasior1:
Benchmark results in comparison to feature/rocksdb-improvements (https://github.com/radixdlt/babylon-node/pull/259)

```
with DB txn:
1000 {batches} x 50 {txns}: [171289, 166697, 168775, 168988], avg 168937
4000 {batches} x 3 {txns} : [43446, 45996, 48543, 42852], avg 45209

without DB txn (writeBatch):
1000 {batches} x 50 {txns}: [149876, 149950, 147244, 156403], avg 150868; ~10% faster
4000 {batches} x 3 {txns} : [34025, 42932, 41174, 40068], avg 39549; ~12% faster

```


core-rust/state-manager/src/state_manager.rs
```diff
@@ -805,12 +805,20 @@ where
             );
         }
 
-        let mut db_transaction = self.store.create_db_transaction();
+        // Create a transient substate store for all txns in the commit batch,
+        // backed by a StagedSubstateStore.
+        let mut commit_substate_store = CommitTransientSubstateStore::new(&mut self.store);
```
malexandru-rdx, line 810: With introduction of execution cache you will not be able to get a hold of `&mut self.store`. In that case you would have to construct `substates` from the retrieved `TransactionReceipt`s afterwards.

LukasGasior1: Thanks for a heads-up! It shouldn't be a problem

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -150,232 +147,148 @@ impl RocksDBStore {
...
             .collect();
 
-        let db = TransactionDB::open_cf_descriptors(
-            &db_opts,
-            &transactional_db_opts,
-            root.as_path(),
-            column_families,
-        )
-        .unwrap();
+        let db = DB::open_cf_descriptors(&db_opts, root.as_path(), column_families).unwrap();
```
dhedey, line 155: If we want to change back to a transaction DB, I assume it's something we can change at runtime with the same database files? Or are the files themselves different?

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -150,232 +147,148 @@ impl RocksDBStore {
...
-                    notarized_transaction.intent_hash(),
-                    true,
-                )
-                .expect("RocksDB: failure to read intent hash");
-            if let Some(existing_intent_hash_mapping) = existing_intent_hash_mapping_opt {
+            /* For user transactions we only need to check for duplicate intent hashes to know
+            that user payload hash and ledger payload hash are also unique. */
+            let intent_hash = notarized_transaction.intent_hash();
+
+            let maybe_existing_intent_hash = self
```
dhedey, line 175: This check has been watered down now and doesn't check for a collision of intent hashes inside the batch itself.

Maybe we can do this in the commit function, by storing a set of intent hashes as we go? And checking the size matches the batch size at the end?

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -150,232 +147,148 @@ impl RocksDBStore {
...
-                .put_cf(
-                    self.cf_handle(&StateVersionByTxnUserPayloadHash),
-                    notarized_transaction.user_payload_hash(),
-                    state_version.to_be_bytes(),
-                )
-                .expect("RocksDB: failure to put user payload hash");
+            if let Some(state_version) = maybe_existing_ledger_payload_hash {
+                warn!("Duplicate transaction {:?}", transaction);
+                panic!(
+                    "Attempted to save ledger payload hash {:?} which already exists at state version {:?}",
```
dhedey, line 212: Ditto here?

core/src/test/java/com/radixdlt/benchmark/TxnCommitAndReadBenchmarkTest.java
```diff
@@ -190,9 +193,13 @@ private void readAllTxnsThroughLedgerSync(DeterministicTest test) {
...
 
-  private List<RawLedgerTransaction> createUniqueTransactions(int numTransactions) {
+  private List<RawLedgerTransaction> createUniqueTransactions(
+      int numTransactions, int notaryPrivKeySeed) throws PrivateKeyException, PublicKeyException {
     final List<RawLedgerTransaction> res = new ArrayList<>();
-    final var notary = ECKeyPair.generateNew();
+    final byte[] prvBytes = new byte[ECKeyPair.BYTES];
+    final byte[] seedBytes = Longs.toByteArray(notaryPrivKeySeed + 1);
+    System.arraycopy(seedBytes, 0, prvBytes, prvBytes.length - seedBytes.length, seedBytes.length);
+    final var notary = ECKeyPair.fromPrivateKey(prvBytes);
```
dhedey, line 202: Out of interest, what was the bug with the previous version? Was `generateNew()` not sufficient random to guarantee unique keys (and so unique transactions?)

dhedey (approved):
I like this :).

A couple of concerns regarding reading to uphold index uniqueness invariants during commit, but otherwise, I like the simplicity of this!

RE merging please let Alexandru merge first, then I imagine you can merge this, and then Josh will merge his.

core-rust/state-manager/src/store/rocks_db.rs
LukasGasior1: Good question, I'll give it a try

core/src/test/java/com/radixdlt/benchmark/TxnCommitAndReadBenchmarkTest.java
LukasGasior1: not really a bug, but it should by slightly faster and fair when comparing results

core-rust/state-manager/src/store/rocks_db.rs
LukasGasior1: yup, seems to work fine

PR #268, opened 2023-01-25
Bugfix/fix end of epoch tc bug

talekhinezh:


LukasGasior1 (approved):
still a few **post_commit_** vertex_store left on the Rust side :) LGTM once renamed :+1:

talekhinezh:
> still a few **post_commit_** vertex_store left on the Rust side :) LGTM once renamed +1

@LukasGasior1 Where exactly?

LukasGasior1:
In `JavaCommitRequest`

core/src/integration/java/com/radixdlt/integration/targeted/consensus/EpochTimeoutCertTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+  private static final long ROUNDS_PER_EPOCH = 100;
+
+  private DeterministicTest createTest() {
+    return DeterministicTest.builder()
+        .addPhysicalNodes(PhysicalNodeConfig.createBasicBatch(4))
+        .messageSelector(MessageSelector.firstSelector())
+        .messageMutator(
+            (m, q) -> {
+              // Drop single proposal at end of epoch to cause a TC
```
dhedey, line 92: Just to check - was this test broken before your fix?

In its current formulation, I think it causes everyone to get a TC for round 102, so everyone would commit the same ledger header in round (say, 105?)

Can we have tests for both scenarios:
* A timeout in round 101/102 (everyone should progress to epoch at round 105, say)
* Proposal with QC to commit epoch change dropped in round 103 to people other than the leader and maybe 1 other. The leader would go to the next epoch before the others.

Actually, I think we can still get a liveness break if we're unlucky - if we keep getting a repeat of a leader going to the next epoch in (say) rounds 103 and rounds 106 - if the leaders drop out of consensus for epoch N when they progress to epoch N + 1? So can we have a third test for that too?

dhedey (approved):
Really nice changes, thanks.

One comment regarding the epoch timeout cert test, and possible lingering liveness issue 👍

core/src/integration/java/com/radixdlt/integration/targeted/consensus/EpochTimeoutCertTest.java
dhedey: Actually maybe this isn't an issue because round 103 would timeout, people would resend their votes, and everyone would form a timeout vertex QC...

dhedey: BUT if the proposal _did_ go to a few people, nodes would form a TC first, and advance to the next round without forming the timeout vertex QC

dhedey: If half the validator set form the TC and advance without committing hte epoch change; and hte other half are happily on the next epoch, we have a problem.

core/src/main/java/com/radixdlt/consensus/Vote.java
```diff
@@ -226,8 +226,14 @@ private ECDSASecp256k1Signature getSerializerTimeoutSignature() {
   @Override
   public String toString() {
     return String.format(
-        "%s{epoch=%s round=%s author=%s timeout?=%s %s}",
-        getClass().getSimpleName(), getEpoch(), getRound(), author, isTimeout(), highQC);
+        "%s{e=%s p=%s c=%s author=%s timeout?=%s %s}",
```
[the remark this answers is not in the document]
dhedey: Chatted to Josh - `e` / `p` / `c` aligns with the `toString` on `QuorumCertificate` - `p` and `c` correspond to "proposed" and "committed"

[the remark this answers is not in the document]
dhedey: Happy with this :)

PR #274, opened 2023-01-27
Fix: Temporarily remove docker tests from ci

dhedey:
I know Les/Theo is working on a better separation, but this just disables them fully for now until we have a better solution in place.

Note that we still have "deploy and smoke test" against Gilganet when pushing to `develop`


les-sheppard (approved):
Looks good to me - I'll get the updates done next week to only fail the action if one of the node recovery tests fail. The resource test failures will still be run but just flagged as warnings in Jenkins (and sent to the Jenkins ephemeral slack channel), so we can monitor them and flag any potentially valid issues (i.e. ones that aren't related to scrypto/core API changes)

PR #275, opened 2023-01-28
Feature: Liquid staking

talekhinezh:


core-rust/core-api-server/src/core_api/handlers/transaction_preview.rs
```diff
@@ -151,6 +151,10 @@ fn to_api_response(
             resource_changes: vec![],
             logs: vec![],
         },
+        TransactionResult::Abort(_) => {
+            // TODO: Should remove this
```
dhedey, line 155: I think all of these TODOs should be removed? I think panicking here is fine.

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -2377,6 +2386,12 @@ components:
               $ref: '#/components/schemas/EcdsaSecp256k1PublicKey'
             stake_vault:
               $ref: '#/components/schemas/EntityReference'
+            unstake_vault:
+              $ref: '#/components/schemas/EntityReference'
+            liquidity_token:
```
dhedey, line 2391: To match our documentation: https://learn.radixdlt.com/article/how-will-staking-work-on-babylon
This should be `liquid_stake_unit_resource_address`

And perhaps this should be `unstake_claim_token_resource_address`

dhedey (approved):
Looks good. Couple of very minor things

core-rust/core-api-server/src/core_api/conversions/addressing.rs
```diff
@@ -271,6 +273,18 @@ fn to_mapped_substate_id(substate_id: SubstateId) -> Result<MappedSubstateId, Ma
...
         }
+        SubstateId(RENodeId::AccessController(_), offset) => {
+            let substate_type_key = match offset {
+                SubstateOffset::AccessController(offset) => match offset {
+                    AccessControllerOffset::AccessController => (
+                        SubstateType::AccessController,
+                        SubstateKeyType::AccessController,
+                    ),
+                },
+                _ => return Err(unknown_substate_error("AccessController", &substate_id)),
```
dhedey, line 284: I imagine we're missing at least AccessRulesChain and Metadata here?

core-rust/core-api-server/src/core_api/handlers/status_network_configuration.rs
```diff
@@ -121,6 +121,11 @@ fn to_api_address_type(hrp_set: &HrpSet, entity_type: EntityType) -> models::Add
             models::EntityType::Clock,
             extract_length(ComponentAddress::Clock),
         ),
+        EntityType::AccessControllerComponent => (
```
dhedey, line 124: This, and all the three identity components, need adding to `ALL_ENTITY_TYPES` above

dhedey (approved):
Couple of small things - and the test failures to resolve

core-rust/core-api-server/src/core_api/conversions/addressing.rs
dhedey: Actually looking at https://github.com/radixdlt/radixdlt-scrypto/blob/release/betanet-v2/radix-engine/src/engine/node.rs it looks like it only supports `AccessRulesChain`. But perhaps we should also map out Metadata anyway?

talekhinezh: Cool, yeah good catch, will add these.

PR #279, opened 2023-01-30
Update max txn size limit to 1mb

LukasGasior1:


core/src/test/java/com/radixdlt/rev2/REv2LargeTransactionTest.java
```diff
@@ -124,9 +124,10 @@ private DeterministicTest createTest() {
   }
 
   private static RawNotarizedTransaction createLargeValidTransaction() {
-    var intentBytes =
+    final var blobsSize = 1024 * 1024 - 462 /* space for other intent/notarized txn fields */;
```
malexandru-rdx, line 127: How is `462` calculated? Number of signatures * signature size + some header size?

core/src/main/java/com/radixdlt/p2p/transport/PeerChannelInitializer.java
```diff
@@ -98,7 +98,9 @@ public final class PeerChannelInitializer extends ChannelInitializer<SocketChann
   // This needs to be larger than proposals / vertices
   // See also constant/s in REv2TransactionsAndProofReader
   // TODO - update this number once we have a REP/RNP for these numbers
-  private static final int MAX_PACKET_LENGTH = 25 * 1024 * 1024;
+  // Current max txn size is 1024*1024 and there are max 10 txns in a proposal
+  // + adding an extra mb for any vertex/proposal/qc overhead
+  private static final int MAX_PACKET_LENGTH = 11 * 1024 * 1024;
```
malexandru-rdx, line 103: I'd prefer it to be more verbose/built on top of other limits.
Something along the lines:
```java
private static final int MAX_PROPOSAL_TRANSACTION_COUNT = 10;
private static final int MAX_PROPOSAL_HEADER_SIZE = 1024 * 1024;
private static final int MAX_PROPOSAL_PACKET_SIZE = MAX_PROPOSAL_HEADER_SIZE + MAX_PROPOSAL_TRANSACTION_COUNT * MAX_TRANSACTION_SIZE;
```

But this is out of scope for this PR :).

malexandru-rdx (approved):

core/src/test/java/com/radixdlt/rev2/REv2LargeTransactionTest.java
LukasGasior1: yeah, it's everything else. intent header and signatures

PR #280, opened 2023-01-31
Refactoring: putting the staging into the Node

jakrawcz-rdx:
This is a refactoring (no behaviour change) in preparation for integrating the state hash into staging.

Please see the individual commit messages:
- the first one mechanically moves some staging-related files from radix-engine repo to Node's repo, as discussed with @dhedey
  - this is needed because the next PR will need to introduce the radix-engine-stores dependency (where the hash tree lives) into these structs
- the next one hides the "staged store manager" behind the "execution cache" API (i.e. makes it a true facade; not just a hashmap-slotmap key mapper)
- the last one removes some misc unneeded parts
  - @malexandru-rdx plz let me know if I mistakenly removed too much :)


jakrawcz-rdx:
Related scrypto file removals: https://github.com/radixdlt/radixdlt-scrypto/pull/794

malexandru-rdx:
Nice refactor!
Indeed the `parent_key` was not really used in staging store and probably not needed any time soon.

malexandru-rdx (approved):

jakrawcz-rdx:
@malexandru-rdx thank you for having a look so quickly 👍 let me merge.

core-rust/state-manager/src/staging/cache.rs
```diff
@@ -0,0 +1,146 @@
...
+
+use super::stage_tree::{StagedSubstateStoreKey, StagedSubstateStoreNodeKey};
+use crate::staging::stage_tree::{StagedSubstateStore, StagedSubstateStoreManager};
+use crate::AccumulatorHash;
+use radix_engine::ledger::ReadableSubstateStore;
+use radix_engine::transaction::TransactionReceipt;
+use sbor::rust::collections::HashMap;
+use slotmap::SecondaryMap;
+
+pub struct ExecutionCache<S: ReadableSubstateStore> {
```
dhedey, line 73: I'm not sold on this name - perhaps `StagedResultsCache` ?

dhedey: +1 to wrapping of the staged store.

core-rust/state-manager/src/state_manager.rs
```diff
@@ -958,14 +938,7 @@ where
...
 
-        let new_root_key = self.execution_cache.get(&parent_accumulator_hash).unwrap();
-        self.staged_store.reparent(new_root_key, &mut |key| {
-            self.execution_cache.remove_node(key);
-        });
-        self.execution_cache.set(
-            &parent_accumulator_hash,
-            StagedSubstateStoreKey::RootStoreKey,
-        );
+        self.execution_cache.progress_root(&parent_accumulator_hash);
```
dhedey, line 941: This should be next to the commit call imo.

In fact, now that this is a proper facade, this could even be part of the commit call?

dhedey: Although TBH I'd like to see:
* Store moved up under state manager
* Store managing its own internal RW Lock
* StagedStoreManager _not_ to have a store -- if absolutely necessary, it could have an Arc to it, although it probably doesn't need that

dhedey: The latter doesn't need to be part of this PR or workstream, but would be interested to hear your thoughts on it! 👍

core-rust/state-manager/src/staging/cache.rs
jakrawcz-rdx: hmm, when it was purely a "hashmap-slotmap key mapper", the `ExecutionCache` indeed seemed a bit random.
But it does have an `execute()` method right now, and it caches these executions, so the original naming suddenly clicked (to me).

I agree that its _final_ name still might differ, but it depends on how we evolve it (i.e. based on your https://github.com/radixdlt/babylon-node/pull/280#discussion_r1092102002).

dhedey: Fair enough - that makes sense.

I just just thinking about ways to distinguish it from the `pending_transaction_results_cache` - which could also be described as an execution cache. But I don't feel too strongly 👍

core-rust/state-manager/src/state_manager.rs
jakrawcz-rdx: > This should be next to the commit call imo.

I am not sure what "commit call" means here.
But in general, I left this logic precisely in the same place where it was before (I strive to not change behavior during refactorings).
(just previously there were `.reparent()` + `.set()` calls and now they are bundled together under `.progress_root()`)

Regarding "where does root store belong" - I assumed that current state is the result of compromise with Rust The Language... I will have a look whether moving store's ownership is feasible (or makes sense conceptually) and we can of course discuss options. But this may be a refactoring unrelated to a hash tree.

And "Store managing its own internal RW Lock" - yup, totally a different work stream, I'd rather not be concerned with that "RW lock" thing here.

PR #285, opened 2023-02-01
Add proposal size limit

LukasGasior1:


dhedey (approved):
Looks great, thanks Lukasz! Approving. Can you merge as soon as tests pass and let me know?

There's probably some room for some bundling of constants together (eg parameter-object refactors) and making default constants easier for integration tests -- but both of these are out of scope for this PR, and perhaps more related to @malexandru-rdx 's work :)

PR #287, opened 2023-02-01
fix: iterative staged store

malexandru-rdx:
Reimplement `delete` and `recompute_data` iteratively to avoid stack overflow issues when nodes SyncUp and when TPS will increase.


core-rust/state-manager/src/staging/stage_tree.rs
```diff
@@ -218,36 +218,38 @@ impl<S: ReadableSubstateStore> StagedSubstateStoreManager<S> {
...
-        }
-    }
-
     /// Rebuilds ImmutableStores by starting from the root with new, empty ones
     /// and recursively reapplies the [`receipt`]s.
     fn recompute_data(&mut self) {
         // Reset the [`dead_weight`]
         self.dead_weight = 0;
 
+        let mut stack = Vec::new();
```
dhedey, line 227: Do you think it might be worth a quick comment explaining why we have done it this way instead of recursively? TO prevent some future well-meaning soul refactoring it back into recursion? :')

(Ditto below?)

dhedey (approved):
Looks good! Minor suggestion if you agree with it, then merge this when you're happy and let me know so I can create a new tag for Krzysztof

mozi-rdx (approved):

PR #290, opened 2023-02-01
Refactoring: Generic staging tree

jakrawcz-rdx:
This is a follow-up to https://github.com/radixdlt/babylon-node/pull/280.

The obvious next step is to add `TreeDelta` field (to `Delta` struct) and `TreeStore` field (to `Accumulator`), which solves the hash tree staging.

### DIFF
--- a/core-rust/state-manager/src/staging/stage_tree.rs
+++ b/core-rust/state-manager/src/staging/stage_tree.rs
@@ -82,12 +82,26 @@ pub enum StageKey {
 
 /// A change to the accumulated state.
 pub trait Delta {
+    /// Returns an estimated memory impact of this delta (measured in arbitrary units, relative to
+    /// other possible deltas).
+    /// This value is used during decision on amortized "reparenting" of the stage tree.
     fn weight(&self) -> usize;
 }
 
 /// An accumulated state coming from multiple [`Delta`]s.
-pub trait Accumulator<D: Delta>: Default + Clone {
+pub trait Accumulator<D: Delta> {
+    /// Creates a new, empty instance.
+    fn create_empty() -> Self;
+
+    /// Incorporates the changes coming from the given delta into this accumulator.
     fn accumulate(&mut self, delta: &D);
+
+    /// Performs a constant-time (and thus, necessarily, constant-space) clone of this accumulator.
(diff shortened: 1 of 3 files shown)


jakrawcz-rdx:
@malexandru-rdx I value your opinion here, please take a look.

@dhedey I see that this PR is not crazy large or complex yet; so I will probably fit in one more commit, which moves the `root_store` ownership even higher (as we discussed today) - it is coming soon.

jakrawcz-rdx:
> I will probably fit in one more commit,(...)

done, 100% ready for review.

core-rust/state-manager/src/staging/stage_tree.rs
```diff
@@ -62,243 +62,168 @@
...
-    }
+/// A change to the accumulated state.
+pub trait Delta {
+    fn weight(&self) -> usize;
 }
 
-new_key_type! {
-    pub struct StagedSubstateStoreNodeKey;
+/// An accumulated state coming from multiple [`Delta`]s.
+pub trait Accumulator<D: Delta>: Default + Clone {
```
malexandru-rdx, line 89: I would add a comment to let the user know that `Clone` should be O(1) in order to keep the complexity as intended.

malexandru-rdx (approved):
Excellent. I love how cleaner it feels. :)

Approved. I've left 1 small comment suggestion for future usage.

core-rust/state-manager/src/staging/stage_tree.rs
malexandru-rdx: Maybe rename it to `ImmutableAccumulator` to be more clear.

jakrawcz-rdx: I see why you are concerned, so I now did something even clearer 👍 PTAL.

_(I deliberately avoid using the word "immutable" here, because future developers [incl. future me] will definitely understand it wrong)_

malexandru-rdx: Thank you! That is better.

jakrawcz-rdx:
Let me merge this, since I have next changes sitting on top of it.

PR #292, opened 2023-02-01
Add max num txns and payload size limits to mempool relayer

LukasGasior1:


dhedey (approved):
Really nice changes, thanks!

PR #294, opened 2023-02-02
Lazy execute initial vertices in VertexStore

LukasGasior1:


core/src/test/java/com/radixdlt/recovery/REv2ConsensusLedgerRecoveryTest.java
```diff
@@ -131,8 +130,8 @@ public void recovery_should_work_when_consensus_is_behind_ledger() {
       test.runNext(
           msg -> msg.message() instanceof Proposal && msg.channelId().isLocal(behindNodeIndex));
 
-      // Assert: Can run bft syncing with no issue
-      test.runForCount(100, onlyBFTSyncEvents());
+      // Assert: Can run consensus with no issue
+      test.runUntilState(allAtOrOverStateVersion(50), 300, onlyConsensusEvents());
```
LukasGasior1, line 134: We don't prune the vertex store state anymore, so there's no need to sync after reboot. Instead, making sure that consensus runs fine.

core/src/main/java/com/radixdlt/consensus/bft/VertexStoreJavaImpl.java
```diff
@@ -360,10 +369,15 @@ private Optional<CommittedUpdate> commit(BFTHeader header, QuorumCertificate com
...
-    while (vertex != null) {
-      path.addFirst(vertex);
-      vertex = vertices.get(vertex.getParentId());
+    var executedVertexOpt = getExecutedVertex(vertexId);
+    while (executedVertexOpt.isPresent()) {
+      final var v = executedVertexOpt.unwrap();
+      path.addFirst(v);
+      if (v.getVertexHash().equals(v.vertex().getParentVertexId())
+          || v.getVertexHash().equals(this.rootVertex.hash())) {
+        break;
```
dhedey, line 378: Shouldn't it be a `return path` here, and we should throw an Exception (or return None) if we execute the path via an ExecutedVertex being `Some`?

(other than if `vertexId = root`, obvs)

core/src/main/java/com/radixdlt/consensus/bft/VertexStoreJavaImpl.java
```diff
@@ -250,7 +234,31 @@ public boolean insertTimeoutCertificate(TimeoutCertificate timeoutCertificate) {
   // TODO: reimplement in async way
   @Override
```
dhedey, line 235: Maybe it's worth a comment here (and/or in start-up) explaining why we execute lazily? 👍

dhedey (approved):
Looks good :)... A couple of minor thoughts to slightly improve handling of edge cases

core/src/main/java/com/radixdlt/consensus/bft/VertexStoreJavaImpl.java
LukasGasior1: :+1: added a comment

core/src/main/java/com/radixdlt/consensus/bft/VertexStoreJavaImpl.java
LukasGasior1: hmh not sure I get it, `return path` is equivalent to `break` here? why would we throw if executed vertices are some?

dhedey: Well if we don't find a vertex on the path to the root we simply return an incomplete path at the moment.

This feels wrong / buggy?

core/src/main/java/com/radixdlt/consensus/bft/VertexStoreJavaImpl.java
dhedey: Looks good! But can we make the comment an actual Java Doc, so it appears in peoples' IDEs if they hover over the method?

core/src/main/java/com/radixdlt/consensus/bft/VertexStoreJavaImpl.java
dhedey: Can we also possibly add a JavaDoc comment to explain that the path doesn't contain the root vertex? :)

LukasGasior1: ah, but I think that's needed for a scenario when ledger is ahead of vertex store

PR #299, opened 2023-02-03
Additional round time (delay timeout event) if proposal was received

LukasGasior1:


core/src/integration/java/com/radixdlt/integration/steady_state/deterministic/consensus/OneProposalTimeoutResponsiveTest.java
```diff
@@ -104,24 +104,28 @@ private void run(int numValidatorNodes, long numRounds, long dropPeriod) {
     test.runUntilMessage(DeterministicTest.hasReachedRound(Round.of(numRounds)));
 
     long requiredIndirectParents =
-        numValidatorNodes <= 3
-            ? 0 // there are no indirect parents for 3 nodes (QC is always formed)
+        numValidatorNodes <= 4
```
LukasGasior1, line 107: the change is because the leader receives their own proposal and has their round extended, while others form a timeout QC (instead of a TC from mixed proposal and empty vertex votes)

LukasGasior1:
Summary of changes:
* round timeout can be extended if
a) we have received a message containing a QC for this round (but not necessarily synced up to it yet) - this prevents (to some degree) TC formation for rounds that already have a QC
b) we have received a proposal for this or next round - this prevents timeouts (and forming TC) when the proposal has been received but takes longer than usual to sync/process

* limit (only on the proposal generation side, not strictly enforced) the maximum size for a proposal and its previous vertices, which effectively limits the size of a commit batch under a single proof

* tweak mempool relayer limits

* make p2p liveness monitor less agressive (send ping every 60s and wait up to 30s for a response before dropping the connection)

core/src/main/java/com/radixdlt/consensus/bft/processor/BFTEventStatelessVerifier.java
```diff
@@ -182,15 +186,20 @@ public void processVote(Vote vote) {
 
   @Override
   public void processProposal(Proposal proposal) {
-    if (!isAuthorInValidatorSet(proposal)) {
+    final var expectedAuthor = proposerElection.getProposer(proposal.getRound());
+    if (!proposal.getAuthor().equals(expectedAuthor)) {
```
dhedey, line 190: Looks like `isAuthorInValidatorSet()` can be removed?

core/src/main/java/com/radixdlt/rev2/REv2StateComputer.java
```diff
@@ -97,32 +98,46 @@ public final class REv2StateComputer implements StateComputerLedger.StateCompute
...
       EventDispatcher<LedgerUpdate> ledgerUpdateEventDispatcher,
       EventDispatcher<MempoolAddSuccess> mempoolAddSuccessEventDispatcher,
       EventDispatcher<ConsensusByzantineEvent> consensusByzantineEventEventDispatcher,
-      Serialization serialization) {
+      Serialization serialization,
+      Metrics metrics) {
     this.stateComputer = stateComputer;
     this.maxNumTransactionsPerProposal = maxNumTransactionsPerProposal;
     this.maxProposalTotalTxnsPayloadSize = maxProposalTotalTxnsPayloadSize;
+    this.maxProposalAndUncommittedVerticesTotalTxnPayloadSize =
```
dhedey, line 133: Ahh so this is really `maxUncommittedUserTransactionsTotalTxnPayloadSize`? (ie Transactions as opposed to Vertices)

I think it'd make sense to have `maxUncommittedUserTransactionsTotalCount` - and I guess to validate this during proposal validation time, based on the "Committed" in its parent QC.

But these can all wait for when we implement that ticket in the work tracker 👍

core/src/main/java/com/radixdlt/rev2/REv2StateComputer.java
```diff
@@ -165,10 +176,41 @@ public List<RawNotarizedTransaction> getTransactionsForProposal(
...
+                maxNumTransactionsPerProposal, maxPayloadSize, rawPreviousExecutedTransactions)
+            : List.<RawNotarizedTransaction>of();
+
+    final var resultTotalTxnPayloadSize =
+        result.stream().map(tx -> tx.getPayload().length).reduce(0, Integer::sum);
+
+    final var totalUncommittedTxnBytesIncludingThisProposal =
+        resultTotalTxnPayloadSize + rawPreviousExecutedTransactionsTotalSize;
+
+    metrics.ledger().numTransactionsIncludedInProposal().inc(result.size());
```
dhedey, line 206: These metrics should really be Summaries (or histograms - although we don't have support for these in Java), not Counters - summaries track both "total" and "count" - see: https://prometheus.io/docs/practices/histograms/

dhedey: Out of interest why did you put these in `ledger`, rather than `bft`?

It's also a little weird they are leader-only metrics? Shouldn't these metrics be incremented during proposal processing, so they can be recorded by all nodes in consensus?

If you want to keep them here, they should include `leader` in their name imo 👍

dhedey (approved):
Thanks Lukasz. Some really nice work.

Have a few different comments / markups - I'd like all of them resolved soon, some are easier fixes than others - ideally I'd like all of them in for Betanet V2, but none are must-includes for Betanet.

Let's do the low-hanging fruit, and then get this merged for Monday lunchtime so I can pull together Betanet V2 RC2, and we can always pull together a Betanet V2.1 in the coming few days 👍

core/src/main/java/com/radixdlt/rev2/REv2StateComputer.java
LukasGasior1: Yeah my idea was to report these values for each node as a leader.
:+1: for Summaries
:+1: for moving to `bft`
:+1: for prepending `leader`

core/src/main/java/com/radixdlt/rev2/REv2StateComputer.java
LukasGasior1: > Ahh so this is really maxUncommittedUserTransactionsTotalTxnPayloadSize? (ie Transactions as opposed to Vertices)

yeah, I guess so :sweat_smile:
naming is hard sometimes, what I meant was
`maxProposalAndUncommittedVerticesTotalTxnPayloadSize` == `max` `TotalTxnPayloadSize` in `Proposal` `And` `UncommittedVertices`

LukasGasior1: going for `maxUncommittedUserTransactionsTotalPayloadSize` then :+1:  (without the redundant `Txn`)

core/src/main/java/com/radixdlt/consensus/bft/processor/BFTEventStatelessVerifier.java
LukasGasior1: we still use that for verifying votes

core/src/main/java/com/radixdlt/consensus/bft/processor/SyncUpPreprocessor.java
```diff
@@ -238,53 +254,20 @@ public void start() {
...
           NodeId.fromPublicKey(proposal.getAuthor().getKey()),
           () -> processOnCurrentRoundOrCache(proposal, forwardTo::processProposal));
-    } else {
-      log.trace(
-          "Proposal: Ignoring for past round {}, current round is {}", proposal, currentRound);
-      return true;
+      case Vote vote -> syncUp(
+          vote.highQC(),
+          NodeId.fromPublicKey(vote.getAuthor().getKey()),
+          () -> processOnCurrentRoundOrCache(vote, forwardTo::processVote));
```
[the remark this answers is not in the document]
LukasGasior1: Hmh, I think this is okay (although confusing!):
* new events (`processProposal`, `processVote`) and events that have been synced to the correct round (`processRoundCachedEvent`) are processed using `syncUpAndProcess`, which caches `IN_PROGRESS` events, immediately processes `SYNCED` events and ignores `INVALID` events
* events that have been bft synced up (`processQueuedConsensusEvent`) seem to just re-use the `syncUp` method to pass the even through, assuming the result is always `SYNCED` (it should be) :man_shrugging:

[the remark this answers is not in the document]
LukasGasior1: the could be re-using `syncUp` because of the additional end-of-epoch logic there

core/src/main/java/com/radixdlt/consensus/bft/processor/SyncUpPreprocessor.java
```diff
@@ -238,53 +254,20 @@ public void start() {
     forwardTo.start();
   }
 
-  // TODO: rework processQueuedConsensusEvent(), processVoteInternal() and processProposalInternal()
+  // TODO: rework processQueuedConsensusEvent()
```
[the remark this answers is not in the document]
LukasGasior1: removed; if any rework is to happen it's not just `processQueuedConsensusEvent` :laughing:

core/src/main/java/com/radixdlt/consensus/bft/processor/ProposalTimestampVerifier.java
```diff
@@ -79,16 +79,26 @@
 import org.apache.logging.log4j.Logger;
 
 /** Verifies proposal timestamps against the local system time. */
+/*
+TODO: address the following
+Currently max allowed daley time is quite permissive to account for
+processing proposals that take much time to sync/execute.
```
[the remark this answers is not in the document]
LukasGasior1: you're right, this should really be "that take much time to sync ~~/execute~~"

core/src/main/java/com/radixdlt/rev2/REv2TransactionsAndProofReader.java
```diff
@@ -82,11 +82,11 @@ public final class REv2TransactionsAndProofReader implements TransactionsAndProo
    *    can still be returned (given they fit under MAX_TXN_BYTES_FOR_A_SINGLE_RESPONSE limit)
    * b) if next proof contains less than MAX_TXNS_FOR_RESPONSES_SPANNING_MORE_THAN_ONE_PROOF txns then
    *    subsequent proof can only be used if its txns fit under this limit */
-  private static final int MAX_TXNS_FOR_RESPONSES_SPANNING_MORE_THAN_ONE_PROOF = 1000;
+  private static final int MAX_TXNS_FOR_RESPONSES_SPANNING_MORE_THAN_ONE_PROOF = 100;
```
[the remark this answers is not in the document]
LukasGasior1: this might just be a relic of me testing various combinations :sweat_smile:

core/src/main/java/com/radixdlt/rev2/REv2StateComputer.java
dhedey: Sounds good - in a future ticket, I think we should _also_ report these values as someone receiving the proposals - to ensure we can EG debug mainnet using our foundation validators or understand if validators are acting up :)

core/src/main/java/com/radixdlt/consensus/bft/processor/SyncUpPreprocessor.java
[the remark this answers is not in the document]
dhedey: Maybe creation of separate methods: `syncToQc(HighQC highQC, @Nullable NodeId author) -> boolean`, and `processVoteGivenSyncedUp` / `processProposalGivenSyncedUp` which do the end of epoch check?

PR #303, opened 2023-02-07
Refactoring NativeCalls to introduce a stepped builder (for future metrics integration)

jakrawcz-rdx:
This is a pure refactoring (no behavior change!) which should enable easier impl of point 5. of https://radixdlt.atlassian.net/browse/NODE-509.

In this PR:
- I replace a single `NativeCalls.with(...)` factory method with a stepped builder
  - currently there are not that many steps, but the entire idea is to have an optional `.measured(...)` building step - coming in the next PR
- I remove the specialized impls of `Func` vs `StaticFunc`
  - ofc these are still 2 different APIs, but backed by a single impl
- I reduce the number of `new TypeToken<>() {}` expressions to 1 per call definition


dhedey (commented):
Looks good :)

I'm a little sad that we've lost the distinction in the definition and/or type signature of whether it's a StateManager function or not in the call signature - because it matters on the Rust side.

But actually, it looks like this is still included in the native definition, so maybe this isn't so bad! 👍

dhedey (approved):

PR #304, opened 2023-02-08
Update scrypto dep

jakrawcz-rdx:
To `develop-baf0142b2`.


jakrawcz-rdx:
⚠️ WIP ⚠️ 

This PR is WIP builds, but most likely maaaany tests fail, since I left a few TODOs for cases where I was not sure how to proceed.

These TODOs are marked with `// TODO(code review): <details>`. They are related to:
- static `ACCOUNT_PACKAGE_ADDRESS` no longer existing, but still used (at least in tests?) - @talekhinezh can you advice?
- new `Account` substate type: I assumed it is supposed to be propagated up to the core API, but I have no idea about its representation - @dhedey?
- introduction of `NodeModuleId`:
  - in many places I just inserted `::SELF` - @talekhinezh please verify
  - also, maybe this one has to somehow reach the core API? - @dhedey please see my question regarding `MappedSubstateId`
- the `new_account_with_resource` got deleted, and I assume that now a `new_account` + `deposit entire worktop` combo is supposed to be used, but I am too lame to write a working piece of manifest like that - @talekhinezh I will appreciate your help!

core-rust/core-api-server/src/core_api/conversions/addressing.rs
```diff
@@ -227,13 +226,14 @@ fn to_mapped_substate_id(substate_id: SubstateId) -> Result<MappedSubstateId, Ma
 
     // Start body of method
     let entity_id_bytes = re_node_id_to_entity_id_bytes(&substate_id.0)?;
-    let substate_key_bytes = substate_offset_to_substate_key_bytes(&substate_id.1)?;
+    // TODO(code review): do we need to include the node module id in the MappedSubstateId too?
```
talekhinezh, line 229: My preference here is yes so that there is little abstraction between the Core API and the underlying ledger state but I think it mostly depends on what is the Gateway consumer ready to handle.

core-rust-bridge/src/main/java/com/radixdlt/rev2/ScryptoConstants.java
```diff
@@ -79,10 +79,8 @@ public final class ScryptoConstants {
           .build(new TypeToken<Natives.Call1<Tuple.Tuple0, ComponentAddress>>() {})
           .call(Tuple.Tuple0.of());
 
-  public static final PackageAddress ACCOUNT_PACKAGE_ADDRESS =
-      Natives.builder(ScryptoConstants::getAccountPackageAddress)
-          .build(new TypeToken<Natives.Call1<Tuple.Tuple0, PackageAddress>>() {})
-          .call(Tuple.Tuple0.of());
+  // TODO(code review): how to replace this no-longer-existing static address?
+  public static final PackageAddress ACCOUNT_PACKAGE_ADDRESS = new PackageAddress(new byte[27]);
```
talekhinezh, line 83: There is no address currently but there will be one in a soon to come PR on the scrypto side. For now, this can be removed.

core-rust/core-api-server/src/core_api/conversions/addressing.rs
```diff
@@ -255,11 +255,14 @@ fn to_mapped_substate_id(substate_id: SubstateId) -> Result<MappedSubstateId, Ma
                         SubstateType::ComponentState,
                         SubstateKeyType::ComponentState,
                     ),
-                    ComponentOffset::RoyaltyConfig => (
+                },
+                // TODO(code review): is this the right mapping?
```
talekhinezh, line 259: This looks fine.

core-rust/core-api-server/src/core_api/conversions/addressing.rs
```diff
@@ -277,7 +280,21 @@ fn to_mapped_substate_id(substate_id: SubstateId) -> Result<MappedSubstateId, Ma
             };
             (EntityType::Component, substate_type_key)
         }
-        SubstateId(RENodeId::AccessController(_), offset) => {
+
+        SubstateId(RENodeId::Account(_), _, offset) => {
+            let substate_type_key = match offset {
+                SubstateOffset::Account(offset) => match offset {
+                    AccountOffset::Account => {
+                        // TODO(code review): I just blindly defined these - correctly?
```
talekhinezh, line 288: This is fine.

core-rust/core-api-server/src/core_api/conversions/addressing.rs
```diff
@@ -299,17 +316,20 @@ fn to_mapped_substate_id(substate_id: SubstateId) -> Result<MappedSubstateId, Ma
...
-        SubstateId(RENodeId::Package(_), offset) => {
+        SubstateId(RENodeId::Package(_), _, offset) => {
             let substate_type_key = match offset {
                 SubstateOffset::Package(offset) => match offset {
                     PackageOffset::Info => {
                         (SubstateType::PackageInfo, SubstateKeyType::PackageInfo)
                     }
-                    PackageOffset::RoyaltyConfig => (
+                },
+                // TODO(code review): is the right mapping?
```
talekhinezh, line 326: This is fine.

core-rust/core-api-server/src/core_api/conversions/receipt.rs
```diff
@@ -188,7 +188,8 @@ pub fn to_api_fee_summary(
         cost_unit_execution_breakdown: fee_summary
             .execution_cost_unit_breakdown
             .into_iter()
-            .map(|(key, cost_unit_amount)| (key, to_api_u32_as_i64(cost_unit_amount)))
+            // TODO(code review): the fee summary now uses CostingReason enum; we map it like that?
```
talekhinezh, line 191: Sounds okay to me.

core-rust/core-api-server/src/core_api/generated/models/substate.rs
```diff
@@ -13,6 +13,10 @@
 #[derive(Clone, Debug, PartialEq, serde::Serialize, serde::Deserialize)]
 #[serde(tag = "substate_type")]
 pub enum Substate {
+    #[serde(rename="Account")]
+    AccountSubstate {
+        // TODO(code review): what do we return?
```
talekhinezh, line 18: This should be generated by updating the `core-api-schema.yaml` file and then running:
```
python3 core-rust/core-api-server/scripts/generate-openapi-server.py
```

I would suggest following the same pattern used by AccessController.

core-rust/state-manager/src/transaction/builder.rs
```diff
@@ -106,10 +105,9 @@ pub fn create_new_account_intent_bytes(
     let manifest = ManifestBuilder::new()
         .lock_fee(FAUCET_COMPONENT, 100.into())
         .call_method(FAUCET_COMPONENT, "free", args!())
-        .take_from_worktop(RADIX_TOKEN, |builder, bucket_id| {
-            builder.new_account_with_resource(&AccessRule::AllowAll, bucket_id)
-        })
+        .new_account(&AccessRule::AllowAll)
         .build();
+    // TODO(code review): now somehow deposit the worktop to the created account?
```
talekhinezh, line 110: ooo this may be impossible with the current setup since there is no way to get a reference to the new account just created.

I'd either suggest holding off on the PR until the scrypto side gets updated, or ignoring the tests which currently use this functionality, or update the tests which use to use virtual accounts instead of new accounts.

core-rust/core-api-server/src/core_api/generated/models/substate.rs
jakrawcz-rdx: > This should be generated

Huh, I guess the `/generated/` in the path is not enough of a hint for me 🤷‍♂️ 

> following the same pattern used by AccessController

I interpret this as "literally return the same `data_struct`" (which reveals I have no clue about that api) 👍

core-rust/core-api-server/src/core_api/helpers.rs
```diff
@@ -50,7 +51,7 @@ pub(crate) fn read_known_substate(
     renode_id: RENodeId,
     substate_offset: &SubstateOffset,
 ) -> Result<PersistedSubstate, ResponseError<()>> {
-    let substate_id = SubstateId(renode_id, substate_offset.clone());
+    let substate_id = SubstateId(renode_id, NodeModuleId::SELF, substate_offset.clone());
```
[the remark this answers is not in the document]
jakrawcz-rdx: ah, thanks, I now tried to pass my best guesses, PTAL.

core-rust/core-api-server/src/core_api/conversions/addressing.rs
jakrawcz-rdx: alright, I consulted @KrzysztofLabus-RDX and we came up with `module_type` name for the new field; please see if this mapping is good enough.

core-rust/state-manager/src/transaction/builder.rs
jakrawcz-rdx: AFAIS there is only 1 test relying on this, so I guess it doesn't hurt to ignore it temporarily 👍

core/src/test/java/com/radixdlt/rev2/REv2ConsensusTransferTest.java
```diff
@@ -132,6 +133,7 @@ private static RawNotarizedTransaction createNewAccountTransaction() {
   }
 
   @Test
+  @Ignore("the TransactionBuilder's create_new_account_intent_bytes() helper is temporarily broken")
```
dhedey, line 136: I'd suggest creating a `createNewVirtualAccountWithXrdTransaction(publicKey)` via using:
```
var virtualAccountAddress = Address.virtualAccountAddress(publicKey);
...
```

And a manifest which transfers XRD to the virtual account, thereby creating it:
```
CALL_METHOD
    ComponentAddress("${virtualAccountAddress}") 
    "deposit_batch"
    Expression("ENTIRE_WORKTOP");
```

dhedey (approved):
Really nice work. Happy to merge once the package address issue is resolved.

core/src/test/java/com/radixdlt/rev2/REv2ConsensusTransferTest.java
jakrawcz-rdx: The "virtual account" snippets are super helpful for adjusting many of the failing tests, thanks 👍 

However this one here is named `new_account_creates_transfer_of_xrd_to_account`, which suggests that it wants to test exactly the "new account with resource" function, which is no longer available (in fact, this is the only test that calls Rust jni `create_new_account_intent_bytes()` method).

If I adjust the `create_new_account_intent_bytes()` to use virtual account, then it would mean renaming it (to e.g. `deposit_xrd_to_virtual_account_intent_bytes()`) and also renaming the test to `can_transfer_xrd_to_account`. I could do it, but it is something already tested by other tests.

So I think we are left with options:
A) Keep the test `@Ignore`d temporarily, since Josh suggests that "scrypto side will get updated" (making this function available [and worth testing] again, I assume).
B) Remove the test and the entire jni `create_new_account_intent_bytes()`.

KrzysztofLabus-RDX (approved):

core/src/test/java/com/radixdlt/rev2/REv2ConsensusTransferTest.java
dhedey: I'm not sure (A) will be added back, and if it is, I don't think it will be tested here, so happy with (B).

That said, I think this might be the only test of the `get_component_resources` function - but that's just a function for use by tests anyway so I don't think it's a problem.

jakrawcz-rdx: if I am looking at the right `get_component_resources`, then it is still used by other tests (via `get_component_xrd`).

So with no remorse, I deleted this test 🤷‍♂️

jakrawcz-rdx:
I've been able to fix most of the tests. There are 2 mysteries left, both marked with `TODO(code review)` right now in this PR:

1. `state_reader_on_genesis_returns_correct_amounts()` - we expect X staked, but 0 gets staked. @LukasGasior1 might be able to help?

2. `test_core_api_can_submit_and_commit_transaction()` - suddenly `streamTransactionsPost()` started throwing 500 (this is the only test trying to do that, AFAIS). Looks treatable, I just gotta learn how to enter with some debug / extensive logging there.

jakrawcz-rdx:
The mystery 1. of https://github.com/radixdlt/babylon-node/pull/304#issuecomment-1424271184 is clarified a bit: our test util for reading out account balances does not know about the new `Account` substate, and it now reports 0 (thanks @talekhinezh for confirming!). I am on the right track to fix it 👍

jakrawcz-rdx:
Finally, all tests pass 🎉

PR #309, opened 2023-02-10
feature: Core API state for validators and access controllers

dhedey:
Another PR to expose more data about validators, access controllers and the epoch manager (including active validator set).

This means that:
* You can get the active validator set from `core/state/epoch`
* You can query validator state from `core/state/validator`
* You can query access controller state from `core/state/access-controller`


LukasGasior1 (approved):

core-rust/core-api-server/src/core_api/server.rs
```diff
@@ -142,6 +142,11 @@ pub async fn create_server<F>(
         .route("/state/epoch", post(handle_state_epoch))
         .route("/state/clock", post(handle_state_clock))
         .route("/state/component", post(handle_state_component))
+        .route("/state/validator", post(handle_state_validator))
+        .route(
```
les-sheppard, line 146: Extremely minor, but should this be on a single line, for consistency?

dhedey: So we use autoformatters (cargo format and clippy) which do a good job at generally cleaning things up, but sadly in specific circumstances, they can make things look rubbish or inconsistent. Sadly it looks like this is one of those cases! :)

PR #310, opened 2023-02-11
tweak: Adjust XRD genesis stake amounts

dhedey:
Tweaks the Genesis stake amounts so that:
* On testnets (gilganet/enkinet/hammunet/mardunet/nergalnet/nebunet), we use the "Babylon Testnets Powerful Staking Address" for manually adjusting validator balances. This means that: we stake 1 billion XRD per validator, with 700 billion XRD being available to the staking address for further staking.
* On other environments, we still use 1 XRD per validator - to enable tests to easily change the validator set


LukasGasior1 (approved):
LGTM

PR #311, opened 2023-02-11
feat: Improved pending transaction execution behaviour

dhedey:
Implements the following:
* Removed the 500ms arbitrary execution limit - now that the engine has better limits, and consensus is more resilient, this should be safe to remove (it was a temporary addition to restrict certain kinds of issues in Betanet V1)
* Pending transactions are only calculated up to fee loan repayment
* Transactions submitted with a future from_epoch can be successfully resubmitted as soon as the epoch is hit


jakrawcz-rdx (approved):
If this means anything, it LGTM 👍

PR #312, opened 2023-02-13
Including the state hash returned from Rust in the LedgerHeader

jakrawcz-rdx:
Part of https://radixdlt.atlassian.net/browse/NODE-515.

This returns the results of Rust's JMT up to the Java's `LedgerHeader`.

Notes:
- I have left a single very-far-future `// TODO:` (about including the state hash in the consensus)
  - we will want it implemented only after making sure there are no engine non-det bugs
- I feel like we would like to do something more with that `stateHash`:
  - expose it via some API?
  - store it in some rocksdb table?
  - log a warn somewhere if it doesn't match across nodes?

  Please advice @dhedey @talekhinezh on what/where/how ☝️


jakrawcz-rdx:
Weird:
`Test recovery_should_work_when_consensus_is_behind_ledger FAILED`
(passes on develop fine)

Investigating.

jakrawcz-rdx:
I did some heavy debugging and found only a couple of typos (fixed in the newest commit), which didn't affect the failing test :/

The problem is probably caused by non-determinism (either in the engine or in the hash tree calculations) exposed by Java test `recovery_should_work_when_consensus_is_behind_ledger()` (in which, AFAIU, there is a liveness break, although I may be wrong about why it really fails).

Regardless, you may ask: why non-determinism would affect the consensus or liveness.
In this PR, the `stateHash` is added to `LedgerHeader`, but it is never directly compared with anything (i.e. the consensus logic does not do `.getStateHash()`).
Removing the `stateHash` from `LedgerHeader.hashCode()` and `.equals()` also does NOT fix anything.

However, removing the `stateHash` from serialization, like this...
image
... makes the test pass 🤔 
(with that field out of serialized form, I can even return `random()` from `.getStateHash()` and everything works)

Is it possible that a serialized form of `LedgerHeader` is compared across nodes or something?

jakrawcz-rdx:
Ok, the answer to the above question is: "yes" (from @LukasGasior1), e.g. here:
https://github.com/radixdlt/babylon-node/blob/55de99f1e0b3bf0fd9a831a48f4d2af8a1d85391/core/src/main/java/com/radixdlt/consensus/QuorumCertificate.java#L158
(the `voteData` contains `LedgerHeader`)

So it means, that this very PR indirectly incorporates the `stateHash` into consensus' logic :(
We have to get rid of that.

jakrawcz-rdx:
Ok, I have some findings regarding the state non-determinism exposed by that test.

I caught a single accumulator hash which maps to 2 different state hashes, during the execution of the test. I captured the rocksdb's actual substate store dumps, and they differ in the following ways:
- A has 104 substates, B has 118 substates
- almost all the changes are "B has a substate that A has not", e.g.:
   image
- there is also a couple of more subtle changes, where value of the same substateId differs just a bit, e.g.:
   image
   (here the `amount` got super-weird in one case?)
   image
   (here the EpochManager says a round is different!?)
- there are no changes of type "A has a substate that B has not"

jakrawcz-rdx:
(for curious, here are my exact captured substate stores for the same accu hash: [dup1.txt](https://github.com/radixdlt/babylon-node/files/10744074/dup1.txt) and [dup2.txt](https://github.com/radixdlt/babylon-node/files/10744083/dup2.txt))

jakrawcz-rdx:
@dhedey FYI:

Together with @LukasGasior1 we concluded that skipping the `stateHash` field in Dson.HASH is ugly, but can be used temporarily, until we fix the most pressing non-det bugs (including the one in that UT), in some future PR.

This PR is ready for review as-is.

core/src/main/java/com/radixdlt/consensus/LedgerHeader.java
```diff
@@ -108,6 +109,10 @@ public final class LedgerHeader {
   @DsonOutput(Output.ALL)
   private final AccumulatorState accumulatorState;
 
+  @JsonProperty("state_hash")
+  @DsonOutput(value = Output.HASH, include = false)
```
dhedey, line 113: I think we want to include it when we save/load and when we send it over the API, but just not include it for hashing, right? so EG:
```
  @DsonOutput(value = {Output.API, Output.WIRE, Output.PERSIST})
```

dhedey:
RE this question:

> I feel like we would like to do something more with that stateHash

Possibly at commit, if the calculated hash mismatches with the expected state hash from the ledger proof, we can log and maybe have a "state_hash_diverged" metric?

This might mean _lots_ of logs, so we may wish to have some kind of log limiter, but I don't mind too much.
On prod, it'll be maximum of once per round (maybe 6 times per second).

core/src/main/java/com/radixdlt/consensus/LedgerHeader.java
jakrawcz-rdx: > I think we want to include it when we save/load and when we send it over the API, but just not include it for hashing, right?

Yes. And `@DsonOutput(value = Output.HASH, include = false)` means exactly "use it for every output except HASH". That's how I interpret the `include` javadoc and in the impl:
https://github.com/radixdlt/babylon-node/blob/93df7f8a81f480371b070e1c33399aef0dcdd43a/common/src/main/java/com/radixdlt/serialization/DsonOutput.java#L171-L172
(and the runtime behaviour seems to match this expectation)

jakrawcz-rdx:
And regarding the exposed non-det bug itself:
image

As I previously noted (https://github.com/radixdlt/babylon-node/pull/312#issuecomment-1430890235), this unit test was PASSING for me in this very PR before update from develop. So if this is a Node's bug, then it must have been introduced to develop (or merged from betanet-v2 to develop) very recently, 2-3 days ago. FYI @LukasGasior1 if you are planning to look at this.

jakrawcz-rdx:
alright, so only `MultiNodeRebootTest.restart_all_nodes_intermittently` is failing, but I see it also failing for other pending PRs - I am assuming it is not related to my changes. Let me merge.

core/src/main/java/com/radixdlt/consensus/LedgerHeader.java
dhedey: Ahh cool! Nice - didn't know that 👍

jakrawcz-rdx:
@LukasGasior1 I have analyzed one of your suspects for the non-determinism: https://github.com/radixdlt/babylon-node/pull/294
I was running the `recovery_should_work_when_consensus_is_behind_ledger()` in different configurations.

- When enabled `@DsonOutput(Output.ALL)` for `stateHash`:
  - When when `VertexStoreJavaImpl` (+ `VertexStoreState`) has your recent changes:
    - When the test has your recent changes (`test.runUntilState(..., onlyConsensusEvents());`
      - The test FAILS 🔴 
    - When the test does NOT have your recent changes (`test.runForCount(100, onlyBFTSyncEvents())`):
      - The test FAILS 🔴
  - When when `VertexStoreJavaImpl` (+ `VertexStoreState`) does NOT have your recent changes:
    - When the test has your recent changes (`test.runUntilState(..., onlyConsensusEvents());`
      - The test FAILS 🔴 
    - When the test does NOT have your recent changes (`test.runForCount(100, onlyBFTSyncEvents())`):
      - The test PASSES 🟢 _(as if your change did NOT happened, but my did)_
- When disabled (`@DsonOutput(Output.HASH, include=false)`):
  - When when `VertexStoreJavaImpl` (+ `VertexStoreState`) has your recent changes:
    - When the test has your recent changes (`test.runUntilState(..., onlyConsensusEvents());`
      - The test PASSES 🟢 _(no surprise: this is kinda `develop`'s state **after** your PR)_
    - When the test does NOT have your recent changes (`test.runForCount(100, onlyBFTSyncEvents())`):
      - The test FAILS 🔴 
  - When when `VertexStoreJavaImpl` (+ `VertexStoreState`) does NOT have your recent changes:
    - When the test has your recent changes (`test.runUntilState(..., onlyConsensusEvents());`
      - The test FAILS 🔴  
    - When the test does NOT have your recent changes (`test.runForCount(100, onlyBFTSyncEvents())`):
      - The test PASSES 🟢 _(no surprise: this is kinda `develop`'s state **before** your PR)_

_(I would put it into a table, but that would be a 3-dimensional table)_

Anyways: the first green dot above explains why this test was passing for me before I updated from develop - it was most likely the #294. Is it possible that it somehow causes some substate updates to be skipped / committed by mistake?

PR #313, opened 2023-02-13
Detect state hash inconsistency before commit

jakrawcz-rdx:
Note: this builds on top of https://github.com/radixdlt/babylon-node/pull/312

Here is an example usage of state hash (to detect engine non-det bugs).


jakrawcz-rdx:
@talekhinezh please let me know if you prefer to `panic!` (instead of `warn!`) on engine non-determinism.

core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/CommitRequest.java
```diff
@@ -77,6 +78,7 @@
 public record CommitRequest(
     List<RawLedgerTransaction> transactions,
     UInt64 stateVersion,
+    HashCode stateHash,
```
dhedey, line 81: I might be tempted to call it `expectedResultantStateHash` or something, but that's also me loving long descriptive names.

Happy to keep with `stateHash` too :D

core-rust/state-manager/src/types.rs
```diff
@@ -480,10 +480,11 @@ pub enum CommitError {
     MissingEpochProof,
 }
 
-#[derive(Debug, Decode, Encode, Categorize)]
+#[derive(Debug, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
```
dhedey, line 483: We should be able to just use `Decode`/`Encode`/`Categorize` here - (and maybe `StateHash` if necessary?).

ScryptoX is only needed for types which require the Scrypto SBOR custom value model (ie things referencing runtime engine types).

dhedey (approved):
Looks good!

core-rust/state-manager/src/types.rs
jakrawcz-rdx: ahh, indeed ✅

PR #317, opened 2023-02-14
Add round change and round duration metrics

LukasGasior1:


core/src/main/java/com/radixdlt/consensus/bft/BFTSyncer.java
```diff
@@ -88,5 +89,5 @@ enum SyncResult {
    * @param author the original author of the qc
    * @return {@code SyncResult.SYNCED} if already synced
    */
-  SyncResult syncToQC(HighQC highQC, @Nullable NodeId author);
+  SyncResult syncToQC(HighQC highQC, @Nullable NodeId author, HighQcSource highQcSource);
```
jakrawcz-rdx, line 92: `if (there are javadoc @params) { update javadoc @params (); }`

Also, it is a good place to indicate that this parameter is not driving any logic, but is only needed for measurement purposes.

core/src/main/java/com/radixdlt/consensus/bft/RoundQuorumReached.java
```diff
@@ -64,54 +64,9 @@
...
-        + "votingResult="
-        + votingResult
-        + ", lastAuthor="
-        + lastAuthor
-        + '}';
-  }
-}
+public record RoundQuorumReached(
+    RoundVotingResult votingResult,
+    Vote lastVote /* the last received vote event that led to forming a quorum */) {}
```
jakrawcz-rdx, line 72: (nitpick) there is a way of documenting this in the record-level javadoc

core/src/main/java/com/radixdlt/consensus/liveness/PacemakerState.java
```diff
@@ -83,49 +86,53 @@
...
       ProposerElection proposerElection,
-      EventDispatcher<RoundUpdate> roundUpdateSender) {
+      EventDispatcher<RoundUpdate> roundUpdateSender,
+      Metrics metrics) {
     this.proposerElection = Objects.requireNonNull(proposerElection);
     this.roundUpdateSender = Objects.requireNonNull(roundUpdateSender);
+    this.metrics = metrics;
     this.highQC = roundUpdate.getHighQC();
     this.currentRound = roundUpdate.getCurrentRound();
+    this.currentRoundStopwatch = Stopwatch.createStarted();
```
jakrawcz-rdx, line 108: (minor) It feels a bit weird to start a stopwatch during app boot-up (I assume that's when we `@Inject`). AFAIU, the first measured duration may be significantly shorter than the round (if validator booted-up mid round)?

It is probably not a huge measurement skew (i.e. boot-ups are rare), but it may be just more elegant to construct the `PacemakerState` with a stopped stopwatch and start it on first round update?

(if this results in some awkward code, please disregard)

core/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
```diff
@@ -240,8 +251,50 @@ public SyncResult syncToQC(HighQC highQC, @Nullable NodeId author) {
...
+                              roundUpdate
+                                  .getLeader()
+                                  .getValidatorAddress()
+                                  .map(addressing::encodeValidatorAddress)
+                                  .orElse("none"),
+                              highQcSource,
+                              certificateType))
+                      .inc();
+                });
+
```
jakrawcz-rdx, line 297: please extract this logic to some method (which makes it clear that its only purpose is to bump a metric)

core/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
```diff
@@ -240,8 +251,50 @@ public SyncResult syncToQC(HighQC highQC, @Nullable NodeId author) {
...
+                  if (tcAlsoInserted) {
+                    // TC was inserted so its round must be higher than that of a QC, round change
+                    // cert type is then a TC
+                    certificateType = CertificateType.TC;
+                  } else {
+                    // QC was inserted, need to determine whether its corresponding vertex was a
+                    // timeout vertex or not
+                    final var isTimeoutVertex =
+                        vertexStore
+                            .getExecutedVertex(qc.getProposedHeader().getVertexId())
```
jakrawcz-rdx, line 271: it looks scary (almost as if we were doing a DB read in order to bump a metric) :P
AFAIU, this only ever goes to a mem store, so is probably not a performance concern. On a conceptual level though, isn't there any other way of obtaining this information from context?

core/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
```diff
@@ -250,7 +303,24 @@ public SyncResult syncToQC(HighQC highQC, @Nullable NodeId author) {
...
+                              new Metrics.RoundChange(
+                                  roundUpdate.getLeader().getKey().toHex(),
+                                  roundUpdate
+                                      .getLeader()
+                                      .getValidatorAddress()
+                                      .map(addressing::encodeValidatorAddress)
+                                      .orElse("none"),
+                                  highQcSource,
+                                  CertificateType.TC))
+                          .inc());
```
jakrawcz-rdx, line 323: (minor) also may be worth extracting a method

common/src/main/java/com/radixdlt/monitoring/Metrics.java
```diff
@@ -281,6 +283,26 @@ public MethodId(Class<?> cls, String methodName) {
     }
   }
 
+  public record RoundChange(
+      String leaderKey,
+      String leaderComponentAddress,
```
jakrawcz-rdx, line 288: just making sure: in all code paths, these 2 strings can only have some well-bounded cardinality, right? (assuming malicious other nodes)

jakrawcz-rdx (commented):
Sorry for nitpicking that hard; I spent my previous 4 years in the metrics industry 🤷

core/src/main/java/com/radixdlt/consensus/liveness/PacemakerState.java
LukasGasior1: yeah, good point, I think it might be better to move this to `Pacemaker` and start the stopwatch in `start`

core/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
LukasGasior1: not really, I guess we could carry the timeout flag over to BFTHeader :thinking: but I think this is out of scope of this PR

LukasGasior1: and yeah, this is just reading from memory

common/src/main/java/com/radixdlt/monitoring/Metrics.java
LukasGasior1: well, not strictly bounded as the validator set can change across epochs, but I don't think it's a concern

jakrawcz-rdx (approved):

dhedey (approved):
So excited to have this in - I might even suggest this (and the liveness fixes?) get merged as a hotfix into `release/betanet-v2` just in case?

A couple of comments regarding where we put code and how we track the leader 👍

core/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
```diff
@@ -275,19 +300,58 @@ public SyncResult syncToQC(HighQC highQC, @Nullable NodeId author) {
...
+  }
+
+  private void updateRoundChangesMetrics(
+      RoundUpdate roundUpdate, CertificateType certificateType, HighQcSource highQcSource) {
+    metrics
+        .bft()
+        .roundChanges()
+        .label(
+            new Metrics.RoundChange(
+                roundUpdate.getLeader().getKey().toHex(),
```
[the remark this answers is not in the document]
LukasGasior1: Yup, that was the leader of the next round. Updated to use the leader of a completed round

dhedey (approved):

common/src/main/java/com/radixdlt/monitoring/Metrics.java
dhedey: Thinking about this a little more, perhaps we should split this metric and remove the leader from this to avoid combinatorial explosion possibilities.

And then we can have a separate metric for round results which we pull from the engine like Olympia - that way we don't miss our own round results if we lose connection for a while and ledger-sync our missed round.

PR #322, opened 2023-02-16
Add a timeout flag to a vote for proposal that took long to compute; …

LukasGasior1:
Notable changes:

* moved all vote-sending logic to Pacemaker (both timeout and non-timeout votes)
* removed RoundTimeoutModerator - moved to Pacemaker
* quorum assembly moved from BFTEventReducer to BFTQuorumAssembler (it's a BFTEventReducer without voting)
* when a timeout trigger occurs during proposal processing, once that's done a vote for the proposal will include a timeout flag and will be broadcasted to everyone
* moved some events round checking logic from BFTSyncUpProcessor to a dedicated ObsoleteEventsFilter (optimization as we don't need to verify any obsolete events)
* some cleanup: moved vertex store (and related classes) to a dedicated package


core/src/main/java/com/radixdlt/consensus/liveness/Pacemaker.java
```diff
@@ -218,84 +203,188 @@ private void startRound() {
...
+      return;
     }
 
-    final var alreadyExecutedVertices =
-        vertexStore.getPathFromRoot(highestQC.getProposedHeader().getVertexId());
-    final var nextTransactions =
-        proposalGenerator.getTransactionsForProposal(round, alreadyExecutedVertices);
-    this.metrics.bft().pacemaker().proposedTransactions().inc(nextTransactions.size());
-    return nextTransactions;
+    this.maybeLatestInsertUpdate = Optional.of(update);
```
dhedey, line 292: I find this `maybeLatestInsertUpdate` confusing - why do we need to store it? Is it only for the future round?

In which case, maybe we separate it into "this.latestInsertUpdateForFutureRound" - only in the case it's for a future round?

And have `this.tryToVoteForTheProposalForTheFirstTime(sendToAll)` become: `this.attemptVoteOnExecutedVertex(BFTInsertUpdate)`, where sendToAll can be read from `this.scheduledRoundTimeoutHasOccurred`?

Similarly, `this.createAndSendTimeoutVoteToAll` and `this.createAndSendTimeoutVote` can be merged into `this.attemptVoteOnExecutedVertex(BFTInsertUpdate)`...

Inside `this.attemptVoteOnExecutedVertex` it also works out whether it should be a timeout vote using `this.scheduledRoundTimeoutHasOccurred` :).

So basically:
* We can vote on a "timeout"/"rejected" vertex without it being a timeout vote (this is why I suggest a rename to "rejection vertex" or "default vertex")
* We add the "timeout" flag to our vote once `this.scheduledRoundTimeoutHasOccurred` and instead send to everyone

core/src/main/java/com/radixdlt/consensus/liveness/Pacemaker.java
```diff
@@ -218,84 +203,188 @@ private void startRound() {
...
+
+    switch (this.roundStatus) {
+      case UNDISTURBED -> {
+        // The round is undisturbed but if the timeout has already
+        // occurred (even if it was prolonged), we'll add a timeout flag to
+        // our initial vote and broadcast it to everyone.
+        this.tryToVoteForTheProposalForTheFirstTime(this.scheduledRoundTimeoutHasOccurred);
+      }
+      case PROPOSAL_REJECTED -> {
+        if (updateIsInsertionOfTimeoutVertex) {
```
dhedey, line 307: I know this is maybe a semi-big change/rename suggestion, but I wonder if we should call it a  `RejectionVertex` or `DefaultVertex`, as it's not just for timeouts any more? :D

IE this would separate the concept of "timeout vote" (goes to everyone, only after local round timeout), and "vertex" to vote on if there's no proposal or you're rejecting the proposal for some reason.

Not necessarily for this PR - just getting your thoughts 👍

dhedey (approved):
Really nice changes, it's becoming so much easier to read/follow with all your comments!

That said, one or two things are still confusing me a little in Pacemaker, and I think a tiny bit more refactoring might give us some simplifications. Have suggested it in a few places, but I dunno what you think :)

I haven't spotted any big issues though, so happy for you to merge once you've dealt with the comments as you see fit

core/src/main/java/com/radixdlt/consensus/liveness/Pacemaker.java
```diff
@@ -218,84 +203,188 @@ private void startRound() {
...
+      return;
+    }
+
+    /* Assuming that the vertex we're about to vote for has either been:
+     a) received in a proposal for the current round
+     b) received in a QC (f.e. along with the next proposal, or during sync)
+    in either case, if the vertex is valid (and at this point we know it is, as it has been inserted),
+    we send our vote for it - if safety rules allow. */
+
+    // TODO: what if insertUpdate occurs before roundUpdate
```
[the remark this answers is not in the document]
dhedey: EDIT: see next comment where this.maybeLatestInsertUpdate is set.

core/src/main/java/com/radixdlt/consensus/bft/processor/SyncUpPreprocessor.java
```diff
@@ -121,26 +121,28 @@ public SyncUpPreprocessor(
     this.latestRoundUpdate = Objects.requireNonNull(initialRoundUpdate);
   }
```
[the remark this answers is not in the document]
LukasGasior1: It doesn't :) Did you mean `ProposalTimestampVerifier`? I've added a comment there

core/src/main/java/com/radixdlt/consensus/liveness/Pacemaker.java
```diff
@@ -218,84 +203,188 @@ private void startRound() {
...
-  private Optional<Proposal> generateProposal(Round round) {
-    final var highQC = this.latestRoundUpdate.getHighQC();
-    final var highestQC = highQC.highestQC();
-    final var nextTransactions = getTransactionsForProposal(round, highestQC);
-    final var proposerTimestamp = determineNextProposalTimestamp(highestQC);
-    final var proposedVertex =
-        Vertex.create(highestQC, round, nextTransactions, self, proposerTimestamp).withId(hasher);
-    return safetyRules.signProposal(
-        proposedVertex, highQC.highestCommittedQC(), highQC.highestTC());
+  private void tryToVoteForTheProposalForTheFirstTime(boolean broadcastToEveryoneAsTimeout) {
```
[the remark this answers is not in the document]
LukasGasior1: That's mostly to decouple the logic from this quite specific flag. In two cases I call this explicitly with true/false for clarity (even though the actual value corresponds with `scheduledRoundTimeoutHasOccurred`)

core/src/main/java/com/radixdlt/consensus/liveness/Pacemaker.java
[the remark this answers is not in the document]
LukasGasior1: tbh, I've just moved this comment over from BFTEventReducer  :sweat_smile: it seems to be handled correctly

core/src/main/java/com/radixdlt/consensus/liveness/Pacemaker.java
LukasGasior1: `FallbackVertex` perhaps?

core/src/test/java/com/radixdlt/consensus/bft/processor/BFTEventReducerTest.java
```diff
@@ -1,201 +0,0 @@
...
-
-    when(update.getHeader()).thenReturn(header);
-    when(header.getRound()).thenReturn(Round.of(2));
-    this.bftEventReducer.processBFTUpdate(update);
-
-    verifyNoMoreInteractions(this.pacemaker);
-  }
-
-  @Test
-  public void when_round_is_timed_out__then_dont_vote() {
```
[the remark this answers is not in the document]
LukasGasior1: Just once of them applies to BFTQuorumAssembler (which I've brought back), others are mostly for the Pacemaker and I believe are already covered (indirectly) by the existing integration tests and the PendingVotesTest

core/src/main/java/com/radixdlt/consensus/liveness/Pacemaker.java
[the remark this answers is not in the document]
LukasGasior1: disregard my above comment, I ended up using the flag :)

core/src/main/java/com/radixdlt/consensus/liveness/Pacemaker.java
LukasGasior1: @dhedey I've refactored it following your ideas (mostly :smile: ), and added some more comments

core/src/main/java/com/radixdlt/consensus/Vertex.java
```diff
@@ -114,9 +114,10 @@ public final class Vertex {
   @DsonOutput(Output.ALL)
   private final List<byte[]> transactions;
 
+  // TODO: rename this JSON property (or remove it entirely?)
```
dhedey, line 117: I guess we're not renaming it now for backwards compatibility of the hotfix?
Could we maybe add a little comment explaining when this flag would be true? 👍

core/src/main/java/com/radixdlt/consensus/liveness/Pacemaker.java
```diff
@@ -280,82 +219,165 @@ public void processBFTUpdate(BFTInsertUpdate update) {
...
-            .filter(update.getInserted().getVertexHash()::equals)
-            .isPresent();
+    maybeBaseVote.ifPresentOrElse(
+        baseVote -> {
+          // A timeout flag is included if:
+          // a) any scheduled timeout event was observed (even if in the end the round was
+          // prolonged)
+          // b) a proposal has been rejected (which implicitly means that this must be a vote on a
+          // fallback vertex)
+          final var includeTimeoutFlag =
```
dhedey, line 281: Out of interest why do we include a timeout flag on the vote (to let people form a TC) in the "PROPOSAL_REJECTED" case? Isn't that already handled by voting on a fallback vertex?

I think it may be safer (from a 3-chain perspective) to hold back the timeout flag for if a timeout has occurred?

core/src/main/java/com/radixdlt/consensus/liveness/Pacemaker.java
```diff
@@ -280,82 +219,165 @@ public void processBFTUpdate(BFTInsertUpdate update) {
...
+        .map(this.safetyRules::timeoutVote)
+        .ifPresentOrElse(
+            this::dispatchVote,
+            () -> voteForFallbackVertexOrInsertIfMissing(this.latestRoundUpdate));
+  }
+
+  /**
+   * A helper for `resendPreviousVoteWithTimeoutOrVoteForFallbackVertex`, see its doc for details
+   */
+  private void voteForFallbackVertexOrInsertIfMissing(RoundUpdate roundUpdate) {
```
dhedey, line 359: I think the `insertIfMissing` is irrelevant its function, so I think it could just be called `voteForFallbackVertex`? 👍

core/src/main/java/com/radixdlt/consensus/liveness/Pacemaker.java
```diff
@@ -280,82 +219,165 @@ public void processBFTUpdate(BFTInsertUpdate update) {
...
+        Vertex.createFallback(
+                highQC.highestQC(), roundUpdate.getCurrentRound(), roundUpdate.getLeader())
+            .withId(hasher);
+
+    this.vertexStore
+        .getExecutedVertex(vertex.hash())
+        .ifPresentOrElse(
+            this::attemptVoteOnVertex,
+            () -> {
+              // Fallback vertex doesn't yet exist, so try inserting one
```
dhedey, line 371: Perhaps it's worth briefly mentioning here that we'll vote on it when we received the BFTInsertUpdate

dhedey (approved):
This is awesome! So much easier to follow for me now 👍.

Thanks for making all these tweaks. I also really like the separation of "fallback vertex" from "timeout vote" / "timeout certificate" - this distinction makes things a lot easier to follow for me :)

core/src/main/java/com/radixdlt/consensus/Vertex.java
LukasGasior1: yup, not including it for compatibility

core/src/main/java/com/radixdlt/consensus/liveness/Pacemaker.java
LukasGasior1: hmh, yeah I guess that makes sense (3-chain perspective), but I feel like there was some reason why we've included the flag, which I can't recall now :P I guess it won't be that much of an issue once we implement "slight delay for processing TC"

dhedey: The concern was roughly if the timestamp was accepted by half of quorum but not the other half, we could end up without being able to form a QC, so we'd need to timeout - so why not include the timeout flag immediately.

But I didn't understand the nuances about fallback vertices and timeout certificates at that point. Essentially it doesn't help, because the timeout for votes on the proposed vertex are only going to be received after the timeout delay anyway.

All that happens is that it gives rise  for the ability of a single validator to cause a TC instead of a QC, before the round's actually timed out.

And causes slightly confusing logic here accounting for it :)

LukasGasior1: haha, okay, fair enough :+1:

PR #323, opened 2023-02-16
Adding a convenience method for easier record codec registration

jakrawcz-rdx:
As in the title: a convenience method built on top of existing `Fields`-style codec.
Allows to register a codec for any standard `record` in a single line, without caring about `TypeToken<>`s or the record's component names.

Note: the impl actually does not do any _custom_ reflection "magic"; it relies 100% on documented guarantees of `record` classes.

### DIFF
--- a/common/src/main/java/com/radixdlt/sbor/codec/StructCodec.java
+++ b/common/src/main/java/com/radixdlt/sbor/codec/StructCodec.java
@@ -64,11 +64,17 @@
 
 package com.radixdlt.sbor.codec;
 
+import com.google.common.reflect.TypeToken;
 import com.radixdlt.lang.Functions;
 import com.radixdlt.sbor.codec.FieldsEncoders.*;
 import com.radixdlt.sbor.codec.constants.TypeId;
 import com.radixdlt.sbor.coding.DecoderApi;
 import com.radixdlt.sbor.coding.EncoderApi;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.RecordComponent;
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Stream;
 
 @SuppressWarnings("unused")
 public interface StructCodec<T> extends Codec<T> {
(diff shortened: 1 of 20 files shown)


common/src/main/java/com/radixdlt/sbor/codec/StructCodec.java
```diff
@@ -417,4 +423,57 @@ static <T, T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12> StructCodec<T> fro
...
+    Function<R, C> componentGetter =
+        record -> {
+          try {
+            return (C) component.getAccessor().invoke(record);
+          } catch (ReflectiveOperationException roe) {
+            throw new IllegalStateException(
+                "failed to invoke %s on %s".formatted(component.getAccessor(), record), roe);
+          }
+        };
+    TypeToken<C> componentType = (TypeToken<C>) recordType.resolveType(component.getGenericType());
```
dhedey, line 475: Ohhhh psych. This is super cool :D - didn't realise that resolve type could do this - it seems to be able to create a TypeToken for the child out of the parent! ✨. Sweet!

dhedey: This works for nested generic types, right?

EG if we had a `SomeFieldType<T>`, and we were encoding `Record<Integer>` where `Record<T>` has a field of `List<SomeFieldType<T>>` then would this work / capture the generic properly as `List<SomeFieldType<Integer>>`? I imagine it would?

dhedey (approved):
Really nice! Thanks 👍 

Can this easily be integrated with Enums (ie sealed interfaces) too? See eg `PublicKey` / `EnumEntry.with`?

common/src/main/java/com/radixdlt/sbor/codec/StructCodec.java
jakrawcz-rdx: I assume you talk about situation where `Record<Integer>` is a field in some other record/class.

Then: yes, all these types will be captured correctly. I mean: they are resolvable, and `.resolveType()` is meant to do it, so if I don't have any bug in this PR, it will work as you expect.

(however, if you meant that `Record<Integer>` is just some runtime type of some var, and you serialize that instance - then no, the type won't be captured. A type param will be captured instead, which will result in an invocation like `codecs.of(new TypeToken<T>() {})`, which will fail. In short: the `.resolveType()` cannot win with the type erasure)

jakrawcz-rdx: the tests currently do
`record SimpleRecord(int first, String second, Either<Long, String> third, Option<Boolean> fourth)`;
this kinda demonstrates the `.resolveType()`'s capabilities already, but AFAIU you would like to change it to e.g.:
`record SimpleRecord(int first, String second, Either<Option<Long>, String> third)`;
for the extra challenge, right?

jakrawcz-rdx:
> Can this easily be integrated with Enums (ie sealed interfaces) too? See eg PublicKey / EnumEntry.with?

Yes. I started some prototyping on that and looks not-too-hardcore, but let's do it in next PR.

Merging this one.

common/src/main/java/com/radixdlt/sbor/codec/StructCodec.java
dhedey: I guess what I'm asking is - can this work with registering `record MyRecord<T1, T2>(T, List<T2>)` via `codecMap.registerForGeneric(MyRecord.class, (codecs, concreteTypeToken) -> StructCodec.fromRecordComponents(concreteTypeToken, codecs));`? 👍 

I think the answer is "yes, it totally can :D".

On a side note, I tend to include contextual arguments (such as `codecs`) first in argument lists, and I see you've put `codecs` last here - I don't mind either way, but maybe consistency might be good 👍

jakrawcz-rdx: ahh yes, it should just work fine with current `.registerForGeneric()` 👍 

pulling the helpers up to `codecMap` level? maaaybe, although you always gotta stop somewhere (for "the right abstraction" reasons), and I think maybe `codecMap` does not need to know stuff about possible impls of codecs.

I will try to remember about "contextual arguments first" 👍 (and migrate the current code next time I touch it).

PR #324, opened 2023-02-16
Adding a convenience method for easier enum (sealed type) codec registration

jakrawcz-rdx:
A follow-up to https://github.com/radixdlt/babylon-node/pull/323#issuecomment-1433418441.

### DIFF
--- a/common/src/main/java/com/radixdlt/crypto/PublicKey.java
+++ b/common/src/main/java/com/radixdlt/crypto/PublicKey.java
@@ -66,24 +66,12 @@ package com.radixdlt.crypto;
 
 import com.radixdlt.sbor.codec.CodecMap;
 import com.radixdlt.sbor.codec.EnumCodec;
-import com.radixdlt.sbor.codec.EnumEntry;
 
 public sealed interface PublicKey {
   static void registerCodec(CodecMap codecMap) {
     codecMap.register(
         PublicKey.class,
-        (codecs) ->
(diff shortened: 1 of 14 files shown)


jakrawcz-rdx:
@dhedey after this PR, very large parts of `EnumEntry, Fields, StructCodec & co` become 3735929054 (base 16).
I would seriously consider removing them (or rather, reducing to a very small and not-necessarily-so-convenient API that allows for implementing a very custom codec if ever really needed). I like deleting code, just give me a 🟢.

dhedey:
> @dhedey after this PR, very large parts of `EnumEntry, Fields, StructCodec & co` become 3735929054 (base 16).
> I would seriously consider removing them (or rather, reducing to a very small and not-necessarily-so-convenient API that allows for implementing a very custom codec if ever really needed). I like deleting code, just give me a 🟢.

I wish I'd realised how easy it would have been to create these codecs with modern Java before creating them all, because man was it painful to write all those variants ;)

As much as it pains me, I'm happy for them to be removed, as long as it's possible to create the custom codecs (but as you say, a slightly less convenient API would be fine then 👍)

jakrawcz-rdx:
> I'm happy for them to be removed

That is going to be a future PR, for some other time when I need break from a more serious task 👍 

Let's merge this one?

dhedey (approved):
Love this! Thanks again for neatening all this up. It's so nice/easy to add codecs now!

PR #325, opened 2023-02-21
Feature: Use Blake2B Hashing

talekhinezh:
* Update `scrypto-lib` to the point where Blake2B Hashing is used
* Remove deprecated `hid` field and `SerializeWithHid` annotation
* Replace use of `sha256twice` with `blake2b256` in all of bablyon-node. This includes:
  * Txn Accumulator Hash
  * LedgerPayload Hash
  * UserPayload Hash
  * Signature Hash
  * Intent Hash


core-rust/core-api-server/src/core_api/conversions/substate.rs
```diff
@@ -220,6 +227,14 @@ pub fn to_api_component_info_substate(
     })
 }
 
+pub fn to_api_package_type_info_substate(
+    _context: &MappingContext,
+    _substate: &TypeInfoSubstate,
+) -> Result<models::Substate, MappingError> {
+    // TODO: Expand this
```
jakrawcz-rdx, line 234: can it be addressed in this PR? or is it blocked by something?

core/src/test/java/com/radixdlt/Blake2BTest.java
```diff
@@ -62,68 +62,21 @@
...
-    assertTrue(obj.has("hid"));
-  }
-
-  @Test
-  public void testDontIncludeHidFieldInJson() {
-    JSONObject output = serialization.toJsonObject(new DummyWithoutHid(), DsonOutput.Output.API);
-    assertFalse(output.has("hid"));
-  }
-
+public class Blake2BTest {
```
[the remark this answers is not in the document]
talekhinezh: Yes, I think git/github is assuming the addition of this file and the removal of `ApiSerializationModifierTest` is a rename. That file was removed intentionally.

core/src/test/java/com/radixdlt/Blake2BTest.java
```diff
@@ -62,68 +62,21 @@
...
-    byte[] output = serialization.toDson(new DummyWithoutHid(), DsonOutput.Output.API);
-    JSONObject obj = serialization.fromDson(output, JSONObject.class);
-    assertFalse(obj.has("hid"));
+  public void testInjectorNotNullToken() {
+    Security.insertProviderAt(new BouncyCastleProvider(), 1);
+    byte[] test = {'a', 'b', 'c'};
+    // var result = new Blake2b.Blake2b256().digest(test);
+    // var hash = HashCode.fromBytes(result);
+    var hash = HashUtils.blake2b256(test, 0, test.length);
+    System.out.println(hash);
```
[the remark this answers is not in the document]
talekhinezh: Oof, good catch, this test wasn't meant to be committed.

core-rust/core-api-server/src/core_api/conversions/substate.rs
talekhinezh: I am hesitant to put too much more work on this part as this substate is most likely going to be refactored/removed soon and is not immediately useful to the end user. Will expand on the comment though to be more clear.

common/src/main/java/com/radixdlt/crypto/Blake2BHashHandler.java
```diff
@@ -62,62 +62,34 @@
...
-import com.radixdlt.crypto.HashUtils;
-import com.radixdlt.serialization.mapper.JacksonCborMapper;
-import java.io.IOException;
+import com.radixdlt.SecurityCritical;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 
-/**
- * A Jackson bean serialization modifier that handles adding the "hid" property to output data for
- * classes annotated with @SerializeWithHid annotation.
```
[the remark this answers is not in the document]
talekhinezh: Yup good point. I'm slowly trying to remove old, deprecated code :)

jakrawcz-rdx:
@talekhinezh this looks very much related, as if some set-up is missing (at least in tests):

image

(https://github.com/radixdlt/babylon-node/actions/runs/4234073067/jobs/7355869933#step:9:1047)

common/src/main/java/com/radixdlt/crypto/ECKeyPair.java
```diff
@@ -137,7 +137,7 @@ public static ECKeyPair fromSeed(byte[] seed) {
     }
 
     try {
-      return fromPrivateKey(HashUtils.blake2b256(seed).asBytes());
+      return fromPrivateKey(HashUtils.sha256Twice(seed).asBytes());
```
jakrawcz-rdx, line 140: I ack that this test does not rely that much on what hash function is used, but for consistency it would still be cool to use blake2b "everywhere"...?

**A)** I thought that the fix would simply be to call a `TestSetupUtils.installBouncyCastleProvider();` (as many other unit tests do). It seems to do the job on my machine.

**B)** Alternatively, we could stop relying on the "security provider" infra for our hashing purposes. I mean, in the `Blake2BHashHandler` we could simply do this instead:
```
  private final ThreadLocal<MessageDigest> blake2BDigester =
      ThreadLocal.withInitial(Blake2b.Blake2b256::new);
```

(then, many `Security.insertProviderAt(new BouncyCastleProvider(), 1);` lines could be dropped; although probably not all, because I see we want to rely on BouncyCastle for SecureRandom)

Please let me know if any of these options make sense here. If they are not worth the hassle, you can also feel free to merge as-is 👍

jakrawcz-rdx (approved):

common/src/main/java/com/radixdlt/crypto/ECKeyPair.java
talekhinezh: Yup, totally agree, was just being lazy :) Will go with option **A**.

PR #326, opened 2023-02-21
Use JRE instead of the whole JDK on the runtime docker image

ghost:
This change reduces the final image size by around 500MB.

### DIFF
--- a/docker/Dockerfile.core
+++ b/docker/Dockerfile.core
@@ -20,16 +20,12 @@ RUN apt-get -y update > /dev/null && \
     apt-get clean > /dev/null && \
     rm -rf /var/lib/apt/lists/* /var/tmp/* /tmp/*
 
-RUN apt-get -y update > /dev/null && \
-    apt-get -y --no-install-recommends install \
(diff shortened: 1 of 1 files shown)


kofrdx (approved):
Simple and effective. Runs through all available test scenarios.

shambupujar:
As discuss in standup, we would need to test this on some network to see there aren't any issues and then merge it

docker/Dockerfile.core
```diff
@@ -22,10 +22,15 @@ RUN apt-get -y update > /dev/null && \
 
 COPY scripts/fix-vulnerabilities.sh /tmp
 
-RUN apt update -y
-RUN apt-get -y --no-install-recommends install openjdk-17-jdk
+RUN apt-get -y update > /dev/null && \
+    apt-get -y --no-install-recommends install \
+    openjdk-17-jre-headless \
```
kofrdx, line 27: Does this need to be seperated from the install command above? Is openjdk-17-jre-headless updated so much more often?

PR #327, opened 2023-02-21
Recording the stale state tree node keys in the persistence

jakrawcz-rdx:
The last PR that will allow to close https://radixdlt.atlassian.net/browse/NODE-515.

Please note that the actual pruning of old-enough state hash tree nodes is still out of scope (and probably up for discussion first).


core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -315,13 +318,21 @@ impl CommitStore for RocksDBStore {
...
+                self.cf_handle(&StateHashTreeNodes),
                 encode_key(&key),
                 scrypto_encode(&node).unwrap(),
             );
         }
+        for stale_node_keys in state_hash_tree_update.stale_node_keys_at_state_version {
+            batch.put_cf(
+                self.cf_handle(&StaleStateHashTreeNodeKeysByStateVersion),
+                stale_node_keys.0.to_be_bytes(),
+                encode_vec(&stale_node_keys.1, encode_key),
```
jakrawcz-rdx, line 333: Note from author:

I chose to store it like this:
```
db_key = state version | db_value = stale node keys
-------------------------------------------------------------------------
                     7 | vec![{v=2; nibbles=a1b}, {v=6; nibbles=a2}]
                     8 | vec![{v=7; nibbles=a1b}]
```
(because that's the way the pruner will most likely want to consume it: "get stale node keys for versions older than current_version-X")

I wonder if there are any downsides of that... because the OTHER open-sourced JMT-using project stores it like this:
```
db_key = node key  | db_value = state version
------------------------------------------------------------
{v=2; nibbles=a1b} | 7
{v=6; nibbles=a2}  | 7
{v=7; nibbles=a1b} | 8
```
(which doesn't have any obvious advantage to me)

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -704,3 +715,19 @@ impl RecoverableVertexStore for RocksDBStore {
         self.db.get_cf(self.cf_handle(&VertexStore), []).unwrap()
     }
 }
+
+/// Encodes a vector of arbitrary items, given a known encoder of a single item.
+/// This is achieved in a straightforward way: by storing an item count and each item's length
+/// before its encoded bytes.
+fn encode_vec<I>(vec: &Vec<I>, item_encoder: fn(&I) -> Vec<u8>) -> Vec<u8> {
```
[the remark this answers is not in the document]
LukasGasior1: I guess we could use sbor (`scrypto_encode`) for consistency?

LukasGasior1 (approved):
small suggestion related to Vec encoding, LGTM otherwise

core-rust/state-manager/src/store/rocks_db.rs
LukasGasior1: seems reasonable to store it by state version :+1:

LukasGasior1: seems reasonable to store it by the state version :+1:

core-rust/state-manager/src/store/rocks_db.rs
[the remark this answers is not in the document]
jakrawcz-rdx: Done 👍  (the way discussed on Slack)

PR #329, opened 2023-02-22
Moving state version tracking out of staging tree, into state manager

jakrawcz-rdx:
This looks like a refactor, but actually fixes a bug.

## The bug:

A booting-up node (either completely new one, or a restarting one) has an empty staging tree. The very first node in this tree would be initialized (wrongfully) with `state_version = 0`, instead of `state_version = db.top_of_ledger().version`.

## The solution:
I could put `= db.top_of_ledger().version` somehow into the staging tree logic, but it turns out easier to just handle the `state_version` at the state manager's layer (which cares about this value anyway, for other reasons).

### DIFF
--- a/core-rust/state-manager/src/state_manager.rs
+++ b/core-rust/state-manager/src/state_manager.rs
@@ -271,14 +271,15 @@ where
 {
     fn execute_for_staging_with_cache(
         &mut self,
-        parent_accumulator_hash: &AccumulatorHash,
+        parent_state: &StateSummary,
         executable: &Executable,
         payload_hash: &LedgerPayloadHash,
     ) -> (AccumulatorHash, &ProcessedResult) {
-        let new_accumulator_hash = parent_accumulator_hash.accumulate(payload_hash);
+        let new_accumulator_hash = parent_state.accumulator_hash.accumulate(payload_hash);
         let processed_result = self.execution_cache.execute_transaction(
             &self.store,
-            parent_accumulator_hash,
+            Some(parent_state.state_version).filter(|version| *version > 0),
+            &parent_state.accumulator_hash,
             &new_accumulator_hash,
             |store| {
                 let start = Instant::now();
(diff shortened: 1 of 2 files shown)


jakrawcz-rdx:
Note:

After this fix, the `recovery_should_work_when_consensus_is_behind_ledger()` starts PASSing even after `LedgerHeader.stateHash` goes into `Output.ALL`.

However, I think we should not do `Output.ALL` just yet.
Alternative options are:
A) stay with just the current `warn!()` on commit (when hashes don't match)
B) change that `warn!()` into a metric bump
C) change that `warn!()` into a `panic!()`

It is probably a question of "how hard we want to be notified about non-det problems" and "should we break liveness on state inconsistencies". 

My vote would currently go to A).
(ofc, after we repro [and fix] a couple of suspected non-det bugs, we can go directly even to `Output.ALL`; but for that, I'd need to understand better what are our suspects...)

@LukasGasior1 @dhedey @talekhinezh what do you think?

dhedey:
@jakrawcz-rdx (A) sounds good to me!

core-rust/state-manager/src/state_manager.rs
```diff
@@ -934,8 +963,13 @@ where
     S: QueryableProofStore + QueryableTransactionStore,
 {
     pub fn commit(&'db mut self, commit_request: CommitRequest) -> Result<(), CommitError> {
+        let commit_transactions_len = commit_request.transaction_payloads.len();
+        if commit_transactions_len == 0 {
+            return Ok(());
```
LukasGasior1, line 968: we should never try to commit 0 txns, I wonder if this should be an error instead?

core-rust/state-manager/src/state_manager.rs
```diff
@@ -890,25 +891,53 @@ where
...
+    pub fn to_committed_identifiers(&self) -> CommittedTransactionIdentifiers {
+        CommittedTransactionIdentifiers {
+            state_version: self.state_version,
+            accumulator_hash: self.accumulator_hash,
+        }
+    }
+}
+
+struct StateTracker {
+    summary: StateSummary,
```
LukasGasior1, line 922: why don't we reuse `CommittedTransactionIdentifiers`? f.e. `latest_transaction_identifiers: CommittedTransactionIdentifiers`?

LukasGasior1 (approved):
Two minor comments, LGTM

core-rust/state-manager/src/state_manager.rs
jakrawcz-rdx: ahh, I had the need for "state summary" (in the logic) and then I didn't notice it is 100% identical to "transaction identifiers" - you are right, no point copying 👍 

(I unified the name of `latest_state_hash()` too)

PR #336, opened 2023-02-27
chore: cleanup core api default body limit

malexandru-rdx:
Update axum's version. Check breaking [changelog](https://github.com/tokio-rs/axum/blob/main/axum/CHANGELOG.md#060-25-november-2022).

Remove Tower `RequestBodyLimitLayer` middleware and use `DefaultBodyLimit::max` (see TODO and https://github.com/tokio-rs/axum/pull/1397).

### DIFF
--- a/core-rust/core-api-server/src/core_api/extractors.rs
+++ b/core-rust/core-api-server/src/core_api/extractors.rs
@@ -1,7 +1,9 @@
 use axum::{
     async_trait,
     body::HttpBody,
-    extract::{rejection::JsonRejection, FromRequest, RequestParts},
+    extract::{rejection::JsonRejection, FromRequest, FromRequestParts},
+    http::request::Parts,
+    http::Request,
     response::IntoResponse,
 };
 use serde::Serialize;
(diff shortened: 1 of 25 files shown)


jakrawcz-rdx:
Locally, I have set `const LARGE_REQUEST_MAX_BYTES: usize = 13;`, and then I run the most basic test I've found (`test_core_api_can_submit_and_commit_transaction()`):

on `develop`:
```
transactionSubmitPost call failed with: 413 - length limit exceeded
com.radixdlt.api.core.generated.client.ApiException: transactionSubmitPost call failed with: 413 - length limit exceeded
	at app//com.radixdlt.api.core.generated.api.TransactionApi.getApiException(TransactionApi.java:80)
	at app//com.radixdlt.api.core.generated.api.TransactionApi.transactionSubmitPostWithHttpInfo(TransactionApi.java:510)
	at app//com.radixdlt.api.core.generated.api.TransactionApi.transactionSubmitPost(TransactionApi.java:488)
	at app//com.radixdlt.api.core.NetworkSubmitTransactionTest.test_core_api_can_submit_and_commit_transaction(NetworkSubmitTransactionTest.java:89)
```

and here:
```
transactionSubmitPost call failed with: 400 - {"error_type":"Basic","code":400,"message":"BytesRejection(FailedToBufferBody(LengthLimitError(LengthLimitError(Error { inner: LengthLimitError }))))"}
com.radixdlt.api.core.generated.client.ApiException: transactionSubmitPost call failed with: 400 - {"error_type":"Basic","code":400,"message":"BytesRejection(FailedToBufferBody(LengthLimitError(LengthLimitError(Error { inner: LengthLimitError }))))"}
	at app//com.radixdlt.api.core.generated.api.TransactionApi.getApiException(TransactionApi.java:80)
	at app//com.radixdlt.api.core.generated.api.TransactionApi.transactionSubmitPostWithHttpInfo(TransactionApi.java:510)
	at app//com.radixdlt.api.core.generated.api.TransactionApi.transactionSubmitPost(TransactionApi.java:488)
	at app//com.radixdlt.api.core.NetworkSubmitTransactionTest.test_core_api_can_submit_and_commit_transaction(NetworkSubmitTransactionTest.java:89)
```

... so the limit works, but the error is surfaced rather differently (out of which 413 would be my preferred).
I think this would justify adding an actual unit test like `throws_413_on_tx_limit_exceeded()`. I imagine this might require a configurable `LARGE_REQUEST_MAX_BYTES`, and if that is too cumbersome, then I can live with just some manual verifications (like above).

core-rust/core-api-server/src/core_api/extractors.rs
```diff
@@ -12,18 +14,35 @@ use super::{client_error, ResponseError};
...
+    S: Send + Sync,
 {
     type Rejection = ResponseError<()>;
 
-    async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
-        match axum::Json::<T>::from_request(req).await {
+    async fn from_request_parts(parts: &mut Parts, state: &S) -> Result<Self, Self::Rejection> {
+        match axum::Json::<T>::from_request_parts(parts, state).await {
+            Ok(value) => Ok(Self(value.0)),
+            Err(rejection) => Err(client_error(format!("{:?}", rejection))),
```
dhedey, line 30: Looks like this is now throwing the error for request size, so we will probably want to handle it properly and turn it into a 403.

Similarly on line 47? Perhaps this logic can be unified into a method?

core-rust/core-api-server/src/core_api/extractors.rs
```diff
@@ -12,18 +14,35 @@ use super::{client_error, ResponseError};
 
 #[derive(Debug)]
 pub(crate) struct Json<T>(pub T);
-pub use axum::Extension; // Re-export Extension so that it can be used easily
+pub use axum::extract::State; // Re-export Extension so that it can be used easily
```
dhedey, line 17: Comment needs updating :)

core-rust/core-api-server/src/core_api/server.rs
```diff
@@ -93,8 +92,6 @@ pub async fn create_server<F>(
 {
     let core_api_state = CoreApiState { state_manager };
 
-    // TODO - Change to remove the Tower RequestBodyLimitLayer middleware and use DefaultBodyLimit::max
```
dhedey, line 96: Can we remove the tower dependency now? Or is it still used elsewhere?
Also could we update LARGE_REQUEST_MAX_BYTES now? 👍  I imagine it can be more like 3 MB now that the max transaction size is about 1MB?

dhedey (approved):
Nice. Thanks for updating things. A few minor tweaks needed before merging :+1:

core-rust/core-api-server/src/core_api/extractors.rs
malexandru-rdx: Fixed. Removed the unused `from_request_parts`.

core-rust/core-api-server/src/core_api/server.rs
malexandru-rdx: Done.

PR #337, opened 2023-02-27
Fix corner case in ledger.prepare when all "previous" vertices were already committed

LukasGasior1:


jakrawcz-rdx (commented):
My knowledge here is rather limited, so I can't really LGTM this based on the code change alone (even if it seems like "just a small, obvious bug" to others).

I see that a `StateComputerLedgerTest` exists and already tests for some corner cases; could you add a test that demonstrates this behaviour change? (ideally failing before)

LukasGasior1:
:+1: I've added a unit test

jakrawcz-rdx (approved):
Nice, thanks for adding the UT 👍

PR #339, opened 2023-02-28
Propagating transaction and receipt hashes up to the ledger header

jakrawcz-rdx:
This PR is a work-in-progress, first part of https://radixdlt.atlassian.net/browse/NODE-537.

Here I only introduce the transaction root hash and receipt root hash into `LedgerHeader` (unified together with state root hash, in `LedgerHashes`).
The actual values of these root hashes are not yet computed, but just placeholders. They are of course not yet used in any logic.

### DIFF
--- a/core-rust/state-manager/src/jni/common_types.rs
+++ b/core-rust/state-manager/src/jni/common_types.rs
@@ -62,22 +62,16 @@
  * permissions under this License.
  */
 
-package com.radixdlt.statecomputer.commit;
+use sbor::{Categorize, Decode, Encode};
 
-import com.radixdlt.lang.Option;
-import com.radixdlt.lang.Tuple;
-import com.radixdlt.sbor.codec.CodecMap;
-import com.radixdlt.sbor.codec.StructCodec;
-import java.util.List;
+#[derive(Debug, PartialEq, Eq, Categorize, Encode, Decode)]
+pub struct JavaHashCode(Vec<u8>);
 
-public record PrepareResult(
-    List<byte[]> committed,
-    List<Tuple.Tuple2<byte[], String>> rejected,
-    Option<NextEpoch> nextEpoch,
-    LedgerHashes ledgerHashes) {
-  public static void registerCodec(CodecMap codecMap) {
-    codecMap.register(
-        PrepareResult.class,
-        codecs -> StructCodec.fromRecordComponents(PrepareResult.class, codecs));
-  }
+impl JavaHashCode {
+    pub fn from_bytes<const S: usize>(bytes: [u8; S]) -> Self {
+        JavaHashCode(bytes.to_vec())
+    }
+    pub fn into_bytes<const S: usize>(self) -> [u8; S] {
+        self.0.as_slice().try_into().unwrap()
+    }
 }
--- a/core-rust/state-manager/src/types.rs
+++ b/core-rust/state-manager/src/types.rs
@@ -63,7 +63,8 @@
  */
 
 use crate::{
-    jni::common_types::JavaHashCode, transaction::LedgerTransaction, LedgerTransactionReceipt,
+    jni::common_types::JavaHashCode, transaction::LedgerTransaction, LedgerTransactionOutcome,
+    LedgerTransactionReceipt, SubstateChanges,
 };
 use radix_engine::blueprints::epoch_manager::Validator;
 use radix_engine::types::*;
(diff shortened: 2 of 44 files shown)


jakrawcz-rdx:
Note:

The approach shown in this PR re-uses the existing per-transaction "staging cache" structure (i.e. adds fields there). This has some pros and cons.

We discussed some alternatives with @malexandru-rdx (refactoring that cache into per-vertex; or duplicating the caching structure to keep transaction tree separately per-vertex). I also had some input from @LukasGasior1, who is a huge fan of completely replacing the accumulator hash with transaction root hash (which means computing it per-transaction is actually required, AFAIU). Please have a look at the related code here, guys, and maybe it will get clearer which approach we want.

core/src/main/java/com/radixdlt/consensus/LedgerHashes.java
```diff
@@ -0,0 +1,160 @@
...
+
+  @JsonProperty("transaction_root")
+  @DsonOutput(DsonOutput.Output.ALL)
+  private final HashCode transactionRoot;
+
+  @JsonProperty("receipt_root")
+  // TODO: restore `Output.ALL` after fixing non-determinism bugs
+  // The test `REv2ConsensusLedgerRecoveryTest.recovery_should_work_when_consensus_is_behind_ledger`
+  // fails when the receipt root is taken into account, which suggests that nodes can end up with
+  // different receipts in some special cases. We need to investigate and fix it.
```
jakrawcz-rdx, line 103: I investigated this, and indeed there are differing receipts of exactly the same transaction!
(in that test, it was a genesis transaction, but I don't think it's a prerequisite)

Luckily, from the diff, it seems we _only_ have a problem with order of items in a `resource_changes` vec:

image

And indeed, here is where it comes from:
image

A HashMap of HashMaps, what can go wrong? :P

Let me fix this.

jakrawcz-rdx: _(it will be a fix in scrypto lib, so this PR still has to go in as-is)_

core-rust/state-manager/src/types.rs
```diff
@@ -200,6 +202,54 @@ impl HasLedgerPayloadHash for NotarizedTransaction {
...
+    ScryptoEncode,
+    ScryptoDecode,
+)]
+pub struct LedgerReceiptHash([u8; Self::LENGTH]);
+
+impl LedgerReceiptHash {
+    pub const LENGTH: usize = 32;
+
+    pub fn for_receipt(receipt: &LedgerTransactionReceipt) -> Self {
+        Self::for_receipt_bytes(&scrypto_encode(receipt).unwrap())
```
jakrawcz-rdx, line 223: Note from author:

It turns out that the current (very direct) hash is not what we want.
As pointed out by @iamyulong, we want to exclude some 🟧 things from merklization:

image
([source](https://rdxworks.slack.com/archives/C01M90Y2448/p1677660105653719?thread_ts=1677656078.643539&cid=C01M90Y2448))

jakrawcz-rdx: in my interpretation, this means that we do NOT want to hash `LedgerTransactionReceipt`'s:
- `fee_summary` (mentioned explicitly on the whiteboard)
- `application_logs` (mentioned explicitly; we don't have a way to filter for "system logs", though)
- `resource_changes` (that's how I interpret the mentioned "resource movement")
- `entity_changes` (not mentioned by the whiteboard, but very similar to `resource_changes` ❓ )
- `next_epoch` (the whiteboard mentions "epoch+round", so this one too ❓ )

This would mean that `LedgerTransactionReceipt` would only hash:
- `outcome` (obviously)
- `substate_changes` (yeah they are important, but our independently calculated substate root hash kinda captures that anyways ❓ )

@iamyulong you said that the whiteboard may be a little outdated; please advice on the "❓" above.

@LukasGasior1 I am curious what you think here.

jakrawcz-rdx: _(I want to resolve it here, since I believe the change belongs in this very PR)_

malexandru-rdx (approved):
Looks good to me! :)

core-rust/state-manager/src/types.rs
iamyulong: My interpretation:
- `entity_changes`: No, can be derived from `entity_changes`
- `next_epoch`: No, will be covered by events (in follow-up PRs)
- `substate_changes`: Yes
- `outcome`: Suggest we start with mapping it to a `bool` for sucess or failure (the error code not designed to be merkelizable)

LukasGasior1: @jakrawcz-rdx yeah, sounds good. The original plan was to split the receipt into "LedgerReceipt" and "ExtendedReceipt", where LedgerReceipt is what validators agree on (and all nodes store it), while ExtendedReceipt would be a container for all other info about the transaction that the node decides to keep (for querying purposes). But I guess that's not something we want to introduce in this PR

jakrawcz-rdx: Thanks for the clarification, PTAL at the latest commit 👍

LukasGasior1: that'll do for now :)

core-rust/state-manager/src/jni/common_types.rs
```diff
@@ -0,0 +1,77 @@
...
+ * logic, functionality, security and appropriateness of using the Work for any commercial
+ * or non-commercial purpose and for any reproduction or redistribution by You of the
+ * Work. You assume all risks associated with Your use of the Work and the exercise of
+ * permissions under this License.
+ */
+
+use sbor::{Categorize, Decode, Encode};
+
+#[derive(Debug, PartialEq, Eq, Categorize, Encode, Decode)]
+pub struct JavaHashCode(Vec<u8>);
```
LukasGasior1, line 68: Not sure why we need this on the Rust side? Why don't we map that directly to `Vec<u8>`? We use `HashCode` on the java side as an immutable wrapper for any byte arrays

LukasGasior1 (approved):
LGTM

core-rust/state-manager/src/jni/common_types.rs
jakrawcz-rdx: I actually did not introduce this type (it previously lived in `mempool.rs` [here](https://github.com/radixdlt/babylon-node/pull/339/files/18cc968b91db281017cd7480609090177cb7b416#diff-e5cb5c996dff03df099b84a17a7dbd51eeff543d873ddd5b58abd7f978ba8163L211), and I moved it to re-use it).

I thought it has to be `JavaHashCode` in order to be mapped to `HashCode` in Java :D but even if not technically needed, I think it's nice to have a well-named type here (not every `Vec<u8>` is conceptually a hash code), assuming it is cheap in runtime.

jakrawcz-rdx:
Thank you everyone for taking a look 🙇‍♂️

PR #343, opened 2023-03-02
refactor: reapply small refactors

mozi-rdx:
These commits are mostly forgotten simplifications, a commit-by-commit review should be an easy thing to do!

### DIFF
--- a/core-rust/state-manager/src/mempool/simple_mempool.rs
+++ b/core-rust/state-manager/src/mempool/simple_mempool.rs
@@ -151,20 +151,17 @@ impl SimpleMempool {
         &mut self,
         intent_hashes: &[IntentHash],
     ) -> Vec<PendingTransaction> {
-        let mut removed_transactions = Vec::new();
-        for intent_hash in intent_hashes {
-            if let Some(payload_hashes) = self.intent_lookup.remove(intent_hash) {
-                for payload_hash in payload_hashes {
-                    let removed_option = self.data.remove(&payload_hash);
-                    if let Some(mempool_data) = removed_option {
-                        removed_transactions.push(mempool_data.transaction);
-                    } else {
-                        panic!("Mempool intent hash lookup out of sync on handle committed");
-                    }
-                }
-            }
-        }
-        removed_transactions
+        intent_hashes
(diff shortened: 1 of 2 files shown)


core-rust/state-manager/src/mempool/simple_mempool.rs
```diff
@@ -151,20 +151,17 @@ impl SimpleMempool {
...
-            }
-        }
-        removed_transactions
+        intent_hashes
+            .into_iter()
+            .filter_map(|intent_hash| self.intent_lookup.remove(intent_hash))
+            .flat_map(|payload_hashes| payload_hashes.into_iter())
+            .map(|payload_hash| {
+                self.data
+                    .remove(&payload_hash)
```
jakrawcz-rdx, line 160: I like the beginning of that simplification, but maybe we should leave the `for` as-is instead of the last `map()`? I mean, like `for x in ...filter_map().flat_map() { self.data.remove(x) }`?

(I am not sure about rust, but in some language it is frowned upon to cause side-effects inside lambdas passed to FP-like `.map()` & co)

jakrawcz-rdx (approved):

core-rust/state-manager/src/mempool/simple_mempool.rs
mozi-rdx: Hmmm, let me check whether it is generally frowned upon, but I like the `map + collect` approach a bit better than the `for { ... .push() }` calls because that way I can think of the whole thing as a pipeline and the conditions of the actual push/collection is a bit more clear for me.
(Please note, that the expression still returns a Vec, but its construction is now a bit more hidden, which is also a thing I like 😅  )

malexandru-rdx: I kinda agree with both comments. I find it easier to read and understand the "pipelined" version but don't like side effects that much.
 
All I want to add is Rust is not FP :). I would let both patterns live along side until there is a stronger consensus.

malexandru-rdx (approved):

LukasGasior1 (approved):
LGTM

PR #345, opened 2023-03-03
update schema hash description

malexandru-rdx:
Update core api schema hashing function.

### DIFF
--- a/core-rust/core-api-server/core-api-schema.yaml
+++ b/core-rust/core-api-server/core-api-schema.yaml
@@ -35,7 +35,7 @@ openapi: 3.1.0
 #
 # Descriptions guidance:
 # => For descriptions of encoded types, we follow this pattern: 'The xyz-encoded, abc-encoded xx of the yy'
-#    EG 'The hex-encoded, SBOR-encoded data' or 'The hex-encoded double-SHA256 hash of the banana payload'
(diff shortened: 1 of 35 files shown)


core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -35,7 +35,7 @@ openapi: 3.1.0
 #
 # Descriptions guidance:
 # => For descriptions of encoded types, we follow this pattern: 'The xyz-encoded, abc-encoded xx of the yy'
-#    EG 'The hex-encoded, SBOR-encoded data' or 'The hex-encoded double-SHA256 hash of the banana payload'
+#    EG 'The hex-encoded, SBOR-encoded data' or 'The hex-encoded Blake2b hash of the banana payload'
```
jakrawcz-rdx, line 38: I am afraid that on the internet, plain "Blake2b" most often means "Blake2b-512" as default :( which would force us to say "Blake2b-256" everywhere, if we want to be unambiguous.

malexandru-rdx: You can get the digest size by looking at `minLength` and `maxLength` for each particular hash but I'll change it for extra clarity.

mozi-rdx (approved):

PR #346, opened 2023-03-03
Cleanup schema and add entity reference to global entity reference

talekhinezh:
Replaced `entity_type` with `entity_reference` in `GlobalEntityReference` in core-api-schema.yaml


malexandru-rdx (approved):
Looks good to me! In light of @PawelPawelec-RDX slack comment, should we start recording schema change logs?

PR #350, opened 2023-03-08
Refactor: passing LedgerProof as a parsable struct rather than opaque bytes

jakrawcz-rdx:
This is something that was apparently postponed via a TODO.
I decided to finally do it, since I plan to use the actual proof's fields in the logic (in particular: I want a state version of an epoch proof during transaction tree update [and later even more fields from every ledger proof, during generation of transaction inclusion proofs])

### DIFF
--- a/core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/AccumulatorState.java
+++ b/core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/AccumulatorState.java
@@ -62,77 +62,17 @@
  * permissions under this License.
  */
 
-package com.radixdlt.consensus;
+package com.radixdlt.statecomputer.commit;
 
-import static java.util.Objects.requireNonNull;
+import com.google.common.hash.HashCode;
+import com.radixdlt.sbor.codec.CodecMap;
+import com.radixdlt.sbor.codec.StructCodec;
+import com.radixdlt.utils.UInt64;
 
-import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonProperty;
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.ImmutableSet;
-import com.radixdlt.consensus.bft.BFTValidator;
-import com.radixdlt.serialization.DsonOutput;
-import com.radixdlt.serialization.SerializerConstants;
-import com.radixdlt.serialization.SerializerDummy;
-import com.radixdlt.serialization.SerializerId2;
-import java.util.Objects;
-import java.util.Set;
-import javax.annotation.concurrent.Immutable;
-
-@Immutable
-@SerializerId2("consensus.next_epoch")
-public final class NextEpoch {
-  @JsonProperty(SerializerConstants.SERIALIZER_NAME)
-  @DsonOutput(value = {DsonOutput.Output.API, DsonOutput.Output.WIRE, DsonOutput.Output.PERSIST})
-  SerializerDummy serializer = SerializerDummy.DUMMY;
-
-  @JsonProperty("validators")
-  @DsonOutput(DsonOutput.Output.ALL)
-  private final ImmutableSet<BFTValidator> validators;
-
-  @JsonProperty("epoch")
-  @DsonOutput(DsonOutput.Output.ALL)
-  private final long epoch;
-
-  @JsonCreator
-  @VisibleForTesting
-  NextEpoch(
-      @JsonProperty(value = "epoch", required = true) long epoch,
-      @JsonProperty(value = "validators", required = true) ImmutableSet<BFTValidator> validators) {
-    this.epoch = epoch;
-    if (epoch < 0) {
-      throw new IllegalArgumentException("Epoch can't be < 0");
-    }
-    this.validators = requireNonNull(validators);
-  }
-
-  public static NextEpoch create(long epoch, Set<BFTValidator> validators) {
-    return new NextEpoch(epoch, ImmutableSet.copyOf(validators));
-  }
-
-  public ImmutableSet<BFTValidator> getValidators() {
-    return validators;
-  }
-
-  public long getEpoch() {
-    return epoch;
-  }
-
-  @Override
-  public boolean equals(Object o) {
-    if (this == o) return true;
-    if (o == null || getClass() != o.getClass()) return false;
-    NextEpoch nextEpoch = (NextEpoch) o;
-    return epoch == nextEpoch.epoch && Objects.equals(validators, nextEpoch.validators);
-  }
-
-  @Override
-  public int hashCode() {
-    return Objects.hash(validators, epoch);
-  }
-
-  @Override
-  public String toString() {
-    return "NextEpoch{" + "validators=" + validators + ", epoch=" + epoch + '}';
+public record AccumulatorState(UInt64 stateVersion, HashCode accumulatorHash) {
+  public static void registerCodec(CodecMap codecMap) {
+    codecMap.register(
+        AccumulatorState.class,
+        codecs -> StructCodec.fromRecordComponents(AccumulatorState.class, codecs));
   }
 }
--- a/core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/CommitRequest.java
+++ b/core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/CommitRequest.java
@@ -64,21 +64,17 @@
 
 package com.radixdlt.statecomputer.commit;
 
-import com.google.common.hash.HashCode;
 import com.radixdlt.lang.Option;
 import com.radixdlt.sbor.codec.CodecMap;
 import com.radixdlt.sbor.codec.StructCodec;
 import com.radixdlt.transactions.RawLedgerTransaction;
-import com.radixdlt.utils.UInt64;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 
 public record CommitRequest(
     List<RawLedgerTransaction> transactions,
-    UInt64 stateVersion,
-    HashCode stateHash,
-    byte[] proofBytes,
+    LedgerProof proof,
     Option<byte[]> postCommitVertexStoreBytes) {
   public static void registerCodec(CodecMap codecMap) {
     codecMap.register(
@@ -92,17 +88,16 @@ public record CommitRequest(
     if (o == null || getClass() != o.getClass()) return false;
     CommitRequest that = (CommitRequest) o;
     return Objects.equals(transactions, that.transactions)
-        && Objects.equals(stateVersion, that.stateVersion)
-        && Objects.equals(stateHash, that.stateHash)
-        && Arrays.equals(proofBytes, that.proofBytes)
-        && Objects.equals(postCommitVertexStoreBytes, that.postCommitVertexStoreBytes);
+        && Objects.equals(proof, that.proof)
+        && Arrays.equals(
+            postCommitVertexStoreBytes.or((byte[]) null),
+            that.postCommitVertexStoreBytes.or((byte[]) null));
   }
 
   @Override
   public int hashCode() {
-    int result = Objects.hash(transactions, stateVersion, stateHash, postCommitVertexStoreBytes);
-    result = 31 * result + Arrays.hashCode(proofBytes);
-    return result;
+    return Objects.hash(
+        transactions, proof, Arrays.hashCode(postCommitVertexStoreBytes.or((byte[]) null)));
   }
 
   @Override
@@ -110,14 +105,10 @@ public record CommitRequest(
     return "CommitRequest{"
         + "transactions="
         + transactions
-        + ", stateVersion="
-        + stateVersion
-        + ", stateHash="
-        + stateHash
-        + ", proofBytes="
-        + Arrays.toString(proofBytes)
+        + ", proof="
+        + proof
         + ", postCommitVertexStoreBytes="
-        + postCommitVertexStoreBytes
+        + postCommitVertexStoreBytes.map(Arrays::toString)
         + '}';
   }
 }
--- a/core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/LedgerHeader.java
+++ b/core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/LedgerHeader.java
@@ -62,77 +62,23 @@
  * permissions under this License.
  */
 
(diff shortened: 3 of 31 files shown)


core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/CommitRequest.java
```diff
@@ -92,32 +88,27 @@ public boolean equals(Object o) {
     if (o == null || getClass() != o.getClass()) return false;
     CommitRequest that = (CommitRequest) o;
     return Objects.equals(transactions, that.transactions)
-        && Objects.equals(stateVersion, that.stateVersion)
-        && Objects.equals(stateHash, that.stateHash)
-        && Arrays.equals(proofBytes, that.proofBytes)
-        && Objects.equals(postCommitVertexStoreBytes, that.postCommitVertexStoreBytes);
```
jakrawcz-rdx, line 98: _(note: this change is unrelated, but I believe it has always been a bug [equality on `byte[]` identity])_

core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/CommitRequest.java
```diff
@@ -92,32 +88,27 @@ public boolean equals(Object o) {
...
+        && Objects.equals(proof, that.proof)
+        && Arrays.equals(
+            postCommitVertexStoreBytes.or((byte[]) null),
+            that.postCommitVertexStoreBytes.or((byte[]) null));
   }
 
   @Override
   public int hashCode() {
-    int result = Objects.hash(transactions, stateVersion, stateHash, postCommitVertexStoreBytes);
```
jakrawcz-rdx, line 103: _(note: this change is unrelated, but I believe it has always been a bug [hash based on `byte[]` identity])_

core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/NextEpoch.java
```diff
@@ -64,14 +64,12 @@
...
 
-import com.google.common.collect.ImmutableMap;
-import com.radixdlt.rev2.ComponentAddress;
 import com.radixdlt.sbor.codec.CodecMap;
 import com.radixdlt.sbor.codec.StructCodec;
 import com.radixdlt.utils.UInt64;
+import java.util.Set;
 
-public record NextEpoch(
-    ImmutableMap<ComponentAddress, ActiveValidatorInfo> validators, UInt64 epoch) {
```
jakrawcz-rdx, line 74: _(note: it may seem weird why this had to change, so: Right now, we pass `NextEpoch` both ways, and it just happens that on Java->Rust way, the `ComponentAddress` can be `null` [not suitable for a map key]. This is demonstrated e.g. in `test_valid_rev2_transaction_passes()`. Maybe it's just the test that should be adjusted?)_

core-rust/state-manager/src/state_manager.rs
```diff
@@ -1005,26 +1001,27 @@ where
...
             let commit_result = match &processed.receipt().result {
                 TransactionResult::Commit(result) => {
-                    if let Some((_, next_epoch)) = result.next_epoch {
+                    if result.next_epoch.is_some() {
                         let is_last = i == (commit_transactions_len - 1);
                         if !is_last {
                             return Err(CommitError::MissingEpochProof);
                         }
-                        // TODO: Use actual result and verify proof validator set matches transaction receipt validator set
-                        epoch_boundary = Some(next_epoch);
```
jakrawcz-rdx, line 1014: _(note: Right now, we no longer have to "find" the `epoch_boundary`, since it does not go into `CommitBundle {...}` anymore, because the entire `proof` goes there [and it contains the `next_epoch` itself].)_

core-rust/state-manager/src/state_manager.rs
```diff
@@ -1005,26 +1001,27 @@ where
...
+                        // TODO: verify that `result.next_epoch == commit_ledger_header.next_epoch`
```
jakrawcz-rdx, line 1009: _(note: I have left the `TODO` because it was here before, but actually I don't think we have to explicitly check that `result.next_epoch` is same as expected - we will soon be checking that the entire receipt is same as expected [when I add the `transaction_receipt_root_hash` to the ledger header], which also includes the `next_epoch` anyways; do I get it right?)_

core/src/test/java/com/radixdlt/rev2/REv2StateComputerTest.java
```diff
@@ -122,15 +121,18 @@ protected void configure() {
...
     var genesis =
-        TransactionBuilder.createGenesisWithNumValidators(
-            1, Decimal.of(1), UInt64.fromNonNegativeLong(10));
+        TransactionBuilder.createGenesisWithNumValidators(1, stake, UInt64.fromNonNegativeLong(10));
     var accumulatorState =
         accumulator.accumulate(initialAccumulatorState, genesis.getPayloadHash());
     var validatorSet =
         BFTValidatorSet.from(
             PrivateKeys.numeric(1)
-                .map(k -> BFTValidator.from(BFTValidatorId.create(k.getPublicKey()), UInt256.ONE))
```
jakrawcz-rdx, line 133: _(note: I am pretty sure that `UInt256.ONE` was a bug here; however, even after fixing it, we cannot add the `check that next_epoch from ledger proof matches the next_epoch returned by RE` logic, because of `null` validator addresses populated by this very line of code)_

jakrawcz-rdx: BTW the same "sloppy mocking" must also be causing `computed state hash at version {} differs from the one in proof` in the `commit()`, which is currently just a `warn!()` - but if we want to have these checks in place one day, we need to rethink the setup of these tests a bit.

core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/AccumulatorState.java
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+package com.radixdlt.statecomputer.commit;
+
+import com.google.common.hash.HashCode;
+import com.radixdlt.sbor.codec.CodecMap;
+import com.radixdlt.sbor.codec.StructCodec;
+import com.radixdlt.utils.UInt64;
+
+public record AccumulatorState(UInt64 stateVersion, HashCode accumulatorHash) {
```
LukasGasior1, line 72: why don't we reuase the existing AccumulatorState class?

core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/CommitRequest.java
LukasGasior1: huh, nice catch

core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/CommitRequest.java
```diff
@@ -92,32 +88,27 @@ public boolean equals(Object o) {
...
-        + stateVersion
-        + ", stateHash="
-        + stateHash
-        + ", proofBytes="
-        + Arrays.toString(proofBytes)
+        + ", proof="
+        + proof
         + ", postCommitVertexStoreBytes="
-        + postCommitVertexStoreBytes
+        + postCommitVertexStoreBytes.map(Arrays::toString)
```
LukasGasior1, line 111: I know it was already there, but we prolly don't want that in toString?

core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/LedgerHeader.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+package com.radixdlt.statecomputer.commit;
+
+import com.radixdlt.lang.Option;
+import com.radixdlt.sbor.codec.CodecMap;
+import com.radixdlt.sbor.codec.StructCodec;
+import com.radixdlt.utils.UInt64;
+
+public record LedgerHeader(
```
LukasGasior1, line 72: similarly, why don't we reuse `com.radixdlt.consensus.LedgerHeader`?

jakrawcz-rdx: image

core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/AccumulatorState.java
jakrawcz-rdx: same reasons as in https://github.com/radixdlt/babylon-node/pull/350#discussion_r1129327181, let's discuss there

core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/LedgerHeader.java
jakrawcz-rdx: It is interesting though.

We are simply moving more and more responsibilities into rust (as planned, it seems). And suddenly rust starts caring about e.g. safety/consistency checks against ledger proof, so it can no longer be opaque to it. It kinda enters the "consensus" territory. But all the classes handling consensus are in their well-separated, java-only part of codebase.

Maybe this is not wrong - we do have some "duplicated" structures, yes, but their purpose is "DTO on java-rust bridge". A purpose quite different than "business logic for consensus purposes". Technically, these classes could evolve independently (e.g. the DTO could keep some of its parts as opaque `byte[]`).

LukasGasior1: Hmh, it feels a bit weird IMO to have those DTOs for domain objects. I don't think they should ever evolve independently.

How about this: we mentally move the "ownership" of those domain objects to Rust and then have their Java counterparts living in `core-rust-bridge`, and use those objects in consensus etc?

jakrawcz-rdx: They kinda already evolve independently, in some details, e.g.:
- our domain objects are annotated with `@JsonProperty` and `@DsonOutput`
- our domain objects sometimes have extra "logic only" fields/caches (e.g. `BFTValidatorSet.totalPower`)
- our bridge DTOs use rust-friendly types (`Option` instead of `Optional`, `UInt64`instead of `long`, and more)

I know it is tempting to have "just one, true, ultimate representation of a `LedgerProof`, everywhere across all layers", but it is sometimes either a trap, or an unattainable ideal :(

LukasGasior1: It's tempting indeed :) and I think all of your concerns could be addressed. Anyway, happy to go with a separate dto for now

core-rust/state-manager/src/state_manager.rs
LukasGasior1: yup, receipt verification should cover that

core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/LedgerHeader.java
jakrawcz-rdx: Ok, let me start a separate PR, in which I try to migrate the existing `LedgerHashes` class (an easy example of a "duplicated" domain vs dto class).

I will move the `com.radixdlt.consensus.LedgerHashes` into `com.radixdlt.statecomputer.commit` (i.e. replacing the existing `record LedgerHashes` from there), and update all the code that refers it. We will see how that looks (and decide whether to move on with heavier `LedgerHeader`, `LedgerProof` etc) 👍 

This PR is unblocked in this regard 🟢

core-rust/state-manager/src/types.rs
```diff
@@ -663,6 +667,38 @@ pub struct PrepareGenesisRequest {
...
-    pub validator_set: Option<BTreeMap<ComponentAddress, Validator>>,
+    pub validator_set: Option<Vec<ActiveValidatorInfo>>,
     pub ledger_hashes: LedgerHashes,
 }
+
+#[derive(Debug, Clone, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
+pub struct TimestampedValidatorSignature {
+    pub key: EcdsaSecp256k1PublicKey,
+    pub validator_address: Option<ComponentAddress>,
+    pub timestamp_ms: i64,
```
LukasGasior1, line 678: how about making it an unsigned int? we could then add some assertions when converting to `long` on the Java side (or simply use `UInt64`).

same for other timestamps below

core/src/test/java/com/radixdlt/rev2/REv2StateComputerTest.java
jakrawcz-rdx: Capturing my Slack chat with @talekhinezh :

- The validator address should remain an `Option<>`: it is simply up to the StateComputer how it wants to represent validators. Currently the engine always returns the address, but this fact should be opaque to the consensus layer. The current shape of this PR reflects that fine 🟢 
- Some tests (e.g. `test_valid_rev2_transaction_passes()`) currently break this abstraction (and on a much larger scale): they try to mock some StateComputer's outputs, and then pass them to actual StateComputer. Like: call `.commit()` with a LedgerProof that did not come from `.prepare()`. Such tests will soon stop working (when we add some more verifications on receipt hash, validator set, etc). We need to adjust them (as suspected in https://github.com/radixdlt/babylon-node/pull/350#discussion_r1129297454). But this PR still avoids the problem (by not adding these checks just yet) 🟢

LukasGasior1 (approved):
LGTM

core-rust/state-manager/src/types.rs
jakrawcz-rdx: I just tried and propagated that change until I hit this:
https://github.com/radixdlt/babylon-node/blob/48b2d2116750e8812e726ef3f80aee3b179364b9/core-rust/state-manager/src/transaction/validator_transaction.rs#L43
(the `ClockSetCurrentTimeInput.current_time_ms` is i64)

Seems like our convention is to have millis in i64.

One fun argument I've heard:
- with u64, you can only represent dates from year 1970 to year 478000000
- while with i64, you can represent dates from year -239000000 to 239000000.
And people choose the latter, since some domains care about retro-populated events that happened before 1970.

The `i64` isn't really limiting anyone in foreseeable future, and a negative value is perfectly interpretable (so, not illegal - for our domain `-1000` is most likely invalid, but `1000` is then invalid too).

PR #355, opened 2023-03-10
Architecture: Update scrypto lib + transaction preview receipt

talekhinezh:
* Update CoreAPI ResourceChanges
* Remove MetadataSubstate from State CoreAPI
* Add encoded_receipt to TransactionPreviewResponse


core-rust/core-api-server/src/core_api/handlers/stream_transactions.rs
```diff
@@ -106,7 +105,7 @@ pub fn to_api_committed_transaction(
     receipt: LedgerTransactionReceipt,
     identifiers: CommittedTransactionIdentifiers,
 ) -> Result<models::CommittedTransaction, MappingError> {
-    let receipt = to_api_receipt(context, receipt)?;
+    let receipt = to_api_receipt(context, receipt).unwrap();
```
jakrawcz-rdx, line 108: what is the reason for this change?

jakrawcz-rdx (commented):
_(minor and unrelated doubt below, feel free to ignore or postpone this discussion)_

I like the change from `format!("{}", x)` to `format!("{x}")` in general, but:
I assume it was applied by some auto-process (spotless?), but I see it was not really applied absolutely everywhere, which is kinda weird (probably because of dot-syntax not yet supported inside `{}`?). Do we want it applied half-way like that? Can we opt-out?

core-rust/core-api-server/src/core_api/handlers/stream_transactions.rs
talekhinezh: Ah good catch, I had this unwrap for debugging reasons. Will remove.

talekhinezh:
> _(minor and unrelated doubt below, feel free to ignore or postpone this discussion)_
> 
> I like the change from `format!("{}", x)` to `format!("{x}")` in general, but: I assume it was applied by some auto-process (spotless?), but I see it was not really applied absolutely everywhere, which is kinda weird (probably because of dot-syntax not yet supported inside `{}`?). Do we want it applied half-way like that? Can we opt-out?

Yup this was done by the auto formatter. I'm not too concerned about half-way application as this is a high priority PR but I do think we should revisit this issue in the future, unless you know of a quick opt-out or format-all fix?

jakrawcz-rdx (approved):
> unless you know of a quick opt-out or format-all fix?

I tried `./gradlew spotlessApply` and it does not touch this aspect of formatting (neither stable nor nightly). So I guess, if it builds, we are fine 👍

dhedey:
> I tried ./gradlew spotlessApply and it does not touch this aspect of formatting (neither stable nor nightly). So I guess, if it builds, we are fine 👍

Sweet. A few versions ago, rust clippy went through a few versions where it kept changing its mind on how to lay out these format lines! Whatever most recent clippy wants to do after `rustup update` I'm happy with 👍

PR #356, opened 2023-03-12
Merger (compactor) impl for accumulator tree slices

jakrawcz-rdx:
This is the next algo-heavy moving part needed for transaction and receipt hash tree.
I decided to extract this isolated part, to make the next PR less scary.

Many `TreeSlice`s (produced on per-transaction basis, due to our staging cache being per-transaction) need to be merged into a single slice (to be associated with an entire ledger header).
In future, when we also start deleting (or, not storing) some ledger headers, we will also need to merge slices coming from consecutive ledger headers into a single slice (e.g. per-epoch).

That's what `AccuTreeSliceMerger` does.
_(not currently used in any logic, only unit-tested)_


malexandru-rdx (commented):
I don't think I get why it's been chosen to be implemented like this. I was under the impression that we are no longer going to use this structure in prepare and only when actually committing. For that case, why don't we use the bulk approach? I've benched it and it's 100x speed up from 1 transaction to 10 transactions batches - this does not include the merging afterwards.

LE: this comment is kinda stupid. We need the QC sign over the LedgerHashes so it is used in prepare (and staging) - Idk why I had that impression.

jakrawcz-rdx:
@malexandru-rdx your comment above (and our discussion on tree storage scheme) resulted in some real dilemma right now at https://radixdlt.atlassian.net/wiki/spaces/RPNV1/pages/2996600937/RNP-7+-+Transaction+Hash+Tree+Inclusion+Exclusion+verification#Tree-storage-options-comparison :D (I recommend reading the comments there too) 

Anyways, as I declared, migrating from per-slice to per-node storage (precisely in this direction, I mean) is quite straightforward (it reduces complexity) and can be done at any moment. TBH, even if we had that decision right now, I'd still prefer to do this migration later (as a pure refactor), after fully integrating this feature. Do you think we can progress this way?

PR #357, opened 2023-03-13
Calling the actual accu tree for computing transaction and receipt hashes

jakrawcz-rdx:
⚠️ This relies on https://github.com/radixdlt/babylon-node/pull/356

This is an "integration" part of https://radixdlt.atlassian.net/browse/NODE-537 (the tx/rx hashes were exposed in our `LedgerHeader` earlier [as mocks], and the merkle tree builder was implemented earlier [with only UT usage], and this PR marries them).

In this PR, the storage approach follows the "per-slice" option (A) described at https://radixdlt.atlassian.net/wiki/spaces/RPNV1/pages/2996600937/RNP-7+-+Transaction+Hash+Tree+Inclusion+Exclusion+verification#Option-A)-Per-slice-storage. The migration to Option B) is easy to perform as a standalone refactor later, should we decide so.

### DIFF
--- a/core-rust/core-api-server/src/core_api/handlers/status_network_status.rs
+++ b/core-rust/core-api-server/src/core_api/handlers/status_network_status.rs
@@ -17,10 +17,6 @@ pub(crate) fn handle_status_network_status_internal(
     request: models::NetworkStatusRequest,
 ) -> Result<models::NetworkStatusResponse, ResponseError<()>> {
     assert_matching_network(&request.network, &state_manager.network)?;
-
-    let pre_genesis =
-        to_api_committed_state_identifier(CommittedTransactionIdentifiers::pre_genesis())?;
-
     Ok(models::NetworkStatusResponse {
         post_genesis_state_identifier: state_manager
             .store()
@@ -29,17 +25,12 @@ pub(crate) fn handle_status_network_status_internal(
                 Ok(Box::new(to_api_committed_state_identifier(identifiers)?))
             })
             .transpose()?,
-        current_state_identifier: Box::new(
-            state_manager
-                .store()
-                .get_top_of_ledger_transaction_identifiers()
-                .map(|identifiers| -> Result<_, MappingError> {
-                    to_api_committed_state_identifier(identifiers)
-                })
-                .transpose()?
-                .unwrap_or_else(|| pre_genesis.clone()),
-        ),
-        pre_genesis_state_identifier: Box::new(pre_genesis),
+        current_state_identifier: Box::new(to_api_committed_state_identifier(
+            state_manager.store().get_top_transaction_identifiers(),
+        )?),
+        pre_genesis_state_identifier: Box::new(to_api_committed_state_identifier(
+            CommittedTransactionIdentifiers::pre_genesis(),
+        )?),
     })
 }
 
--- a/core-rust/state-manager/src/jni/transaction_store.rs
+++ b/core-rust/state-manager/src/jni/transaction_store.rs
@@ -177,13 +177,10 @@ extern "system" fn Java_com_radixdlt_transaction_REv2TransactionAndProofStore_ge
 }
 
 #[tracing::instrument(skip_all)]
-fn do_get_epoch_proof(
-    state_manager: &ActualStateManager,
-    state_version: u64,
-) -> Option<JavaLedgerProof> {
+fn do_get_epoch_proof(state_manager: &ActualStateManager, epoch: u64) -> Option<JavaLedgerProof> {
     state_manager
         .store()
-        .get_epoch_proof(state_version)
+        .get_epoch_proof(epoch)
         .map(|proof| proof.into())
 }
 
--- a/core-rust/state-manager/src/staging/mod.rs
+++ b/core-rust/state-manager/src/staging/mod.rs
@@ -63,6 +63,41 @@
  */
 
 mod cache;
+mod result;
 mod stage_tree;
 
+use crate::accumulator_tree::storage::ReadableAccuTreeStore;
+use crate::{ReceiptHash, TransactionHash};
+use radix_engine::ledger::ReadableSubstateStore;
+use radix_engine_interface::api::types::SubstateOffset;
+use radix_engine_stores::hash_tree::tree_store::{ReNodeModulePayload, ReadableTreeStore};
+
 pub use cache::*;
+pub use result::*;
+
+pub trait ReadableLayeredTreeStore:
+    ReadableTreeStore<ReNodeModulePayload> + ReadableTreeStore<SubstateOffset>
+{
+}
+impl<T: ReadableTreeStore<ReNodeModulePayload> + ReadableTreeStore<SubstateOffset>>
+    ReadableLayeredTreeStore for T
+{
+}
+
+pub trait ReadableTxAndRxTreeStore:
+    ReadableAccuTreeStore<u64, TransactionHash> + ReadableAccuTreeStore<u64, ReceiptHash>
+{
+}
+impl<T: ReadableAccuTreeStore<u64, TransactionHash> + ReadableAccuTreeStore<u64, ReceiptHash>>
+    ReadableTxAndRxTreeStore for T
+{
+}
+
+pub trait RootStore:
+    ReadableSubstateStore + ReadableLayeredTreeStore + ReadableTxAndRxTreeStore
+{
+}
+impl<T: ReadableSubstateStore + ReadableLayeredTreeStore + ReadableTxAndRxTreeStore> RootStore
+    for T
+{
+}
--- a/core-rust/state-manager/src/staging/mod.rs
+++ b/core-rust/state-manager/src/staging/mod.rs
@@ -75,31 +75,30 @@ use radix_engine_stores::hash_tree::tree_store::{ReNodeModulePayload, ReadableTr
 pub use cache::*;
 pub use result::*;
 
-pub trait ReadableLayeredTreeStore:
+pub trait ReadableStateTreeStore:
     ReadableTreeStore<ReNodeModulePayload> + ReadableTreeStore<SubstateOffset>
 {
 }
-impl<T: ReadableTreeStore<ReNodeModulePayload> + ReadableTreeStore<SubstateOffset>>
-    ReadableLayeredTreeStore for T
+impl<T> ReadableStateTreeStore for T where
+    T: ReadableTreeStore<ReNodeModulePayload> + ReadableTreeStore<SubstateOffset>
 {
 }
 
-pub trait ReadableTxAndRxTreeStore:
+pub trait ReadableTransactionAndReceiptTreeStore:
     ReadableAccuTreeStore<u64, TransactionTreeHash> + ReadableAccuTreeStore<u64, ReceiptTreeHash>
 {
 }
-impl<
-        T: ReadableAccuTreeStore<u64, TransactionTreeHash>
-            + ReadableAccuTreeStore<u64, ReceiptTreeHash>,
-    > ReadableTxAndRxTreeStore for T
+impl<T> ReadableTransactionAndReceiptTreeStore for T where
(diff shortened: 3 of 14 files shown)


core-rust/core-api-server/src/core_api/handlers/status_network_status.rs
```diff
@@ -29,17 +25,12 @@ pub(crate) fn handle_status_network_status_internal(
                 Ok(Box::new(to_api_committed_state_identifier(identifiers)?))
             })
             .transpose()?,
-        current_state_identifier: Box::new(
-            state_manager
-                .store()
-                .get_top_of_ledger_transaction_identifiers()
```
jakrawcz-rdx, line 35: _Note: I now use this function more seriously, so I made the impl explicit (and touched the API a bit)_

core-rust/state-manager/src/jni/transaction_store.rs
```diff
@@ -177,13 +177,10 @@ extern "system" fn Java_com_radixdlt_transaction_REv2TransactionAndProofStore_ge
...
 
 #[tracing::instrument(skip_all)]
-fn do_get_epoch_proof(
-    state_manager: &ActualStateManager,
-    state_version: u64,
-) -> Option<JavaLedgerProof> {
+fn do_get_epoch_proof(state_manager: &ActualStateManager, epoch: u64) -> Option<JavaLedgerProof> {
     state_manager
         .store()
-        .get_epoch_proof(state_version)
```
jakrawcz-rdx, line 186: _Note: pretty sure it was a typo_

core-rust/state-manager/src/staging/result.rs
```diff
@@ -0,0 +1,360 @@
...
+use radix_engine_stores::hash_tree::tree_store::{
+    NodeKey, Payload, ReNodeModulePayload, ReadableTreeStore, TreeNode, WriteableTreeStore,
+};
+use radix_engine_stores::hash_tree::{put_at_next_version, SubstateHashChange};
+
+lazy_static! {
+    static ref EMPTY_STATE_DIFF: StateDiff = StateDiff::new();
+}
+
+pub struct ProcessedResult {
```
jakrawcz-rdx, line 87: _Note: this `ProcessedResult` is not a new thing - I moved it from `cache.rs`, because it started growing and actually cache is not interested in its internals.

In general, I also started a little refactoring here, which is supposed to split `ProcessedResult` into `Success` vs `Failure` enum - this will simplify a few callers, and allow to get rid of a few `.clone()`s. I prefer to do it in next PR._

core-rust/state-manager/src/store/db.rs
```diff
@@ -258,6 +265,18 @@ impl TransactionIndex<&LedgerPayloadHash> for StateManagerDatabase {
     }
 }
 
+impl TransactionIdentifierLoader for StateManagerDatabase {
+    fn get_top_transaction_identifiers(&self) -> CommittedTransactionIdentifiers {
+        match self {
+            StateManagerDatabase::InMemory(store) => store.get_top_transaction_identifiers(),
+            StateManagerDatabase::RocksDB(store) => store.get_top_transaction_identifiers(),
+            // This is a special case, where in tests, the Core API needs to construct a state
+            // manager with no database (stuck at pre-genesis).
```
jakrawcz-rdx, line 274: _Note: literally 1 test (`test_core_api_server_start_and_stop()`) needs this special-casing - can we avoid it?_

core/src/main/java/com/radixdlt/consensus/LedgerHashes.java
```diff
@@ -97,6 +97,8 @@ public final class LedgerHashes {
   private final HashCode transactionRoot;
 
   @JsonProperty("receipt_root")
+  // TODO(resolve during code review): this passes our tests, but shouldn't we exclude it from
+  // consensus currently as well? (see the `stateRoot` above)
```
jakrawcz-rdx, line 101: ⚠️ This is a very serious question @dhedey @talekhinezh - do we expect non-det only in "resulting substate store", or in the receipt (or rather, its [hashable part](https://github.com/radixdlt/babylon-node/blob/d3096573713051e2fd6c6b60f046d08cfda1c751/core-rust/state-manager/src/types.rs#L221)) too? (I'd prefer to be cautious)

core-rust/state-manager/src/store/db.rs
malexandru-rdx: This question came up before. Probably we should just ditch it (and the test as well).

jakrawcz-rdx: https://github.com/radixdlt/babylon-node/blob/b464c5594d1dc2dfd6dc1eb823d6e9f4a804d88e/core/src/test/java/com/radixdlt/api/CoreApiServerTest.java#L81

I guess we are at "some point" right now!
Would it also make `DatabaseConfig.NONE` deletable?

jakrawcz-rdx:
@malexandru-rdx I didn't want this huge PR to grow even larger, so I put the missing integration test here: https://github.com/radixdlt/babylon-node/pull/361

malexandru-rdx (approved):
Nice work! Looks good to me :)

core-rust/state-manager/src/jni/transaction_store.rs
LukasGasior1: I don't think it was a typo, _epoch_ proofs are keyed by _epoch_

jakrawcz-rdx: yes, and the code before this PR said `state_version`. I change it to `epoch` in this PR.

core-rust/state-manager/src/types.rs
```diff
@@ -513,6 +513,8 @@ impl From<Hash> for TransactionHash {
     }
 }
 
+impl IsHash for TransactionHash {}
```
LukasGasior1, line 516: Hmh, I'm not a big fan of this naming convention (TransactionHash, ReceiptHash). TransactionHash sounds like it was "a hash of a transaction", while in reality it's "a hash of a node in the transactions merkle tree". How about `TransactionTreeHash` or `TxTreeHash` for short? or `TransactionsHash`?

LukasGasior1 (approved):
Just a minor comment, but looks good overall :+1:

core-rust/state-manager/src/store/db.rs
LukasGasior1: :+1: for removing `CoreApiServerTest` and `DatabaseConfig.NONE`

core-rust/state-manager/src/types.rs
jakrawcz-rdx: I agree, `TransactionTreeHash. ReceiptTreeHash` sound super unambiguous 👍

core-rust/state-manager/src/staging/mod.rs
```diff
@@ -63,6 +63,41 @@
...
+pub trait ReadableTxAndRxTreeStore:
+    ReadableAccuTreeStore<u64, TransactionHash> + ReadableAccuTreeStore<u64, ReceiptHash>
+{
+}
+impl<T: ReadableAccuTreeStore<u64, TransactionHash> + ReadableAccuTreeStore<u64, ReceiptHash>>
+    ReadableTxAndRxTreeStore for T
+{
+}
+
+pub trait RootStore:
```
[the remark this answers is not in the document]
jakrawcz-rdx: I don't have strong opinion myself, the `ReadableStore` looks wonderful and I will assume it does not clash with any other notion we have :P

core-rust/state-manager/src/staging/mod.rs
```diff
@@ -63,6 +63,43 @@
...
+    ReadableLayeredTreeStore for T
+{
+}
+
+pub trait ReadableTxAndRxTreeStore:
+    ReadableAccuTreeStore<u64, TransactionTreeHash> + ReadableAccuTreeStore<u64, ReceiptTreeHash>
+{
+}
+impl<
+        T: ReadableAccuTreeStore<u64, TransactionTreeHash>
```
dhedey, line 92: What does Accu stand for? Could we maybe give it a clearer name?

core-rust/state-manager/src/staging/result.rs
```diff
@@ -0,0 +1,360 @@
...
+
+    pub fn receipt(&self) -> &TransactionReceipt {
+        &self.transaction_receipt
+    }
+
+    pub fn state_diff(&self) -> &StateDiff {
+        Self::extract_state_diff(&self.transaction_receipt)
+    }
+
+    pub fn commit(&self) -> &ProcessedTransactionCommit {
```
dhedey, line 152: `commit` sounds like a verb to me. Perhaps `commit_details` or `get_commit_details` is clearer to me?

core/src/main/java/com/radixdlt/consensus/LedgerHashes.java
dhedey: The receipt is just as likely to be non-det as the substate changes, so for now, let's not include it in consensus.

We can add them both to consensus at a similar time

dhedey (approved):
Looks good - some nice refactors on the side too.

Most of my comments are bike shedding over names. The code looks good, but I was struggling to follow the PR because of the names (and no Rust docs).

core-rust/state-manager/src/store/db.rs
jakrawcz-rdx: I think it's just this one test ATM.

Thanks for 🟢 , this will happen in its own PR.

core-rust/state-manager/src/staging/mod.rs
```diff
@@ -63,6 +63,41 @@
...
+impl<T: ReadableTreeStore<ReNodeModulePayload> + ReadableTreeStore<SubstateOffset>>
+    ReadableLayeredTreeStore for T
+{
+}
+
+pub trait ReadableTxAndRxTreeStore:
+    ReadableAccuTreeStore<u64, TransactionHash> + ReadableAccuTreeStore<u64, ReceiptHash>
+{
+}
+impl<T: ReadableAccuTreeStore<u64, TransactionHash> + ReadableAccuTreeStore<u64, ReceiptHash>>
```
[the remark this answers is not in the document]
jakrawcz-rdx: already done at https://github.com/radixdlt/babylon-node/pull/357#discussion_r1137241458 👍

core-rust/state-manager/src/staging/mod.rs
```diff
@@ -63,6 +63,41 @@
...
+    ReadableTreeStore<ReNodeModulePayload> + ReadableTreeStore<SubstateOffset>
+{
+}
+impl<T: ReadableTreeStore<ReNodeModulePayload> + ReadableTreeStore<SubstateOffset>>
+    ReadableLayeredTreeStore for T
+{
+}
+
+pub trait ReadableTxAndRxTreeStore:
+    ReadableAccuTreeStore<u64, TransactionHash> + ReadableAccuTreeStore<u64, ReceiptHash>
```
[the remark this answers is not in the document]
jakrawcz-rdx: whoa, whoa, this one actually has a rust doc :D

In its original PR there is an explanation where the "accumulator tree" name came from: https://github.com/radixdlt/babylon-node/pull/347#discussion_r1128008100

I still think it is a pretty good name to adopt from there. Especially since we've been discussing replacing our accumulator hash with that tree.

I guess you want to expand the name into `ReadableAccumulatorTreeStore`...?

core-rust/state-manager/src/staging/mod.rs
jakrawcz-rdx: let's discuss only at https://github.com/radixdlt/babylon-node/pull/357#discussion_r1137227051

core-rust/state-manager/src/staging/result.rs
dhedey: Ahh understood! That makes much more sense :D. Indeed we only need to care about Commits here I imagine.

PR #361, opened 2023-03-14
Integration test for transaction accu tree / root hash

jakrawcz-rdx:
⚠️ This merges into https://github.com/radixdlt/babylon-node/pull/357

This is a much-needed integration test, which should assure the reviewers that the integration PR actually works.

### DIFF
--- a/core/src/test/java/com/radixdlt/api/core/TransactionAccuTreeTest.java
+++ b/core/src/test/java/com/radixdlt/api/core/TransactionAccuTreeTest.java
@@ -78,9 +78,11 @@ import org.junit.Test;
 
 public class TransactionAccuTreeTest extends DeterministicCoreApiTestBase {
 
+  private static final int EPOCH_TRANSACTION_LENGTH = 4;
+
   @Test
   public void stateManagerMaintainsCorrectTransactionMerkleTree() {
-    try (var test = buildRunningServerTest(6)) {
+    try (var test = buildRunningServerTest(EPOCH_TRANSACTION_LENGTH)) {
       // Run the setup until 2 epoch proofs are captured
       var reader = test.getInstance(0, REv2TransactionAndProofStore.class);
       test.runUntilState(nodeAt(0, NodePredicate.atOrOverEpoch(1)), 1000);
@@ -98,7 +100,7 @@ public class TransactionAccuTreeTest extends DeterministicCoreApiTestBase {
               .toList();
 
       // Assert a certain count (on which we rely during latter manual merkle computation)
-      assertThat(epochTransactions.size()).isEqualTo(6);
+      assertThat(epochTransactions.size()).isEqualTo(EPOCH_TRANSACTION_LENGTH);
 
       // Capture the transaction hashes
       var transactionHashes =
@@ -111,11 +113,12 @@ public class TransactionAccuTreeTest extends DeterministicCoreApiTestBase {
           .isEqualTo(
               merkle(
                   merkle(
+                      // previous epoch's root hash as first leaf
                       merkle(genesisHeader.hashes().transactionRoot(), transactionHashes[0]),
                       merkle(transactionHashes[1], transactionHashes[2])),
                   merkle(
-                      merkle(transactionHashes[3], transactionHashes[4]),
-                      merkle(transactionHashes[5], HashUtils.zero256()))));
+                      merkle(transactionHashes[3], HashUtils.zero256()), // placeholder hash at leaf
+                      HashUtils.zero256()))); // placeholder hash at internal node
 
       // TODO: currently, we cannot test the same for receipt hashes, because the returned receipt
       // bytes contain all ledger receipt fields, while the state manager only hashes the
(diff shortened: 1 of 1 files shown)


malexandru-rdx (approved):

core/src/test/java/com/radixdlt/api/core/TransactionAccuTreeTest.java
```diff
@@ -0,0 +1,134 @@
...
+import com.radixdlt.lang.Option;
+import com.radixdlt.transaction.REv2TransactionAndProofStore;
+import java.util.stream.LongStream;
+import org.junit.Test;
+
+public class TransactionAccuTreeTest extends DeterministicCoreApiTestBase {
+
+  @Test
+  public void stateManagerMaintainsCorrectTransactionMerkleTree() {
+    try (var test = buildRunningServerTest(6)) {
```
dhedey, line 83: I think it might be worth taking out an EPOCH_TRANSACTION_LENGTH = 6 constant to make this clearer.

core/src/test/java/com/radixdlt/api/core/TransactionAccuTreeTest.java
```diff
@@ -0,0 +1,134 @@
...
+      assertThat(epochTransactions.size()).isEqualTo(6);
+
+      // Capture the transaction hashes
+      var transactionHashes =
+          epochTransactions.stream()
+              .map(transaction -> HashUtils.blake2b256(transaction.transactionBytes()))
+              .toArray(HashCode[]::new);
+
+      // Assert that header's root hash is equal to manually computed one
+      assertThat(epochHeader.hashes().transactionRoot())
```
dhedey, line 110: Out of interest, what half-empty tree abstraction do we use? Are all empty/childless nodes given HashUtils.zero256()?

core/src/test/java/com/radixdlt/api/core/TransactionAccuTreeTest.java
```diff
@@ -0,0 +1,134 @@
...
+import com.google.common.hash.HashCode;
+import com.radixdlt.api.DeterministicCoreApiTestBase;
+import com.radixdlt.crypto.HashUtils;
+import com.radixdlt.harness.predicates.NodePredicate;
+import com.radixdlt.lang.Option;
+import com.radixdlt.transaction.REv2TransactionAndProofStore;
+import java.util.stream.LongStream;
+import org.junit.Test;
+
+public class TransactionAccuTreeTest extends DeterministicCoreApiTestBase {
```
dhedey, line 79: I don't think this actually needs to be a Core API test? It could probably just be a Rev2 test?

But I'm happy with it living here, particularly as it means if/when we add proofs to the Core API, it will work.

core/src/test/java/com/radixdlt/api/core/TransactionAccuTreeTest.java
```diff
@@ -0,0 +1,134 @@
...
+                      merkle(genesisHeader.hashes().transactionRoot(), transactionHashes[0]),
+                      merkle(transactionHashes[1], transactionHashes[2])),
+                  merkle(
+                      merkle(transactionHashes[3], transactionHashes[4]),
+                      merkle(transactionHashes[5], HashUtils.zero256()))));
+
+      // TODO: currently, we cannot test the same for receipt hashes, because the returned receipt
+      // bytes contain all ledger receipt fields, while the state manager only hashes the
+      // "merklizable" part.
+    }
```
dhedey, line 123: Do you think it's worth us checking for different scenarios which might result in different sized slices? Or is that checked well by the slice logic tests?

dhedey (approved):
LGTM, thanks for adding it.

core/src/test/java/com/radixdlt/api/core/TransactionAccuTreeTest.java
jakrawcz-rdx: honestly, it had that super-convenient `buildRunningServerTest(int roundsPerEpoch)` and was the easiest one to copy-paste for my first int test :D

I really expect proofs in the Core API, so this will indeed become not only convenient but also appropriate 👍

core/src/test/java/com/radixdlt/api/core/TransactionAccuTreeTest.java
jakrawcz-rdx: I think the unit tests are pretty detailed, starting from here:
https://github.com/radixdlt/babylon-node/blob/5e81ae11414f203e24f46de57ac8bf8bf3da0976/core-rust/state-manager/src/accumulator_tree/test.rs#L102-L104

They include scenarios with slices containing >1 leafs, and even a scenario where 1000 leaves are appended in 1 slice vs in 1000 slices, and result in the same root hash (`incrementally_built_tree_has_same_hash_as_directly_built()`).

In integration tests I am only interested in happy-path.

core/src/test/java/com/radixdlt/api/core/TransactionAccuTreeTest.java
jakrawcz-rdx: The placeholder hash "cuts" the tree branch short (not going all the way to the leaves).
I have now updated the test data to demonstrate that 👍

PR #367, opened 2023-03-19
update: Scrypto updated to release/rcnet-v1

dhedey:
Merges (almost) latest Scrypto.

Many more substate changes than I was hoping for - will document in due course for Gateway.


core-rust/core-api-server/src/core_api/conversions/receipt.rs
```diff
@@ -170,55 +170,59 @@ pub fn to_api_next_epoch(
...
-                let payment = models::RoyaltyPayment {
-                    royalty_receiver: Box::new(
-                        to_global_entity_reference(context, &global_address).ok()?,
-                    ),
-                    cost_unit_amount: to_api_u32_as_i64(cost_unit_amount),
-                };
-                Some(payment)
-            })
-            .collect(),
+        // TODO: Add these back when the (I think accidental?) removal in this PR is reverted:
```
jakrawcz-rdx, line 189: so I guess this is happening much later?
(just wanted to confirm this isn't omission)

core-rust/core-api-server/src/core_api/handlers/status_network_configuration.rs
```diff
@@ -63,84 +64,69 @@ const ALL_ENTITY_TYPES: [EntityType; 13] = [
...
     models::AddressType {
         hrp_prefix: hrp_set.get_entity_hrp(&entity_type).to_string(),
-        subtype,
         entity_type: api_entity_type,
+        subtype,
         address_byte_prefix: entity_type.id().into(),
-        address_byte_length: address_length
-            .try_into()
-            .expect("address was longer than expected"),
+        address_byte_length: ADDRESS_LENGTH.try_into().unwrap(),
```
jakrawcz-rdx, line 130: I see that `ADDRESS_LENGTH` is 27, but an example `extract_length(PackageAddress::Normal)` (i.e. the way it was done before this PR) returns 26 for me. 

Is this intentional, or did you mean `ADDRESS_HASH_LENGTH`?

core-rust/core-api-server/src/core_api/helpers.rs
```diff
@@ -45,6 +45,24 @@ pub(crate) fn read_known_substate(
     )
 }
 
+#[tracing::instrument(skip_all)]
+pub(crate) fn read_known_substate_from_id(
```
jakrawcz-rdx, line 49: (nitpick)

```suggestion
pub(crate) fn read_mandatory_substate_from_id(
```

or just `.read_substate_from_id()`? to have a clear distinction against the `read_optional_...()` family.

core-rust/state-manager/src/state_manager.rs
```diff
@@ -282,7 +279,7 @@ impl<S: ReadableStore> StateManager<S> {
                     store,
                     &self.scrypto_interpreter,
                     &self.fee_reserve_config,
-                    &self.execution_config,
+                    if parent_transaction_identifiers.state_version == 0 { &self.execution_config_for_genesis } else { &self.execution_config },
```
jakrawcz-rdx, line 282: (nitpick, just to remember during next refactor here)
 ```suggestion
                    self.get_execution_config(parent_transaction_identifiers),
```

core-rust/core-api-server/src/core_api/handlers/status_network_configuration.rs
dhedey: It was previously incorrect - the full address length includes the prefix byte, so is 27 bytes.

This was due to either a previous confusion of mine around address serialization, or a change in upstream behaviour of how they serialize addresses.

core-rust/core-api-server/src/core_api/helpers.rs
dhedey: How about eg `read_mandatory_substate_from_id_or_server_error`?

core-rust/state-manager/src/state_manager.rs
dhedey: Are you suggesting that we add this method `get_execution_config`?

This is sadly a classic rust-ism.

What you propose doesn't actually work - because of how Rust handles reference aliasing - it wouldn't allow (eg) a mutable reference to `self.execution_cache` at the same time as calling a method `self.get_execution_config(..)` on a shared reference &self, because it doesn't know that `self.get_execution_config` doesn't read from `self.execution_cache`, violating the mutable reference uniqueness.

Or am I misunderstanding you? :)

core-rust/core-api-server/src/core_api/conversions/receipt.rs
dhedey: Thanks! Nice spot. I've picked this up with Yulong and fixed this now :)

core-rust/core-api-server/src/core_api/helpers.rs
jakrawcz-rdx: whenever an `_or_` comes up in a method name, I can't help but recall a 2012 meme:
image

(I mean: what else a method, which returns a `Result<>`, can do, when a mandatory substate turns out to be absent?)

Still, this is only about naming, so feel free to ignore my sentiment :)

core-rust/state-manager/src/state_manager.rs
jakrawcz-rdx: > Or am I misunderstanding you? :)

no, I apparently meant to commit a rust-ism :p I understand now, let's live with the random `if` 👍 
(I do have a https://github.com/radixdlt/babylon-node/pull/364 pending, which touches the same lines and awaits a nasty merge; I'll see if I can do anything about it there)

core-rust/core-api-server/src/core_api/helpers.rs
dhedey: Hahaha love it. Well here `_or` really means `_else_` but yeah - happy to go back to `read_mandatory_substate_from_id` and not over correct! Will do that tomorrow

jakrawcz-rdx (approved):
real hero's work 👍

PR #368, opened 2023-03-20
Pushing the receipt's events from Engine to CoreApi

jakrawcz-rdx:


core-rust/core-api-server/src/core_api/conversions/receipt.rs
```diff
@@ -223,6 +232,39 @@ pub fn to_api_next_epoch(
...
+pub fn to_api_event(
+    context: &MappingContext,
+    type_id: EventTypeIdentifier,
+    data: Vec<u8>,
+) -> Result<models::Event, MappingError> {
+    let EventTypeIdentifier(emitter, local_type_index) = type_id;
+    Ok(models::Event {
+        _type: Box::new(models::EventTypeIdentifier {
+            schema_local_index: match local_type_index {
+                LocalTypeIndex::SchemaLocalIndex(index) => index.try_into().unwrap(),
```
jakrawcz-rdx, line 245: _(note from author)_

I've discussed this with @0xOmarA and looks like we simply have to live with the fact that the `event`'s type is specified as the index of the type definition within the `emitter`'s schema. (please note it is never `WellKnown`, always `SchemaLocalIndex`, by design).

I've explored the ways to map the `index -> event name` here (following the https://github.com/radixdlt/radixdlt-scrypto/blob/0b0b30b04bd3f7c99ea608bf3425958f0fd32bda/scrypto-unit/src/test_runner.rs#L1139-L1222), and sadly, we don't contact the persistence in our API mapping layer. We could do it (preload schemas from the right blueprints?) in the handler, but it's rather cumbersome (and lots of code that would pollute this otherwise simple PR).

@KrzysztofLabus-RDX I imagine that the Gateway may be more interested in performing that mapping - please let me know if it becomes even more impractical on the Gateway's side.

malexandru-rdx (approved):

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -821,6 +821,10 @@ components:
           $ref: "#/components/schemas/FeeSummary"
         state_updates:
           $ref: "#/components/schemas/StateUpdates"
+        events:
```
dhedey, line 824: I imagine all receipts will include this list? So we'd want to include `events` in the required list of this object on the schema?

dhedey: Ahh I see we use this for rejected responses in the preview. Hmmm.

I wonder if maybe we should change preview to use an enum/tagged union in its result, to return a different object upon rejection, so that this can just be for committed (success/failure) cases?

(Not necessarily in this PR though!). What are your thoughts on this?

core-rust/core-api-server/src/core_api/conversions/receipt.rs
dhedey: Right - I think the above approach is broadly correct - we can pass the event index upstream, and it'll be up to the Gateway API to resolve the correct schema at the given time when it responds to queries.

The Core API transaction stream operates on diffs, and I think it'd be wrong/incorrect to include stateful information there. (Although the package schema is - at least at the moment - immutable in the application layer, so this concern is a little less strong).

dhedey: Note that I added a LocalTypeIndex type into the schema that weshould possibly use for this 👍.

I don't think it necessarily matters at the Core API abstraction that the type can never be well known - I'm not even sure that that is an invariant to be honest(!).

EG there will be well known types which pass the struct/enum check in future! So I don't think we should panic here.

dhedey (approved):
Approving to let you merge after markups.

One thing that I definitely want changed is the type index mapping / assertion, otherwise it looks good 👍

core-rust/core-api-server/core-api-schema.yaml
jakrawcz-rdx: yes, exactly (I initially made it required, but then the current preview's code convinced me to do optional...)
This (new shape of the preview's result) is a material for a separate PR, since it will have to be synchronized with the NG (at least regarding deploy).

I think ideally I would see something like:
```
preview(tx) -> Response(Rejected(error_message) | Accepted(receipt));
get_receipt(tx) -> Response(receipt);
```
My doubts:
1. The `receipt` in both cases would be the same api type, right?
2. I assumed that the opposite of `Rejected` is `Accepted`, to avoid using the word `Committed` in preview.

core-rust/core-api-server/src/core_api/conversions/receipt.rs
jakrawcz-rdx: awww the `LocalTypeIndex` api type did not exist when I started this PR - in this case, OFC, let me re-use it now 👍

core-rust/core-api-server/core-api-schema.yaml
dhedey: Yep, exactly. Maybe `Accepted` would be `WouldBeCommitted` or something, to save introducing new nomenclature that people need to understand :)

PR #369, opened 2023-03-20
Feature/release artifacts

kofrdx:
# Release artifacts to github release

This PR aims to publish the binary and the library necessary to run the node natively (as compared to only on docker) to the Github release. 

This enables the node-cli to download and configure them.
Also interested node-runners will not need to build the core-rust library themselves or copy it out of the docker image.

Please have a look at this release:
https://github.com/radixdlt/babylon-node/releases/tag/poc-release-artifacts
and also approve/merge this PR 
 https://github.com/radixdlt/babylon-node/pull/370
 
 
 Can also be compared to radixdlt/radixdlt.
 
 https://github.com/radixdlt/radixdlt/blob/main/.github/workflows/add-artifacts-to-release.yml

### DIFF
--- a/.github/workflows/add-artifacts-to-release.yml
+++ b/.github/workflows/add-artifacts-to-release.yml
@@ -31,8 +31,7 @@ jobs:
         run: ./gradlew :core-rust:buildRustForDocker -P rustBinaryBuildType="release"
       - name: Package libcorerust
         run: |
-          cd ${{ github.workspace }}/core-rust/target/x86_64-unknown-linux-gnu/release/
-          zip babylon-node-rust-arch-linux-x86_64-release-${{ github.event.release.tag_name }}.zip libcorerust.so
+          zip -j babylon-node-rust-arch-linux-x86_64-release-${{ github.event.release.tag_name }}.zip ${{ github.workspace }}/core-rust/target/x86_64-unknown-linux-gnu/release/libcorerust.so
       - name: Get release
         id: get_release
         uses: bruceadams/get-release@f589ce0779c7bef1faf175f7488c972eb47dc046
--- a/.github/workflows/add-artifacts-to-release.yml
+++ b/.github/workflows/add-artifacts-to-release.yml
@@ -1,4 +1,4 @@
-name: radixdlt-artifacts-to-release
+name: Release artifacts
 on:
   release:
     types: [published]
--- a/.github/workflows/add-artifacts-to-release.yml
+++ b/.github/workflows/add-artifacts-to-release.yml
@@ -32,7 +32,6 @@ jobs:
       - name: Package libcorerust
         run: |
           zip babylon-node-rust-arch-linux-x86_64-release-${{ github.event.release.tag_name }}.zip ${{ github.workspace }}/core-rust/target/x86_64-unknown-linux-gnu/release/libcorerust.so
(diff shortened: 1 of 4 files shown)


.github/workflows/add-artifacts-to-release.yml
```diff
@@ -0,0 +1,88 @@
...
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
+      - name: Run local core-rust docker build
+        run: ./gradlew :core-rust:buildRustForDocker -P rustBinaryBuildType="release"
+      - name: Upload libcorerust
+        run:
+          gh release upload ${{ github.event.release.tag_name }} --repo radixdlt/babylon-node ${{ github.workspace }}/core-rust/target/x86_64-unknown-linux-gnu/release/libcorerust.so --clobber
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
shambupujar, line 36: We may need to pull this from AWS secrets

.github/workflows/add-artifacts-to-release.yml
```diff
@@ -0,0 +1,88 @@
...
+        name: Get radixdlt version
+        uses: ./.github/actions/get-version
+        with:
+          is_release: true
+      - name: List files
+        run: ls -la ./core/build/distributions/
+      - name: Upload distribution zip
+        uses: actions/upload-release-asset@e8f9f06c4b078e705bd2ea027f0926603fc9b4d5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
shambupujar, line 83: Same here - Pull if you can from AWS secrets

.github/workflows/add-artifacts-to-release.yml
```diff
@@ -0,0 +1,88 @@
...
+        uses: actions/cache@v1
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
+      - name: Run local core-rust docker build
+        run: ./gradlew :core-rust:buildRustForDocker -P rustBinaryBuildType="release"
+      - name: Upload libcorerust
+        run:
+          gh release upload ${{ github.event.release.tag_name }} --repo radixdlt/babylon-node ${{ github.workspace }}/core-rust/target/x86_64-unknown-linux-gnu/release/libcorerust.so --clobber
```
shambupujar, line 34: Any reason why you have to resort to gh cli .  We could do this using the action `actions/upload-release-asset`

.github/workflows/add-artifacts-to-release.yml
kofrdx: its the standard GITHUB_TOKEN. Its always available on hosted runners. We actually cant put this into AWS.

.github/workflows/add-artifacts-to-release.yml
kofrdx: Sadly I deleted the tries with this action. They ended up canceling due to file type and size when uploading the debug build. The release is much smaller and if we end up zipping it either way like in the discussion, I can use the github action again.

I mentioned this in the daily, but I guess it was not clear.

kofrdx: Does not apply anymore since the library gets zipped now

shambupujar (approved):

PR #371, opened 2023-03-20
feature: exchange API

malexandru-rdx:
This PR implements an schema-isolated Long Term Support/backward compatible API for integrators such as exchanges.
This includes:

- `/lts/transaction/construction` - everything needed in order to build a transaction
- `/lts/transaction/submit` and `/lts/transaction/status` - as of right now, just a copy of the current `/transaction/submit` and `/transaction/status` but with it's own schema
- `/lts/state/account-all-fungible-resource-balances` - query latest list of fungible resources owned by an account
- `/lts/state/account-fungible-resource-balances` - query latest amount of a specific resource owned by an account
- `/lts/stream/transactions-basic-outcomes` - query a range (by state version) of basic outcomes (status and fungible resource changes)

No extra DB indices are added in this PR.


core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -2729,7 +2920,381 @@ components:
...
+      type: object
+      required:
+        - network
+      properties:
+        network:
+          $ref: "#/components/schemas/NetworkIdentifier"
+    LtsTransactionConstructionResponse:
+      type: object
+      required:
+        - epoch
```
dhedey, line 2936: I wonder if we should call this `current_epoch`? What are your thoughts? 👍

dhedey: I wonder also if we should return ledger time here so people can check things are synced up.

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -2729,7 +2920,381 @@ components:
...
+          $ref: "#/components/schemas/IntentHash"
+    LtsTransactionStatusResponse:
+      type: object
+      required:
+        - intent_status
+        - status_description
+        - known_payloads
+      properties:
+        intent_status:
+          $ref: "#/components/schemas/TransactionIntentStatus"
```
dhedey, line 2965: It might be worth copying these down for `TransactionIntentStatus`/`TransactionPayloadStatus`, so that we allow ourselves to change these in the main transaction status endpoint in future if we want

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -2729,7 +2920,381 @@ components:
...
+          type: array
+          items:
+            $ref: "#/components/schemas/LtsCommittedTransactionFungibleOutcome"
+    LtsCommittedTransactionStatus:
+      type: string
+      enum:
+        - Succeeded
+        - Failed
+      description: The status of the transaction
+    LtsCommittedTransactionFungibleOutcome:
```
dhedey, line 3243: Let's just call this `LtsCommittedTransactionOutcome` and rename the `fungible_outcome` field  to `fungible_entity_balance_changes`.

That way, if we wish to add non fungible changes to this later, we can, without breaking anything.

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -2729,7 +2920,381 @@ components:
...
+        - network
+        - account_address
+        - fungible_resource_address
+      properties:
+        network:
+          $ref: "#/components/schemas/NetworkIdentifier"
+        account_address:
+          type: string
+          description: The Bech32m-encoded human readable version of the account's address
+        fungible_resource_address:
```
dhedey, line 3160: I think it's fine / better to just call this `resource_address`, as long as we give a decent error message if they provide a nf resource address.

I care about including accurate naming on our responses, so people don't get confused. I mind less on the inputs.

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -2729,7 +2920,381 @@ components:
...
+    LtsFungibleResourceBalanceChange:
+      type: object
+      required:
+        - fungible_resource_address
+        - balance_change
+      properties:
+        fungible_resource_address:
+          type: string
+          description: The Bech32m-encoded human readable version of the fungible resource's global address
+        balance_change:
```
dhedey, line 3293: We also need to include fee payment information. I wonder if we should have an optional field here `fee_payment` (which is only included if the amount is XRD)?

And include in the description for `balance_change` that it _includes_ the fee payment?

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -2729,7 +2920,381 @@ components:
...
+            The hex-encoded transaction accumulator hash. This hash captures the order of all transactions on ledger.
+            This hash is `ACC_{N+1} = Blake2b-256(CONCAT(ACC_N, LEDGER_HASH_{N}))`, starting with `ACC_0 = 000..000` the pre-genesis accumulator.
+        status:
+          $ref: "#/components/schemas/LtsCommittedTransactionStatus"
+        fungible_outcome:
+          description: A list containing all account fungible balance updates
+          type: array
+          items:
+            $ref: "#/components/schemas/LtsAccountFungibleBalanceChanges"
+    LtsAccountFungibleBalanceChanges:
```
dhedey, line 3271: I think this should actually be LtsEntityFungibleBalanceChanges and include all entities - not just accounts.

As exchanges like sanity checking things like total balance change = 0 (less fees)

core-rust/core-api-server/src/core_api/handlers/lts/state_account_all_fungible_resource_balances.rs
```diff
@@ -0,0 +1,84 @@
+use crate::core_api::*;
+use radix_engine::types::{Decimal, IndexMap};
+use state_manager::store::traits::QueryableProofStore;
+use state_manager::{
+    jni::state_manager::ActualStateManager,
+    query::{dump_component_state, VaultData},
+};
+
+#[tracing::instrument(skip(state), err(Debug))]
+pub(crate) async fn handle_rc_state_account_all_fungible_resource_balances(
```
dhedey, line 10: Should probably have `_lts_` in the name now (and same for all methods in these files)

core-rust/core-api-server/src/core_api/handlers/lts/state_account_resource_balance.rs
```diff
@@ -0,0 +1,84 @@
...
+        return Err(client_error(
+            "The provided resource address is not a fungible resource.",
+        ));
+    }
+
+    let component_address =
+        extract_component_address(&extraction_context, &request.account_address)
+            .map_err(|err| err.into_response_error("account_address"))?;
+
+    let component_dump = dump_component_state(state_manager.store(), component_address)
```
dhedey, line 50: Not super important, but before launch we should definitely make this more efficient - by just looking up the substate path we need, direct to the single vault for this resource in the account.

Could we add a TODO here?

core-rust/core-api-server/src/core_api/handlers/lts/state_account_all_fungible_resource_balances.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    let mapping_context = MappingContext::new(&state_manager.network);
+    let extraction_context = ExtractionContext::new(&state_manager.network);
+
+    let component_address =
+        extract_component_address(&extraction_context, &request.account_address)
+            .map_err(|err| err.into_response_error("account_address"))?;
+
+    let component_dump = dump_component_state(state_manager.store(), component_address)
```
dhedey, line 40: If the component doesn't exist, what error do we get here?

We should ensure that:
* If the account address is a _normal_ account address, then we return a 404
* If the account address is a _virtual_ account address, then we return a response with no resources.

_Note:_ We can tell if the account address is virtual or not by matching on the component address

core-rust/core-api-server/src/core_api/handlers/lts/state_account_resource_balance.rs
dhedey: If the component doesn't exist, what error do we get here?

We should ensure that:
* If the account address is a normal account address, then we return a 404
* If the account address is a virtual account address, then we return a response with 0 balance.

Note: We can tell if the account address is virtual or not by matching on the component address

core-rust/core-api-server/src/core_api/handlers/lts/stream_transactions_basic_outcomes.rs
```diff
@@ -0,0 +1,136 @@
...
+                &mapping_context,
+                receipt,
+                identifiers,
+            )?)
+        })
+        .collect::<Result<Vec<models::LtsCommittedTransactionFungibleOutcome>, ResponseError<()>>>(
+        )?;
+
+    let start_state_version = if api_txns.is_empty() {
+        0
```
dhedey, line 67: This logic feels... wrong?  But I know it was copied from the previous place.
I feel it should just be returning from_state_version - I'm not sure why this check makes sense. If you agree, can you fix it in all the places? 👍 :)

core-rust/core-api-server/src/core_api/handlers/lts/stream_transactions_basic_outcomes.rs
```diff
@@ -0,0 +1,136 @@
...
+                            Some(models::LtsFungibleResourceBalanceChange {
+                                fungible_resource_address: to_api_resource_address(
+                                    context,
+                                    resource_address,
+                                ),
+                                balance_change: to_api_decimal(balance_change),
+                            })
+                        }
+                        BalanceChange::NonFungible {
+                            added: _added,
```
dhedey, line 121: You can just do .. here to avoid listing out the fields you ignore

dhedey (approved):
Really nice work.

A few suggestions for slight name changes in the schema, and some additions / things that were missed, but really nice work 👍.

In terms of next steps / priorities, I suggest the following:
* Markups on this
* Add the account transactions endpoint to the schema, which just returns an error for now.
* Add the index and hook up the account transactions endpoint
* Add some integration tests for each of these endpoints - see tests deriving from `DeterministicCoreApiTestBase`

If we could do 1 and 2 on Monday, and have 3 done by Wednesday, that'd be great. Shout to me or @LukasGasior1 if you need any help on 3.  :)

core-rust/core-api-server/core-api-schema.yaml
malexandru-rdx: If by ledger time you mean the `proposer_timestamp_ms`, it might be useful indeed (what would a healthy range to check against be?). The naive way is to build the transaction with `start_epoch = current_epoch` and `end_epoch = current_epoch + 20` if you want to "shoot and forget" and don't really care about when exactly it gets committed. If you want to better control the epoch try `[current_epoch; current_epoch + 1 (or 0)]` over and over until it get's committed.

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -2933,17 +2933,21 @@ components:
...
+        current_epoch:
           type: integer
           format: int64
           minimum: 0
           maximum: 10000000000
           description: An integer between `0` and `10^10`, marking the current epoch
-###################################
+        ledger_clock:
+          $ref: '#/components/schemas/Instant'
+          description: The latest round proposer's timestamp (rounded down to the current minute).
```
dhedey, line 2947: The reason for including this field is to detect whether the node is synced up - so it's probably worth mentioning in the description that clients may wish to double-check this field before blindly using `current_epoch` for construction.

Alternatively, we could allow an (optional) request parameter with "acceptable_delay_seconds" which causes a 400 to be returned if the node isn't sufficiently synced up

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -3263,20 +3309,25 @@ components:
             This hash is `ACC_{N+1} = Blake2b-256(CONCAT(ACC_N, LEDGER_HASH_{N}))`, starting with `ACC_0 = 000..000` the pre-genesis accumulator.
         status:
           $ref: "#/components/schemas/LtsCommittedTransactionStatus"
-        fungible_outcome:
-          description: A list containing all account fungible balance updates
+        fungible_entity_balance_changes:
+          description: A list containing all account fungible sbalance updates
```
dhedey, line 3313: Typo in balance, and it's not just accounts any more.

Perhaps the following:
`A list of all fungible balance updates which occurred in this transaction, aggregated by the global entity (such as account) which owns the vaults which were updated.`

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -3263,20 +3309,25 @@ components:
...
+        fee:
+          type: string
+          description: |
+            The string-encoded decimal representing the total amount of XRD payed as fee (execution, validator tip and royalties).
+            A decimal is formed of some signed integer `m` of attos (`10^(-18)`) units, where `-2^(256 - 1) <= m < 2^(256 - 1)`.
+    LtsEntityFungibleBalanceChanges:
       type: object
       required:
-        - account_address
+        - address
```
dhedey, line 3325: For consistency with other parts of the API, I might suggest calling it `entity_address`

core-rust/core-api-server/src/core_api/conversions/lts.rs
```diff
@@ -0,0 +1,56 @@
+use radix_engine::transaction::BalanceChange;
+use state_manager::{
+    CommittedTransactionIdentifiers, LedgerTransactionOutcome, LedgerTransactionReceipt,
+};
+
+use crate::core_api::*;
+
+#[tracing::instrument(skip_all)]
+pub fn to_api_lts_comitted_transaction_basic_outcome(
```
dhedey, line 9: `to_api_lts_comitted_transaction_outcome` now I guess? :)

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -3295,6 +3346,66 @@ components:
...
+          minimum: 1
+          maximum: 100000000000000
+          description: An integer between `1` and `10^13`, giving the first (resultant) state version to be returned
+        limit:
+          description: The maximum number of transactions that will be returned.
+          type: integer
+    LtsStreamAccountTransactionsBasicOutcomesResponse:
+      type: object
+      required:
+        - from_state_version
```
dhedey, line 3379: To enable this endpoint to be efficiently polled, we may wish to provide a `next_state_version` which is the state version at which the iteration should start. (ie one more than the max the iterator was at this response).

In particular, even if no entries were returned, it should probably give `max_ledger_state_version + 1` I guess - to save re-scanning over things that have already been checked.

I guess this isn't super important if we have an efficient index and aren't doing any additional filtering beyond using the index; but it's possibly a good pattern regardless?

Happy to hear your thoughts :)

dhedey (approved):
Small mark-ups, happy to merge when you're done

core-rust/core-api-server/core-api-schema.yaml
malexandru-rdx: Unless we can offer a more accurate but complex logic for node status I would leave this (simple) behaviour to the client.

core-rust/core-api-server/core-api-schema.yaml
malexandru-rdx: Good point if `limit` is referring to the `state_version` and indeed, the last transaction outcome is not the last scanned. I consider `limit` refers to the number of transactions returned, and in this case, the last transaction would basically have this information.

I'll change this if it makes sense in a later PR.

PR #372, opened 2023-03-21
Split the committed receipt into "on-ledger receipt" vs "local, non-verifiable execution details"

jakrawcz-rdx:
As designed in https://radixdlt.atlassian.net/wiki/spaces/S/pages/3016884225/REP-65+Consensus+Transaction+Receipt#Splitting-the-Engine-Receipt .

This PR splits our existing `LedgerTransactionReceipt` into 2 parts:
- the actual "what goes onto ledger and will be verifiable against hashes found in ledger proofs"
  - this part re-uses the very appropriate name `LedgerTransactionReceipt`
  - it contains:
     - a simple success/failure outcome 
     - substate changes
     - _(coming soon, after https://github.com/radixdlt/babylon-node/pull/368 merges)_ events
  - **every** piece of information from this part will be reflected (either directly or through some intermediate merkle hash) in the consensus' receipt hash (see the `LedgerTransactionReceipt.get_consensus_receipt() -> ConsensusReceipt` and `ConsensusReceipt.get_hash() -> LedgerReceiptHash`)
- the additional, non-verifiable information coming from the local execution of this transaction (i.e. from the Engine's "receipt")
   - this part is named `LocalTransactionExecution` 
   - it contains basically everything that didn't stay in `LedgerTransactionReceipt` (so: logs, entity changes, next epoch info)
   - I assume that **every** piece of information from this part:
     - is either irrelevant to ~all ledger use-cases (e.g. logs)
     - or can be derived from the `LedgerTransactionReceipt` (e.g. entity changes), so it only exists for convenience/cache

The 2 parts combined together are called `LocalTransactionReceipt`. This struct is handy to pass around, and in fact, in this PR, it is only split into 2 at the persistence layer (where each part lives in its own table).


core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -96,8 +96,10 @@ use crate::transaction::LedgerTransaction;
 enum RocksDBColumnFamily {
     /// Committed transactions
     TxnByStateVersion,
-    TxnReceiptByStateVersion,
     TxnAccumulatorHashByStateVersion,
+    /// Receipts of committed transactions
+    LedgerReceiptByStateVersion,
+    LocalTransactionExecutionByStateVersion,
```
jakrawcz-rdx, line 102: _(note from author: this split at DB level is today strictly "conceptual", but @LukasGasior1 recommends it hard, and it may have real uses in future)_

core-rust/state-manager/src/jni/transaction_store.rs
```diff
@@ -116,24 +116,20 @@ fn do_get_transaction_at_state_version(
...
-            LedgerTransactionOutcome::Success(output) => TransactionOutcomeJava::Success(output),
-            LedgerTransactionOutcome::Failure(err) => {
+        outcome: match local_execution.outcome {
+            DetailedTransactionOutcome::Success(output) => TransactionOutcomeJava::Success(output),
+            DetailedTransactionOutcome::Failure(err) => {
                 TransactionOutcomeJava::Failure(format!("{err:?}"))
             }
         },
-        ledger_receipt_bytes,
+        ledger_receipt_bytes: scrypto_encode(&ledger_receipt.get_consensus_receipt()).unwrap(),
```
dhedey, line 129: I think the general direction is good, but there's slightly confusing naming / renaming going on here - we have a:
* CommittedTransactionReceipt
* LocalReceipt
* LedgerReceipt
* ConsensusReceipt
* LocalExecution (which includes both Outcome and ExecutionTrace from the engine receipt)
* In the engine: All the engine receipt names
* In other places in this PR: "full receipt"

I'm not entirely sure I understand clearly what they all are :) - and there are still some consistency clean-ups we need.

I suggest we have a clear "Proposed Naming and Concept Grouping" section of `REP-65` - which covers both the receipts in the Engine, and remapped receipts in the node, which we can all agree on. (and would also cover which things - such as Balance Changes, and Parent information - can be derived from which receipt)

EG I think maybe we just need to call this "consensus_receipt_bytes" in the API and we'd almost be there.

core-rust/state-manager/src/receipt.rs
```diff
@@ -64,15 +58,30 @@ pub struct DeletedSubstateVersion {
 
 #[derive(Debug, Clone, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
 pub enum LedgerTransactionOutcome {
+    Success,
+    Failure,
+}
+
+impl From<&TransactionOutcome> for LedgerTransactionOutcome {
```
dhedey, line 65: It's rare to see `From` implementations from references because the "From" trait suggests a conversion / ownership move, and actually using them is a bit messy - but this one looks like it might be okay 👍.

Alternatively, you can do this via a constructor on `LedgerTransactionOutcome` or a method on TransactionOutcome (although I guess we don't control that).

core-rust/state-manager/src/receipt.rs
```diff
@@ -85,63 +94,68 @@ impl From<TransactionOutcome> for LedgerTransactionOutcome {
...
     pub state_update_summary: StateUpdateSummary,
     // These will be removed once we have the parent_map for the toolkit to use
     pub resource_changes: IndexMap<usize, Vec<ResourceChange>>,
     pub next_epoch: Option<(BTreeMap<ComponentAddress, Validator>, u64)>,
 }
 
 impl LedgerTransactionReceipt {
-    pub fn get_hash(&self) -> LedgerReceiptHash {
-        LedgerReceiptHash::for_receipt(self)
+    pub fn get_consensus_receipt(&self) -> ConsensusReceipt {
```
dhedey, line 129: If lifetime is quite big, and the consensus receipt is a transient structure, we could also consider it containing a reference or Cow (with a lifetime) to outcome (using a Cow rather than a lifetime would also enable it to own the outcome so it can be a deserialisation target as well).

core-rust/state-manager/src/receipt.rs
```diff
@@ -85,63 +94,68 @@ impl From<TransactionOutcome> for LedgerTransactionOutcome {
...
+                substate_changes: fix_state_updates(commit_result.state_updates),
+            },
+            local_execution: LocalTransactionExecution {
+                outcome: commit_result.outcome.into(),
+                fee_summary: commit_result.fee_summary,
+                fee_payments: commit_result.fee_payments,
+                application_logs: commit_result.application_logs,
+                state_update_summary: commit_result.state_update_summary,
+                resource_changes: execution_trace.resource_changes,
+                next_epoch,
```
dhedey, line 157: As a heads up - some of these things (eg next_epoch), and eventually balance_changes, state_update_summary, etc - will be computable from the LedgerTransactionReceipt, and so won't need to be explicitly stored 👍

core-rust/state-manager/src/store/in_memory.rs
```diff
@@ -103,8 +104,9 @@ impl InMemoryStore {
     pub fn new() -> InMemoryStore {
         InMemoryStore {
             transactions: BTreeMap::new(),
-            transaction_receipts: BTreeMap::new(),
             transaction_identifiers: BTreeMap::new(),
+            ledger_receipts: BTreeMap::new(),
```
dhedey, line 108: I wonder if it's actually worth it to store these separately or whether it just introduces overhead?

I guess storing them separately allows us to possibly prune them separately easily? 👍

dhedey (approved):
Really nice - my only real suggestion is that we clearly document / agree the naming / grouping for the different receipt abstractions between Network and Builders teams, and then work to ensure we're using the right nomenclature - because without clear docs and mutual agreement, we're going to pull in different directions and use different names for things and it's going to become a mess :)

core-rust/state-manager/src/receipt.rs
jakrawcz-rdx: > will be computable

wooooo, finally 🎉

core-rust/state-manager/src/store/in_memory.rs
jakrawcz-rdx: The main factor for splitting at DB came from Łukasz (https://github.com/radixdlt/babylon-node/pull/372#discussion_r1143667148): I think his suggested use-case was that we could have an app flag driving whether an instance should store the "local execution" at all or not (I guess it touches your "full node vs validator").

The DB code change wasn't too cumbersome and the overhead doesn't seem crazy ("do 2 lookups instead of 1"), so I went with it.

One more minor advantage: in some places in the code I've seen that (after some extra changes _not_ belonging in this PR) we could load just the smaller LedgerTransactionReceipt.

Generally, my opinion here is not very strong. If the DB-level split causes any problems, I can easily revert this part.

jakrawcz-rdx: > [sometimes] we could load just the smaller LedgerTransactionReceipt

...ah, and potentially more of such places will pop up after https://github.com/radixdlt/babylon-node/pull/372#discussion_r1146167124 happens.

core-rust/state-manager/src/receipt.rs
jakrawcz-rdx: I see; I guess this is mostly about the `From<>`'s convention.
I hope there is nothing wrong about resolving the `LedgerTransactionOutcome` from a `&TransactionOutcome`, so I simply renamed it to `LedgerTransactionOutcome::resolve(&TransactionOutcome)` 👍

jakrawcz-rdx:
Let me merge this - I assume you are fine with the naming described at https://github.com/radixdlt/babylon-node/pull/372#discussion_r1146435753 - if not, then I will do renames in a separate PR, because there are already 2 more PRs stacked on top of this one.

core-rust/state-manager/src/receipt.rs
dhedey: Sounds grand, thanks :)

core-rust/state-manager/src/jni/transaction_store.rs
dhedey: Looks great, thanks. I just want to make sure we're all clear on the naming scheme so we can be consistent with it!

PR #383, opened 2023-03-24
Putting application events' merkle root into consensus receipt

jakrawcz-rdx:
⚠️ This stack on top of https://github.com/radixdlt/babylon-node/pull/374

Actually it simply re-applies the same pattern (from substate changes) to events.

### DIFF
--- a/core-rust/core-api-server/src/core_api/conversions/receipt.rs
+++ b/core-rust/core-api-server/src/core_api/conversions/receipt.rs
@@ -252,7 +252,10 @@ pub fn to_api_event(
     context: &MappingContext,
     event: ApplicationEvent,
 ) -> Result<models::Event, MappingError> {
-    let EventTypeIdentifier(emitter, local_type_index) = event.type_id;
+    let ApplicationEvent {
+        type_id: EventTypeIdentifier(emitter, local_type_index),
+        data,
+    } = event;
     Ok(models::Event {
         _type: Box::new(models::EventTypeIdentifier {
             emitter: Some(match emitter {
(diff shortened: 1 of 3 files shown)


core-rust/core-api-server/src/core_api/conversions/receipt.rs
```diff
@@ -249,10 +250,9 @@ pub fn to_api_next_epoch(
 #[tracing::instrument(skip_all)]
 pub fn to_api_event(
     context: &MappingContext,
-    type_id: EventTypeIdentifier,
-    data: Vec<u8>,
+    event: ApplicationEvent,
 ) -> Result<models::Event, MappingError> {
-    let EventTypeIdentifier(emitter, local_type_index) = type_id;
+    let EventTypeIdentifier(emitter, local_type_index) = event.type_id;
```
dhedey, line 255: In these methods, I suggest that we use a struct-destructuring. This way, if the upstream models change / get new fields, we know we need to map them at compile time :)

core-rust/state-manager/src/receipt.rs
```diff
@@ -36,6 +37,18 @@ impl CommittedTransactionIdentifiers {
...
+pub struct ApplicationEvent {
+    pub type_id: EventTypeIdentifier,
+    pub data: Vec<u8>,
+}
+
+impl ApplicationEvent {
+    pub fn new(type_id: EventTypeIdentifier, data: Vec<u8>) -> Self {
+        Self { type_id, data }
+    }
+}
```
dhedey, line 50: Ditto - as per my comment on the other PR - I wonder if it's worth us explicitly spelling out how the receipt event hash is constructed in a method here?

dhedey (approved):
Nice! 👍

core-rust/state-manager/src/receipt.rs
jakrawcz-rdx: I've now added a few more rustdocs 👍

PR #386, opened 2023-03-29
tweak: Update scrypto version to rcnet-v1-71c24148

dhedey:
Looks like no other changes required in node to get it working 👍


jakrawcz-rdx (approved):
just works? suspicious, but if it compiles, then 🟢

PR #387, opened 2023-03-29
Including the state root and receipt root into ledger header's hash

jakrawcz-rdx:
I've also improved the `warn!` that we log on hash mismatch, but TBH after the change, it has a rather low chance of ever being logged (I mean, 2/3 validators have already voted on these very hashes matching) - so I am open to deleting the `warn!` altogether.


LukasGasior1 (approved):
LGTM

LukasGasior1:
> so I am open to deleting the warn! altogether.

I'd replace it with an error :)

jakrawcz-rdx:
> I'd replace it with an error :)

Ok, I did it, but you click merge? 😈

LukasGasior1:
haha, I meant an error as in `CommitError`, not a different log level :D why would you even commit if you have a guarantee that your state is in some way wrong?

jakrawcz-rdx:
aaaaaah I see, ok, done 👍 

(I left the `error!` there too, since it gives some details; but a new `CommitError::LedgerHashesMismatch` is then returned - please see if this is what you meant)

jakrawcz-rdx:
huh, nope... the unit tests with pre-computed ledger-hashes run fine on my machine, but fail on jenkins, e.g. `test_invalid_rev2_transaction_fails()` there gives:
```
ERROR do_commit: state_manager::state_manager: computed ledger hashes at version 1 differ from the ones in proof (LedgerHashes { state_root: StateHash("e0ed76edc983db092048d05bedeb10dae658ced71ca5627f3de715b4d7f3fa5f"), transaction_root: TransactionTreeHash("289b3c43623c980d295e3a9cf27e18a66e3db68558feb8d84b26bad6248ece2f"), receipt_root: ReceiptTreeHash("b599bb5ea5147c3a5fe2cb23acd5d126dc576a407068c32d7e69b45dd20349ef") } != LedgerHashes { state_root: StateHash("2f89b82f25d266c8f5d33b0edebb7350b5db5aa3834dd0140b5e4730d3e2d00f"), transaction_root: TransactionTreeHash("99fe2610e5e29650a5967ad4a78bd1e9e3c33819b7b0f6e0700638f39efa6218"), receipt_root: ReceiptTreeHash("45ed5e164b89b048abb88593a1cdefe79ecdb202c9095d800f9a67dbe852db45") })
```
_(the jenkins-computed hashes are different than on my machine)_

Also, some integrations tests fail too, with `Could not commit: LedgerHashesMismatch[]` (e.g. `FProposalsPerRoundDropperTest`, `IncreasingValidatorsTest`). They are harder to investigate (I suspect again, half-mocked test data, but cannot be sure?).

I am afraid that with our current test set-up, we cannot return `Err` from commit on result-hash mismatch :(

LukasGasior1:
> I am afraid that with our current test set-up, we cannot return Err from commit on result-hash mismatch :(

That's fine, let's just keep the log message for now and perhaps add a todo comment to turn it into an actual error at some point.

jakrawcz-rdx:
There is classic off-by-one bug in Node's transaction/receipt tree integration code (not the accu tree itself). I plan to fix it in this PR, but let me first fix tests...

⚠️ blocked by https://github.com/radixdlt/babylon-node/pull/390

core-rust/state-manager/src/state_manager.rs
```diff
@@ -909,23 +910,21 @@ where
...
-                commit_request_start_state_version, base_transaction_identifiers.state_version
+                commit_request_start_state_version, base_state_version
             );
         }
 
         let mut state_tracker = StateTracker::initial(base_transaction_identifiers);
         let mut committed_transaction_bundles = Vec::new();
         let mut substate_store_update = SubstateStoreUpdate::new();
         let mut state_tree_update = HashTreeUpdate::new();
-        let transaction_tree_len = epoch_transactions_count + 1; // starts with previous epoch root
```
jakrawcz-rdx, line 928: _Note from author:_

This was the off-by-one error, which did not take the "empty tree at start of epoch" case into account. Other places (that were also dealing with "per-epoch transaction trees") got this part right. So I extracted that logic into a common helper (kinda overkill, but that's the only way I could make it obvious; it is really hard to assert that thing in integration test).

jakrawcz-rdx:
sweet... the `IncreasingValidatorsTest` is passing on CI (and locally), but now the sibling `RandomValidatorsTest` fails, with different error (`thread '<unnamed>' panicked at 'not captured yet', state-manager/src/receipt.rs:321:23`).

Locally too.

And on plain `develop` too 🤦‍♂️ . I think I know what's happening.

jakrawcz-rdx:
Huh, that's another genuine bug found in the same, unrelated PR:

Guess what substate change root hash we return when a transaction happened to generate no substate changes?
Yup, `panic!`, that's the hash we return...
_(same for event root)_

PR #389, opened 2023-03-30
DO 1336 - Optimise Dockerfiles

kofrdx:
babylon-node-docker-build

This PR aims to do the following:

- One Dockerfile to build (1) the java core app, (2) rust library and (3) the docker image with multi-stage. This way we can just build everything with docker build . -t radixdlt/babylon-node:local-build
- Gradle is only used for the java build. (I always find gradle hard to read)
- The rust build is done with cargo chef which makes caching dependencies in Docker more readable.
- The artifacts are directly copied from the build-stages. No deb-package needed (might increase compatability for other distros down the line?) The build artifacts can be copied to local seperately by specifying the --target and --output parameter in docker. No dependencies to local builds necessary. Anyone can build it for their local environment.
- the cargo build --target <xyz> gets selected depending on the --platform parameter in docker. This fixes a bug with arm64 docker images we had earlier.

**Changes**

- [x] Remove fixable security vulnerabilites (done by using latest images and binaries)
- [x] Fixate binary version to install/download
- [x] Fix Hadolint findings on all Dockerfiles
- [x] User remote caching with buildx on Google Artifact Hub for Core-rust(libcorerust.so) builds
- [x] Allow arm64 builds for Core-rust (was missing platform switch to be deterministic)
- [x] Remove network utilities from babylon-node docker image
- [x] Add script to install network utilities on demand in docker image
- [x] (Proposal) Rewrite Rust-builder image to use '[cargo-chef](https://github.com/LukeMathWalker/cargo-chef)' to streamline cargo caching
- [x] Verify effectiveness of Caching changes
- [x] Update .dockerignore to improve caching and layersize
- [x] Use only one Dockerfile for building the application
- [x] Write documentation on how to build with docker
- [ ] Write documentation for users on how to install network utils

The full dockerbuild takes almost 30 minutes without cache.
https://github.com/radixdlt/babylon-node/actions/runs/4683921345/attempts/1

After that, if only the Java code has changed, its reduced to around 4 minutes:
https://github.com/radixdlt/babylon-node/actions/runs/4683921345/attempts/2

Consecutive runs can differ:
https://github.com/radixdlt/babylon-node/actions/runs/4683921345/attempts/3

There is still a need to improve the caching of the Java Application layer.

Despite the long initial build time, I believe it to be a very important change to streamline the development of this applications.


docker/scripts/install_network_utils.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+apt-get -y update
```
dhedey, line 2: Can we add a quick comment explaining what this is - is that it's a script that is intended to be manually run with sudo as part of debugging a container?

docs/development/README.md
```diff
@@ -94,3 +94,39 @@ If you meet this issue, check following configuration options:
  - `Project Structure -> Project Settings -> Project`, make sure `Project SDK` and `Project Language Level` is set to `17 (Preview) - Pattern matching for switch`.
  - `Project Structure -> Project Settings -> Modules`, make sure that every module has `Language Level` set to `17 (Preview) - Pattern matching for switch (Project default)`  
  - `Settings -> Build, Execution, Deployment -> Build Tools -> Gradle`, make sure that `Gradle JVM` is set to `Project JDK`. 
+
+### Building with docker
```
dhedey, line 98: Really like this! Thanks for documenting :)

docs/development/README.md
```diff
@@ -94,3 +94,39 @@ If you meet this issue, check following configuration options:
...
+![The structure of the docker image](babylon-node-docker-build.png)
+
+
+The application can be build using a single Dockerfile:
+
+```
+docker build . -t radixdlt/babylon-node 
+```
+
+If local testing is required, the artifacts can be produced locally with docker aswell by specifying the build target and outputting the result.
```
dhedey, line 109: Can we list the possible targets here and explain them? 👍

docs/development/README.md
```diff
@@ -94,3 +94,39 @@ If you meet this issue, check following configuration options:
...
+
+### Building with docker
+
+![The structure of the docker image](babylon-node-docker-build.png)
+
+
+The application can be build using a single Dockerfile:
+
+```
+docker build . -t radixdlt/babylon-node
```
dhedey, line 106: Can we also use `docker-compose up`?

dhedey: Actually scratch this - I'll let @malexandru-rdx pick this up in his workstream 👍

Dockerfile
```diff
@@ -0,0 +1,140 @@
...
+  libssl-dev=1.1.1n-0+deb11u4 \
+  pkg-config=0.29.2-1 \
+  unzip=6.0-26+deb11u1 \
+  wget=1.21-1+deb11u1 \
+  software-properties-common=0.96.20.2-2.1 
+RUN add-apt-repository -y ppa:openjdk-r/ppa && \
+  apt-get install -y --no-install-recommends \
+  openjdk-17-jdk=17.0.6+10-1~deb11u1
+
+#install Gradle
```
dhedey, line 27: Overly nit-picky comment - but it might be nice if all our comments in this file follow the same pattern - eg `# Hello` - hash, space, capital first letter?

Dockerfile
```diff
@@ -0,0 +1,140 @@
...
+FROM lukemathwalker/cargo-chef:0.1.52-rust-1.68.0 AS chef
+LABEL org.opencontainers.image.authors="devops@radixdlt.com"
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    clang=1:11.0-51+nmu5 \
+    libmpc-dev=1.2.0-1 \
+    libmpfr-dev=4.1.0-3 \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+WORKDIR /app
+WORKDIR /app
```
dhedey, line 58: Accidental duplicate line?

Dockerfile
```diff
@@ -0,0 +1,140 @@
...
+
+# Building core-rust library
+FROM lukemathwalker/cargo-chef:0.1.52-rust-1.68.0 AS chef
+LABEL org.opencontainers.image.authors="devops@radixdlt.com"
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    clang=1:11.0-51+nmu5 \
+    libmpc-dev=1.2.0-1 \
+    libmpfr-dev=4.1.0-3 \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
```
dhedey, line 56: Can you explain why we do this line?

Dockerfile
```diff
@@ -0,0 +1,140 @@
...
+RUN apt-get update -y && \
+    apt-get -y --no-install-recommends install \
+    openjdk-17-jre-headless=17.0.6+10-1~deb11u1 \
+    unzip=6.0-26+deb11u1 \
+    daemontools=1:0.76-7 \
+    libssl-dev=1.1.1n-0+deb11u4 \
+    software-properties-common=0.96.20.2-2.1 \
+    gettext-base=0.21-4 && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
```
dhedey, line 94: Can you explain why we do this line?

Dockerfile
```diff
@@ -0,0 +1,140 @@
...
+LABEL org.opencontainers.image.authors="devops@radixdlt.com"
+
+RUN apt-get update -y && \
+    apt-get -y --no-install-recommends install \
+    openjdk-17-jre-headless=17.0.6+10-1~deb11u1 \
+    unzip=6.0-26+deb11u1 \
+    daemontools=1:0.76-7 \
+    libssl-dev=1.1.1n-0+deb11u4 \
+    software-properties-common=0.96.20.2-2.1 \
+    gettext-base=0.21-4 && \
```
dhedey, line 92: Have you worked out why we need these dependencies? If so, can we add comments for why we need them? If not, no worries

> daemontools=1:0.76-7 \
    libssl-dev=1.1.1n-0+deb11u4 \
    software-properties-common=0.96.20.2-2.1 \
    gettext-base

Dockerfile
```diff
@@ -0,0 +1,140 @@
...
+
+# create configuration automatically when starting
+COPY docker/scripts/config_radixdlt.sh /opt/radixdlt/config_radixdlt.sh
+
+# copy configuration templates
+WORKDIR /home/radixdlt
+COPY docker/config/ /home/radixdlt/
+
+# Add script to install optional network utils
+COPY docker/scripts/install_network_utils.sh /opt/radixdlt/install_network_utils.sh
```
dhedey, line 107: Do you think it might be worth setting up a [motd](https://en.wikipedia.org/wiki/Message_of_the_day) to alert an ssh'ing user that they may wish to run this script?

core-rust/build.gradle
```diff
@@ -104,18 +104,26 @@ task buildRustDebug(type: Exec) {
 
 task buildRustForDocker(type: Exec) {
     def target = "x86_64-unknown-linux-gnu"
-
+    def docker_plattform = "linux/amd64"
```
dhedey, line 107: Minor typo: `docker_platform`* but also to be consistent with the other variable names, we should probably case it as `dockerPlatform`

docs/development/babylon-node-docker-build.png
dhedey, line 1: I love a good diagram, but I must admit - it took me a lot of context to understand this.

I think the a few tweaks might help to make it a lot clearer:
* Add a `Root Dockerfile` heading to the grey box
* Add a `Stages` heading under the ARG boxes
* Under each stage, update the explanations:
    * **binary-builder**: Builds the .jar and run scripts
    * **binary-container**: Contains just the built application at `/`
    * **library-builder**: Builds the native library from rust
    * **library-container**: Contains just the built native library at `/`
    * **babylon-node**: The final application container to run babylon node.
 * Add copy arrows from `binary-builder` to `binary-container` and from `library-builder` to `library-container`

docker/Dockerfile.keygen
```diff
@@ -1,10 +1,10 @@
 FROM eclipse-temurin:17-alpine
-MAINTAINER radixdlt <devops@radixdlt.com>
+LABEL org.opencontainers.image.authors="devops@radixdlt.com"
```
dhedey, line 2: I'm not sure how this build works - I imagine we might wish to update this to use the new docker file - probably in a follow up PR

dhedey (approved):
Fantastic work, thanks Kim for taking on this big task and approaching it with a lot of attention to detail and autonomy - I really like what we've ended up with.

Have a few small suggestions here and there - but most of these can come in a follow up PR, as we need to get this merged early next week so we can publish a x86 and arm docker image to dockerhub for easy development for integrators 👍

.github/workflows/docker.yml
```diff
@@ -19,73 +19,9 @@ jobs:
         uses: styfle/cancel-workflow-action@0.8.0
         with:
           access_token: ${{ github.token }}
-
-  build_deb:
```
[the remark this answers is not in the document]
kofrdx: The node cli does not download the debian package but the other build artifacts.
Those two for example:
- babylon-node-rcnet-v1-c6360105d.zip
- babylon-node-rust-arch-linux-x86_64-release-rcnet-v1-c6360105d.zip

We actually build all the parts for the systemd run ourselves. (like the systemd config file and dependencies)
I agree however that building a working debian package can be useful and should be published. But atm, I have a strong hunch that its not working smoothly. I would re-add it at a later stage, or just keep it now and not use the artifact. 

Here I just wanted to safe space and build time.

leave me a thumbs up for adding it back in or a thumbs down for keeping it deleted.

.github/workflows/docker.yml
```diff
@@ -99,20 +35,18 @@ jobs:
...
-      context: "docker"
-      dockerfile: "./docker/Dockerfile.core"
-      platforms: "linux/amd64,linux/arm64"
-      restore_artifact: "true"
-      artifact_name: "deb4docker"
-      artifact_location: "docker"
+      context: "."
+      dockerfile: "./Dockerfile"
+      platforms: "linux/amd64"
+      # TODO: remove this
```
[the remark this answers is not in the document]
kofrdx: before merging, yes and no (in german we say jein). 

There is a feature to use different tags for different caches. On main I would like to use the default cache.

.github/workflows/docker.yml
```diff
@@ -99,20 +35,18 @@ jobs:
...
         type=sha
-      context: "docker"
-      dockerfile: "./docker/Dockerfile.core"
-      platforms: "linux/amd64,linux/arm64"
-      restore_artifact: "true"
-      artifact_name: "deb4docker"
-      artifact_location: "docker"
+      context: "."
+      dockerfile: "./Dockerfile"
+      platforms: "linux/amd64"
```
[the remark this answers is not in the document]
kofrdx: Good catch. Removed it for temporary speed up.  Will readd and test again.

.github/workflows/docker.yml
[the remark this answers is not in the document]
dhedey: I’m happy for us to ditch it for this PR. But could I ask you to follow up with it later?

We may even want to ditch it for good. It might be worth us asking node runners if anyone’s using it.

If we choose to ditch it, we should ditch all the gradle stuff for building the image.

If we choose to keep it, we should probably test it somehow.

On a separate note I think the deb build might be responsible for giving the build a name which somehow gets onto the top-right of the dashboard - so we should probably check that the build number / tag is included via some other means in this build :)

docker/Dockerfile.keygen
kofrdx: I also think about follow up PR. Because I have no idea about the build/publish process of this yet.

Dockerfile
kofrdx: @dhedey What do you think of the message I just added?

Dockerfile
```diff
@@ -82,14 +82,19 @@ COPY --from=library-builder /libcorerust.so /
 FROM debian:11-slim as app-container
 LABEL org.opencontainers.image.authors="devops@radixdlt.com"
 
+# unzip is needed for unpacking the java build artifacts
+# daemontools is needed at application runtime for async tasks
+# libssl-dev is needed for encryption methods used in the keystore.ks
+# software-properties-common is neeed for installing debian packages with dpkg
+# gettext-base is needed for envsubst in config_radixdlt.sh
```
dhedey, line 89: Fantastic, thank you! :)

Dockerfile
dhedey: Really nice, thanks! ✨ 

Couple of minor thoughts:
* There are some Github Actions lint notices that we might need to work around - possibly extracting a set_motd.sh script (which can use a heredoc) or `motd.txt` might make it easier to edit 👍 
* "However we left you the instructions to install them in this file:" doesn't like they're written instructions rather than bash comments, I'd suggest instead "Install them manually by running /opt/radixdlt/install_network_utils.sh"

core-rust/build.gradle
```diff
@@ -104,17 +104,17 @@ task buildRustDebug(type: Exec) {
 
 task buildRustForDocker(type: Exec) {
     def target = "x86_64-unknown-linux-gnu"
-    def docker_plattform = "linux/amd64"
+    def dockerPlattform = "linux/amd64"
```
dhedey, line 107: Super minor - but Plattform should be Platform

docs/development/babylon-node-docker-build.png
dhedey: Really nice now, thanks!

dhedey (approved):

core-rust/build.gradle
kofrdx: It's german. :P

docs/development/README.md
```diff
@@ -94,3 +94,39 @@ If you meet this issue, check following configuration options:
...
+docker build . -t radixdlt/babylon-node 
+```
+
+If local testing is required, the artifacts can be produced locally with docker aswell by specifying the build target and outputting the result.
+
+Example core-rust/libcorerust.so:
+
+```
+docker build  . \
+    -t eu.gcr.io/dev-container-repo/babylon-node:local-test-core-rust \
```
shambupujar, line 115: Example uses gcr repo. Probably you can do without that.

Dockerfile
kofrdx: The apt-get clean and rm -rf /var/lib/apt/lists/* reduce the image size after installation. This is done automatically on ubuntu based images. On debian we have to do this ourselves.

The installed packages seem to be dependencies to build our specific rust target.

Dockerfile
kofrdx: Yes

Dockerfile
kofrdx: Same as above. :)

core-rust/build.gradle
```diff
@@ -126,27 +134,25 @@ task buildRustForDocker(type: Exec) {
...
+                .
+
              [ -f $outputDir/${project.property("rustBinaryBuildType")}/libcorerust.so ]
                 """
 
         commandLine 'sh', '-c', dockerMultiStageCmd
     } else{
+        project.logger.lifecycle('Not running in CI!')
         def env_rust_profile = project.property("rustBinaryBuildType") == "debug" ? "dev" : project.property("rustBinaryBuildType")
         def dockerComposeRun = """
```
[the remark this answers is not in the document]
kofrdx: That command uses a way to build which I can not reproduce. It uses volumes to leverage already existing rust libraries to speed up the build. Adding this to the main Dockerfile seems confusing and cluttered. I see its benefit for local development.

shambupujar (approved):

core-rust/build.gradle
```diff
@@ -104,18 +104,26 @@ task buildRustDebug(type: Exec) {
...
 
     // Trigger a custom error message on failure...
     ignoreExitValue true
     if (project.hasProperty('ci')){
+        project.logger.lifecycle('Running in CI: ')
         workingDir rootProject.projectDir
         def outputDir="core-rust/target/$target"
         def rustBuilderRepo = "eu.gcr.io/dev-container-repo/rust-builder-base"
-        def rustCacheRepo = "eu.gcr.io/dev-container-repo/rust-cache-packages"
+        def rustCacheRepo = "europe-west2-docker.pkg.dev/dev-container-repo/eu-cache-repo/babylon-node"
```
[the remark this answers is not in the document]
dhedey: Ahh I see so one's a cache and the other is the actual images.

PR #393, opened 2023-03-31
tweaks: For RCnet-V1

dhedey:
Details:
* Update scrypto version to pull in new costing for package reads
* Minor Core API renames
* Update to Core API and System API schemas for RCNet

This includes lots of changes - mostly just the generated models include the whole API description >_< eugh. Recommended to review in an IDE.


core-rust/core-api-server/src/core_api/handlers/lts/transaction_status.rs
```diff
@@ -106,6 +106,7 @@ fn handle_lts_transaction_status_internal(
         return Ok(models::LtsTransactionStatusResponse {
             intent_status,
             status_description: format!("The transaction has been committed to the ledger, with an outcome of {outcome}. For more information, use the /transaction/receipt endpoint."),
+            committed_state_version: Some(to_api_state_version(txn_state_version)?),
```
jakrawcz-rdx, line 109: I see this added in the LTS version only - intended?

dhedey: Yes - in the sense that it's the only one critical for the RCNet release, and the other one can follow later.

PR #401, opened 2023-04-04
Feature: account change index and associated endpoint

malexandru-rdx:
This PR implements fast index for account entity address => state versions which changed it.
This is needed in order to implement endpoint `/lts/stream/account-transaction-outcomes`.

By default the index is enabled and it can be disabled from config param `extension.account_change_index.enable` or environment variable `RADIXDLT_EXTENSION_ACCOUNT_CHANGE_INDEX_ENABLE` if using the docker image.

There is a lot of room for improvement in an `async` environment (e.g. index building).


core-rust/core-api-server/src/core_api/errors.rs
```diff
@@ -88,6 +88,7 @@ pub(crate) fn not_found_error<E: ErrorDetails>(message: impl Into<String>) -> Re
     }
 }
 
+#[allow(dead_code)]
```
dhedey, line 91: We could probably just remove this? :)

core-rust/core-api-server/src/core_api/handlers/lts/stream_account_transaction_outcomes.rs
```diff
@@ -37,7 +60,45 @@ fn handle_lts_stream_account_transaction_outcomes_internal(
         )));
     }
 
-    let _max_state_version = state_manager.store().max_state_version();
+    let query_account_change_index = state_manager.store().query_account_change_index();
+    if !query_account_change_index.is_enabled() {
+        return Err(server_error(
+            "This endpoint require AccountChangeIndex enabled.",
```
dhedey, line 66: I think we could provide a little more information here:

> This endpoint requires that the `AccountChangeIndex` is enabled on the node. To use this endpoint, you will need to enable the index, delete your database and resync.

core-rust/core-api-server/src/core_api/handlers/lts/stream_account_transaction_outcomes.rs
```diff
@@ -37,7 +60,45 @@ fn handle_lts_stream_account_transaction_outcomes_internal(
...
+    let state_versions = query_account_change_index.get_state_versions(
+        Address::Component(component_address),
+        from_state_version,
+        limit,
+    );
+
+    let committed_transaction_outcomes = state_versions
+        .iter()
+        .map(|state_version| {
+            Ok(to_api_lts_comitted_transaction_outcome(
```
dhedey, line 81: Minor typo - should be: `to_api_lts_committed_transaction_outcome`

core-rust/state-manager/src/store/db.rs
```diff
@@ -338,3 +344,114 @@ impl RecoverableVertexStore for StateManagerDatabase {
         }
     }
 }
+
+pub enum WriteableStateManagerDatabaseAccountChangeIndex<'a> {
```
dhedey, line 348: I'm a little surprised by this - why isn't this part of the StateManagerDatabase?

core-rust/state-manager/src/store/db.rs
```diff
@@ -338,3 +344,114 @@ impl RecoverableVertexStore for StateManagerDatabase {
...
+    InMemory(WriteableInMemoryAccountChangeIndex<'a>),
+    RocksDB(WriteableRocksDBAccountChangeIndex<'a>),
+}
+
+impl WriteableStoreIndexExtension for WriteableStateManagerDatabaseAccountChangeIndex<'_> {
+    fn disable(&mut self) {
```
dhedey, line 354: I'm rather surprised that the index has dynamic enable/disable functionality - you can't dynamically turn on the index else it's incomplete, which would lead to confusion - unless we carefully track which state versions from/to it's been enabled for, and expose that very carefully via the API.

I imagine this is quite complicated to do for now, and we should just ensure either it's enabled or disabled.

I would have assumed this comes from the StateManager settings at start-up time.

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -183,10 +211,15 @@ impl RocksDBStore {
     }
 
     fn add_transaction_to_write_batch(
-        &self,
+        &mut self,
```
dhedey, line 214: Why do we need to take a &mut self here? This feels wrong.

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -284,6 +317,12 @@ impl RocksDBStore {
             .unwrap()
             .map(|pinnable_slice| scrypto_decode(pinnable_slice.as_ref()).unwrap())
     }
+
+    fn put<T: ScryptoEncode>(&self, cf: &RocksDBColumnFamily, key: &[u8], value: &T) {
```
dhedey, line 321: Why are we exposing a generic `put` here? It feels quite dangerous.

I'd rather all updates go through a batch for atomicity.

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -784,3 +823,208 @@ impl RecoverableVertexStore for RocksDBStore {
...
+        self.batch_update_from_receipt(batch, state_version, &transaction_bundle.1);
+
+        batch.put_cf(
+            self.store.cf_handle(&AccountChangeIndexExt),
+            "current_state_version".as_bytes(),
+            state_version.to_be_bytes(),
+        );
+    }
+
+    fn update_from_store(&mut self, start_state_version_inclusive: u64, limit: u64) -> u64 {
```
dhedey, line 868: It feels more natural for this code to live in a separate re-indexing utility, which is invoked at start-up.

Also this functionality would need better logging - because this process could take a long time.

core-rust/state-manager/src/store/db.rs
dhedey: Also how are enable and disable even called?

dhedey (changes requested):
Sorry for the delay looking at this. I'm not too sure on some of the architectural choices here - let's chat about it.

core-rust/core-api-server/src/core_api/handlers/lts/stream_account_transaction_outcomes.rs
malexandru-rdx: Agree more info is nice.

The database does not need to be deleted though (and node resynced).

core/src/main/java/com/radixdlt/RadixNodeModule.java
```diff
@@ -264,7 +264,8 @@ protected void configure() {
     var databasePath = properties.get("db.location", ".//RADIXDB");
     var mempoolMaxSize = properties.get("mempool.maxSize", 50);
     var mempoolConfig = new RustMempoolConfig(mempoolMaxSize);
-    var databaseConfig = new REv2DatabaseConfig.RocksDB(databasePath);
+    var enableAccountChangeIndex = properties.get("extension.account_change_index.enable", false);
```
dhedey, line 267: We'll want to include this on the `default.config.envsubst` - and turn it on for local development in the docker file - and we'll need to talk to Kim to ensure it's included in the full node set-up for a new release of the  Node CLI.

It might even be worth defaulting to `true` to start with.

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -905,80 +910,61 @@ impl<'a> WriteableRocksDBAccountChangeIndex<'a> {
...
-        self.store
-            .get_by_key::<u64>(&AccountChangeIndexExt, "current_state_version".as_bytes())
-            .unwrap_or(0)
-    }
+        while last_processed_state_version < last_state_version {
+            last_processed_state_version = self.update_account_change_index_from_store(
+                last_processed_state_version + 1,
+                MAX_TRANSACTION_BATCH,
+            );
+            info!("Account Change Index updated to {last_processed_state_version}/{last_state_version}");
```
dhedey, line 961: Really like how this looks now, with the semi-regular logging during catch-up 👍

dhedey (approved):
Mostly minor comments - probably the main thing to double check is the `key` which is used for the config - and ensuring the configuration is connected up so that it can default to true for full nodes.

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -841,34 +838,38 @@ impl<'a> WriteableRocksDBAccountChangeIndex<'a> {
...
-        self.batch_update_from_receipt(batch, state_version, &transaction_bundle.1);
+        self.batch_update_account_change_index_from_receipt(
+            batch,
+            state_version,
+            &transaction_bundle.1,
+        );
 
         batch.put_cf(
-            self.store.cf_handle(&AccountChangeIndexExt),
+            self.cf_handle(&ExtensionsData),
```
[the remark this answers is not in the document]
malexandru-rdx: Oops, my bad. Thanks for catching :)

core-rust/state-manager/src/store/db.rs
```diff
@@ -88,15 +88,10 @@ use crate::{
...
-use super::in_memory::QueryableInMemoryAccountChangeIndex;
-use super::in_memory::WriteableInMemoryAccountChangeIndex;
-use super::rocks_db::QueryableRocksDBAccountChangeIndex;
-use super::rocks_db::WriteableRocksDBAccountChangeIndex;
-
 #[derive(Debug, Categorize, Encode, Decode, Clone)]
 pub enum DatabaseConfig {
-    InMemory,
-    RocksDB(String),
+    InMemory(bool),
```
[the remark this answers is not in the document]
malexandru-rdx: With the next configurable behaviour bit I will group them into a struct.

core/src/main/java/com/radixdlt/RadixNodeModule.java
malexandru-rdx: It's added to `default.config.envsubst` (and `Dockerfile.core`) in this PR already.

Should I make it default to `true` everywhere?

dhedey: Ah sorry I missed that! I think let's default it to true for now, and we can consider a more clear Validator / Full Node distinction for RCnet-V2

PR #402, opened 2023-04-04
Initial work towards Babylon genesis from Olympia

LukasGasior1:
Supersedes https://github.com/radixdlt/babylon-node/pull/373 (because of missing commit signatures).


jakrawcz-rdx (approved):
If `these are exactly the same changes we've already seen` then 🟢

LukasGasior1:
> If these are exactly the same changes we've already seen then green_circle

yes, trust me, nothing to see here :smiling_imp:

PR #405, opened 2023-04-07
Refactor: Separate the StateManagerDatabase from StateManager

jakrawcz-rdx:
This is a pure refactoring, with the only potential behavior difference resulting from _less_ locking on Rust `StateManager`. (meaning: lock safety is probably the most important aspect to focus on during code review here).

What this PR does, in short: _(recommended reviewing commit-by-commit)_
1. The first commit introduces a java `Database` class, which is similar to java `StateManager` class.
    - its responsibility is to hold ownership of rust `StateManagerDatabase` instance (similar to how java `StateManager` class holds ownership of rust `StateManager` instance)
    - the rust `StateManagerDatabase` instance lives inside an `Arc<RwLock<...>>` of `JNIDatabase` struct (similar to how rust `StateManager` instance lives inside an `Arc<RwLock<...>>` of `JNIStareManager`)
    - previously, the rust `StateManager` simply owned the `StateManagerDatabase`; now it has `Arc<RwLock<...>` of it.
    - the rust `StateManager` keeps using its `store` the same way, but now it acquires an appropriate (read or write) lock for each usage.
    - previously, the java `REv2TransactionAndProofStore` was using rust `StateManager` (where "using" means "it did `state_manager.store()` and then really was only using the `StateManagerDatabase` obtained this way); now, the java `REv2TransactionAndProofStore` is using `StateManagerDatabase` directly (ofc acquiring the lock), so _`StateManager` is no longer its dependency (and `StateManager`'s lock doesn't have to be acquired for these read-outs anymore)_.
2. The second commit takes the `VertexStoreRecovery` and treats it the same way as `REv2TransactionAndProofStore` was treated above:
    - the `VertexStoreRecovery` was always doing only "read vertex store" operation; in this PR, I moved the sibling operation "save vertex store" (from `StateManager`) to `VertexStoreRecovery` as well.
    - previously, the java `VertexStoreRecovery` was using rust `StateManager` (where "using" means "it did `state_manager.store()` and then really was only using the `StateManagerDatabase` obtained this way); now, the java `VertexStoreRecovery` is using `StateManagerDatabase` directly (ofc acquiring the lock), so _`StateManager` is no longer its dependency (and `StateManager`'s lock doesn't have to be acquired for vertex store operations anymore)_.
3. The third commit takes the `get_component_resources(), get_validator_info(), get_epoch()` queries out of `StateManager` and puts them into a new `RustStateQueryService`, which again, depends only on the `Database` alone, _so `StateManager`'s lock doesn't have to be acquired for these read-outs anymore_

**Why I think this change is beneficial?**

Apart from the "less locking" (performance) aspect indicated above I truly believe that conceptually, all these services extracted out are easier to reason about. They are quite separate, and (as shown) they can do their narrow work with just one narrow dependency (on the database alone rather than on the entire state manager).
In fact, after these changes, we can see that `RustStateComputer` itself has _also_ became narrow - now its JNI dependency surface only contains a few methods that I would expect from a "state computer which runs transactions": `verify(), prepare(), commit()`. 
(well, actually, it also still holds a `RustMempool` and exposes its methods, which still pollutes its otherwise-narrow API - that's a topic for a separate refactoring [similar to this PR!], which actually will reduce locking related to mempool sync, [which is one of the core goals of my current foreground effort](https://radixdlt.atlassian.net/wiki/spaces/RPNV1/pages/3031891977/RNP-9+-+New+MempoolSync+Rules?focusedCommentId=3033956353#%3Ayellow_star%3A-Core-Goals))


jakrawcz-rdx:
After looking at some next steps that should follow this PR _[in order to truly optimize the locking on Rust side]_ I noticed that the approach taken here would result in too-much-granularity at a rather inconvenient JNI layer _[i.e. having to touch both Java and Rust per every change]_.

As an alternative, I suggest https://github.com/radixdlt/babylon-node/pull/407, and I will gladly abandon this PR (unless someone convinces me it is really better and worth the extra changed lines).

PR #406, opened 2023-04-13
Add integrator friendly docker compose

malexandru-rdx:
This PR adds:

1.  `node.key.create_if_missing` config option which when enabled, if no node key is provided, a new one is created on the fly - this is needed to remove the step of creating a private key when wanting to quickly spin up a fullnode. Note: Hardcoded node key would not work since the protocol does not guarantee all IPs (or which IP) under same node key
2. Add `testnet-node` - docker-compose file to connect to RCv1 testnet for integrators  
3. Remove unused `docker/Dockerfile.build` and add multi-stage `Dockerfile`


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "java.configuration.updateBuildConfiguration": "disabled"
```
dhedey, line 2: This likely needs to be explicitly removed - it shouldn't be source controlled given the .gitignore.

core/src/main/java/com/radixdlt/keys/PersistedBFTKeyManager.java
```diff
@@ -77,8 +77,18 @@
 public final class PersistedBFTKeyManager {
   private final ECKeyPair ecKeyPair;
 
-  public PersistedBFTKeyManager(String nodeKeyPath) {
-    this.ecKeyPair = loadNodeKey(nodeKeyPath);
+  public PersistedBFTKeyManager(String nodeKeyPath, boolean createNodeKeyIfMissing) {
+    ECKeyPair nodeEcKeyPair;
+    try {
+      nodeEcKeyPair = loadNodeKey(nodeKeyPath);
```
dhedey, line 83: I don't think is ideal - if this node restarts, its key will be lost - which I don't think we want.

Instead, one of the `Keys.readKey` takes a `create` parameter - which creates a key at the given location - we should be using that.

That way the key will be persisted across docker image restarts.

docker/Dockerfile.core
```diff
@@ -43,7 +43,8 @@ ENV RADIXDLT_HOME=/home/radixdlt \
     RADIXDLT_SYSTEM_API_BIND_ADDRESS=0.0.0.0 \
     RADIXDLT_PROMETHEUS_API_PORT=3335 \
     RADIXDLT_PROMETHEUS_API_BIND_ADDRESS=0.0.0.0 \
-    RADIXDLT_NETWORK_ID=240
+    RADIXDLT_NETWORK_ID=240 \
+    RADIXDLT_NODE_KEY_CREATE_IF_MISSING=false
```
dhedey, line 47: I don't think we necessarily need to put this here, but I guess it shouldn't hurt :)

testnet-node/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+      context: ..
+      dockerfile: Dockerfile
+    # image: radixdlt/babylon-node:rcnet-v1-c6360105d-build-2
+    init: true
+    ulimits:
+      nofile:
+        soft: 65536
+        hard: 65536
+      memlock: -1
+    # restart: unless-stopped
```
dhedey, line 14: Should this be removed? Or if it's an option that developers might want to use, we should add a comment explaining why they might want to change this

testnet-node/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
+version: '2.2'
+services:
+  testnet-node-core:
+    build:
+      context: ..
+      dockerfile: Dockerfile
+    # image: radixdlt/babylon-node:rcnet-v1-c6360105d-build-2
```
dhedey, line 7: Might warrant a comment to explain this is coming soon?

testnet-node/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
+version: '2.2'
```
dhedey, line 1: I imagine we probably want:

* A clear prompt in the main README: "If you're looking to develop an integration to run a Core API node, see the [./testnet-node](testnet-node folder)."
* A clear README in this folder explaining that this is intended for development; how to use this docker file, where to find docs for integrations, and where to find docs for configuring a node in production.

dhedey (changes requested):
So this PR confusingly has an old version of Kim's PR in it.

I think what we should do is:
* Rebase Kim's PR onto `release/rcnet-v1` and merge into `release/rcnet-v1`. 
* Rebase/merge this into `release/rcnet-v1` after Kim's PR has been merged. (for the time being, this PR could have Kim's PR as a base to avoid confusion about what's under review)
* Then merge everything into develop.

core/src/main/java/com/radixdlt/keys/PersistedBFTKeyManager.java
malexandru-rdx: > I don't think is ideal - if this node restarts, its key will be lost - which I don't think we want.
I am viewing from the perspective where anything using private keys should be intentional and explicitly manage by the user - i.e. you want (or need?) your private key to persist - create and manage it yourself :). I don't want people to rely on a private key generated inside the docker.

This is a mere convenience functionality (I don't really care about details, just give me a quick way to scoop in the network).

testnet-node/README.md
```diff
@@ -0,0 +1,29 @@
+## Testnet Node
+
+This is the easiest way to get a Babylon fullnode up and running, ready for integration development.
+A node in the network is identified by it's public key and there can be multiple addresses associated to a public key.
```
dhedey, line 4: Some of this is really good, but some of it is a little too detailed.

I'd recommend replacing all of the public key big with this:
```
A node in the network is identified by its public key. A keystore containing a keypair is automatically generated on startup and stored inside a docker volume called `key-data`, protected by default with password `"password"`.
```

The ledger stuff is good :)

testnet-node/README.md
```diff
@@ -0,0 +1,29 @@
...
+This means you need to generate a new key (and keep it to yourself). Luckly you don't have do anything as long as env
+variable `RADIXDLT_NODE_KEY_CREATE_IF_MISSING` is set to `true`. The only takeaway is that by default, the private key
+is generated and stored inside of docker volume `key-data` (please double check before removing this volume), keystore encrypted
+with the password provided via `RADIX_NODE_KEYSTORE_PASSWORD`.
+The initial syncing of the whole ledger state can take a while. To avoid having to go through this every time
+something is changed in `docker-compose.yml`, we setup `ledger-data` volume for persistance.
+Sometimes a ledger wipe is needed, in which case you can run: `docker volume rm testnet-node_ledger-data` to remove (only) `ledger-data` volume.
+
+### Getting started
+1. Install `docker` and `docker-compose`
```
dhedey, line 15: What version of docker was this tested on?

We say this in the development readme 👍 
```
- Docker version 20.10+
- `docker-compose` version 1.25+
```

testnet-node/README.md
dhedey: I'd also suggest mentioning:
* `The node connects to the latest public testnet (as of April/May 2023, this is RCnet V1, known by its logical name kisharnet).`
* `[Documentation for integrators is available here](https://docs.google.com/document/d/1cjc7_alyzIb2QQIGGn1PEpJyjrMRZYHq3VwkOXRP8J0/edit#).`

testnet-node/README.md
```diff
@@ -0,0 +1,29 @@
...
+2. Run `docker-compose up` in this folder
+
+### Local builds
```
dhedey, line 18: Nice 👍

README.md
```diff
@@ -19,6 +19,7 @@ Here we have:
 - [shell](shell): Examples for how to run Radix shell, which can enable spinning up temporary interactive
    nodes. The Radix Shell code itself is in [cli-tools](cli-tools).
 - [docker](docker): An option for running a network of nodes locally. You can also use the "Run Single Validator" IntelliJ option.
+- [testnet-node](testnet-node): The easiest way to setup a development environment for integrators (check the [README](testnet-node/README.md))
```
dhedey, line 22: I'd suggest to make it really easy to find, we should have an ## Integrators section even above `## SubDirectories` - and say - if you're looking to run a node for developing an integration, see `[details on running a testnet node for development](./testnet-node)` and the `[Babylon documentation for exchanges and integrators](https://docs.google.com/document/d/1cjc7_alyzIb2QQIGGn1PEpJyjrMRZYHq3VwkOXRP8J0/edit#)`

dhedey (approved):
Really nice. A few final comments 👍

PR #407, opened 2023-04-14
Refactor: Separate the StateManagerDatabase from StateManager (Rust-only)

jakrawcz-rdx:
⚠️ this is an ALTERNATIVE to https://github.com/radixdlt/babylon-node/pull/405

This PR achieves the same separation of concerns (and - in future - of locks!) on Rust's side, while not touching the Java side. (it means literally 50% less code changes)

The motivation stays exactly the same as in https://github.com/radixdlt/babylon-node/pull/405#issue-1658765012 
The implementation is also very similar, with just one difference: instead of splitting the ownership at Java side, we still have just one Java `StateManager` and one Rust `JNIStateManager`. The split here happens at `JNIStateManager`, which separately owns an `ActualStateManager` and a `StateManagerDatabase`. This way, some operations may e.g. lock only the `StateManagerDatabase`, without locking `ActualStateManager.

### DIFF
--- a/core-rust-bridge/src/main/java/com/radixdlt/sbor/StateManagerSbor.java
+++ b/core-rust-bridge/src/main/java/com/radixdlt/sbor/StateManagerSbor.java
@@ -79,6 +79,8 @@ import com.radixdlt.statecomputer.commit.*;
 import com.radixdlt.statemanager.*;
 import com.radixdlt.transaction.CommittedTransactionStatus;
 import com.radixdlt.transaction.ExecutedTransaction;
+import com.radixdlt.transaction.TxnsAndProof;
+import com.radixdlt.transaction.TxnsAndProofRequest;
 import com.radixdlt.transactions.RawLedgerTransaction;
 import com.radixdlt.transactions.RawNotarizedTransaction;
 import com.radixdlt.utils.UInt16;
@@ -129,6 +131,8 @@ public final class StateManagerSbor {
     MempoolError.registerCodec(codecMap);
     CommittedTransactionStatus.registerCodec(codecMap);
     ExecutedTransaction.registerCodec(codecMap);
+    TxnsAndProofRequest.registerCodec(codecMap);
+    TxnsAndProof.registerCodec(codecMap);
     PublicKey.registerCodec(codecMap);
     ECDSASecp256k1PublicKey.registerCodec(codecMap);
     EdDSAEd25519PublicKey.registerCodec(codecMap);
--- a/core-rust-bridge/src/main/java/com/radixdlt/statecomputer/RustStateComputer.java
+++ b/core-rust-bridge/src/main/java/com/radixdlt/statecomputer/RustStateComputer.java
@@ -73,14 +73,12 @@ import com.radixdlt.mempool.RustMempool;
 import com.radixdlt.monitoring.LabelledTimer;
 import com.radixdlt.monitoring.Metrics;
 import com.radixdlt.monitoring.Metrics.MethodId;
-import com.radixdlt.recovery.VertexStoreRecovery;
 import com.radixdlt.rev2.ComponentAddress;
 import com.radixdlt.rev2.Decimal;
 import com.radixdlt.rev2.ValidatorInfo;
 import com.radixdlt.sbor.Natives;
 import com.radixdlt.statecomputer.commit.*;
 import com.radixdlt.statemanager.StateManager;
-import com.radixdlt.transaction.REv2TransactionAndProofStore;
 import com.radixdlt.transactions.RawNotarizedTransaction;
 import com.radixdlt.utils.UInt64;
 import java.util.List;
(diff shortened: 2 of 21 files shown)


dhedey (commented):
I've taken a look - I approve of the end goal, but I'm not sure I quite agree with adding more locks at the JNI layer... Although this might be a good stepping stone.

My vision from the start here is that I'd rather each leaf service (eg database/mempool/metrics) owned its own locks / had interior mutability - that way:
* We can get rid of the lock on the state manager completely
* We don't need to pollute everywhere with locks.
* We don't need to worry about cross-service deadlocks because each service is independent and only can hold its own interior lock/s
* I feel it also can more naturally split up into separate actors / modules when we come to rewrite in rust or introduce async.

EG we'd have a &DatabaseManager which owned a RwLock around a Database, and has each method take out either a read/write lock in a single file - rather than this being at the JNIStateManager level.

jakrawcz-rdx:
Either I don't understand you fully, or you would like the reality (of our domain) to be simpler than it is (but it cannot).

> I'd rather each leaf service (eg database/mempool/metrics) owned its own locks

Yeah this would make everything much easier, but I think it is only available for very simple use-cases.
(Please note: if such approach was available for all use-cases, then it would be an ultimate way to "solve" locking: people would never run into deadlocks, and the world wouldn't really need all these different synchronization primitives [i.e. in java, the simplest method-level `synchronized` keyword would suffice] - that's how I believe it has serious limitations or performance implications)

Our use-cases are almost definitely not simple enough to use this approach, and one example is hidden here:

> EG the StateManager would own a &DatabaseManager (or Arc) with the DatabaseManager being the thing which owns a RwLock around a Database, and has each method take out either a read/write lock in a single file - rather than this being at the JNIStateManager level.

How would it work for a `StateManager.commit(tx)` operation?
```
StateManager {
  dbm: Arc<DatabaseManager>
  ...
  pub fn commit(tx) {
    let curr_ver = self.dbm.get_top_of_ledger_version();
    ... // whatever processing
    self.dbm.save_tx_by_state_ver(curr_ver + 1, tx);
  }
}
```

When you say that only `DatabaseManager` handles the locks of the actual `Database`, I cannot help but imagine the above ☝️ 
(I mean: sure, `DatabaseManager.get_top_of_ledger_version()` correctly uses a `Database.read()` lock, and the `DatabaseManager.save_tx_by_state_ver()` correctly uses a `Database.write()` lock - but it doesn't provide any synchronization between these 2 calls [which is needed to prevent the race conditions between two threads that can compute the same `curr_ver + 1` value])

I use the `commit()` example since it shows a nice `load, inc, save` pattern - I acknowledge that because of how consensus (and ledger sync) works, we probably never would get 2 threads racing here.
But I believe we encounter the same problems on potentially-more-racing use-cases (e.g. `preview()`, which needs to do `Database.read()` lock during the execution of `StateManager`'s logic).

Maaaybe we could solve each of these problems by means other than logic (e.g. optimistic concurrency on `commit()`; or versioned historical substate_store reads on `preview()`). But it would have heavy performance implications, and it would introduce its own complexity.

Yeah, so either that ☝️ , or I didn't get your entire picture :D

jakrawcz-rdx:
> So I'm very happy to chat and have you convince me this PR is the right pattern, or a good stepping stone to a better architecture ✨

Ok, I scheduled a call; here is some illustration of the end-state that I hope to get:
```

JNIStateManager {
  state_manager: StateManager, // no lock on its own, as you wanted
  database: Arc<RwLock<StateManagerDatabase>>, // already in this very PR
  mempool: Arc<RwLock<SimpleMempool>>, // example of a part worth splitting out (and still needing locks *)
  metrics_registry: PrometheusRegistry, // example of a part worth splitting out (and not needing locks)
  ...
}

VertexStoreManager { // a part extracted from StateManager; name up for discussion
  database: Arc<RwLock<StateManagerDatabase>>, // it needs just this dep

  fn load/save(vertex_store) {...}
}

MempoolManager { // a part extracted from StateManager; name up for discussion
  mempool: Arc<RwLock<SimpleMempool>>, // it needs just this dep

  fn add_to_mempool(tx) {...}
}

StateManager { // it stays, but after extracting lots of parts out, it could be named e.g. TransactionExecutor
  database: Arc<RwLock<StateManagerDatabase>>, // it still references this one
  mempool: Arc<RwLock<SimpleMempool>>, // yup, and that one

  fn preview/prepare/commit(tx) {
    // a logic that needs a lock on database, and mempool too, and on other things, possibly overlapping
  }
}
```

There is 1x `*` marked above:
Somewhere along the refactorings, it may turn out that e.g. the `SimpleMempool` does NOT need to expose its locks (i.e. all its operations use internal locks, and we have no use-cases of `load + update + save` between them - but I simply don't know it yet). I guess this would partially fulfil [your dream](https://github.com/radixdlt/babylon-node/pull/407#pullrequestreview-1386555375). If this happens, then we ofc do not need the `Arc<RwLock<..>` wrapping it anywhere (i.e. it will become another part similar to `metrics_registry`; but I am sure _not all_ parts can become like that).

dhedey (approved):
Thanks for the chat, I like the direction, this looks good 👍

PR #408, opened 2023-04-15
Optimize /lts/state/account-resource-balance

malexandru-rdx:
Optimize `/lts/state/account-resource-balance` by reading the resource vault directly instead of using recursive `dump_component_state`.

### DIFF
--- a/core-rust/core-api-server/src/core_api/handlers/lts/state_account_resource_balance.rs
+++ b/core-rust/core-api-server/src/core_api/handlers/lts/state_account_resource_balance.rs
@@ -1,10 +1,12 @@
 use crate::core_api::*;
-use radix_engine::types::{ComponentAddress, Decimal, ResourceAddress};
-use state_manager::{
-    jni::state_manager::ActualStateManager,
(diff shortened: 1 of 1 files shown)


core-rust/core-api-server/src/core_api/handlers/lts/state_account_resource_balance.rs
```diff
@@ -47,13 +49,74 @@ fn handle_lts_state_account_fungible_resource_balance_internal(
...
+                                        &mapping_context,
+                                        &fungible_resource_address,
+                                    ),
+                                    amount: to_api_decimal(&Decimal::zero()),
+                                },
+                            ),
+                        })
+                    }
+                    _ => {
+                        // This should not happen
```
dhedey, line 90: I believe this can happen - we've only validated that the address is a component address, not an account address on line 48. So we should either handle this at validation time or at this time - but either way, we should return a client error for wrong address, not a server error.

dhedey (approved):
Nice! :)

PR #410, opened 2023-04-17
Refactor: Splitting the immutable part out from StateManager

jakrawcz-rdx:
This continues the effort started at https://github.com/radixdlt/babylon-node/pull/407 (towards an end-state described at https://github.com/radixdlt/babylon-node/pull/407#issuecomment-1511005042).

In this PR, I pull an _example piece_ of an immutable state up (from `StateManager` to `JNIStateManager`). This way, the callers that only need to access the immutable state (e.g. CoreApi) no longer need to obtain the `StateManager`'s lock (in fact, they don't need any lock). 

In most controllers, this has only limited effect (because soon after reading the immutable state, the controller logic wants something more complex from `StateManager` anyway). But in some (e.g. `handle_status_network_configuration()`) we now avoid locking altogether.

More obvious performance gains will show up in a consecutive PR (enabled by this one), in which I access the `database` directly from applicable controllers (i.e. avoiding the `StateManager`'s lock, and locking only the `StateManagerDatabase`).


malexandru-rdx (approved):
👍

PR #412, opened 2023-04-17
Splitting the Rust metrics into logical parts + Moving Registry ownership out of StateManager

jakrawcz-rdx:
This is the next PR towards the end-state described at https://github.com/radixdlt/babylon-node/pull/407#issuecomment-1511005042.

Here I move the Prometheus' `Registry` up (to `JNIStateManager`), and this way, the `do_prometheus_metrics()` native call does not have to acquire StateManager's lock (or any lock at all, since the `Registry` is thread-safe already, **according to my quick glance through its sources, which I encourage to verify during the review**).

Apart from that, I refactor the Rust metrics a bit (since maintaining nice metric infra is my hobby) and split it into logical parts (in preparation for the "split mempool out" PR which should come soon).


malexandru-rdx (approved):

jakrawcz-rdx:
thanks for taking a look 🙇‍♂️

PR #411, opened 2023-04-17
Accessing the database directly from CoreApi (avoiding StateManager lock)

jakrawcz-rdx:
This is the next PR towards the end-state described at https://github.com/radixdlt/babylon-node/pull/407#issuecomment-1511005042, building on top of https://github.com/radixdlt/babylon-node/pull/410 (though not strictly required - just wanna avoid some conflicts).

In this PR, I access the `database` directly from CoreApi, in many cases avoiding a lock on `StateManager` (in fact, a `StateManager.store()` accessor vanished!).

Note: here we see that a certain pattern forms - the `network + state_manager + database` fields travel together. If anyone suggests a good, _not-yet-taken_ name for this group (remember it may grow!), I promise to refactor that into a struct 🤞.

PR #413, opened 2023-04-18
Splitting the mempool out of StateManager

jakrawcz-rdx:
This is the next step towards an end-goal https://github.com/radixdlt/babylon-node/pull/407#issuecomment-1511005042, very clearly an intermediate step (see `TODO`s), which will help me achieve more un-lock-ing later.

### DIFF
--- a/core/src/main/java/com/radixdlt/ledger/StateComputerLedger.java
+++ b/core/src/main/java/com/radixdlt/ledger/StateComputerLedger.java
@@ -154,7 +154,7 @@ public final class StateComputerLedger implements Ledger, ProposalGenerator {
   private final Metrics metrics;
   private final LedgerAccumulator accumulator;
   private final LedgerAccumulatorVerifier verifier;
-  // TODO(resolve during review): Is it still needed? Was it ever needed?
+  // TODO: This does not seem needed - remove in a narrow-focus PR.
   private final Object lock = new Object();
   private final TimeSupplier timeSupplier;
 


core-rust-bridge/src/main/java/com/radixdlt/mempool/ProposalTransactionsRequest.java
```diff
@@ -0,0 +1,81 @@
...
+
+package com.radixdlt.mempool;
+
+import com.google.common.hash.HashCode;
+import com.radixdlt.sbor.codec.CodecMap;
+import com.radixdlt.sbor.codec.StructCodec;
+import com.radixdlt.utils.UInt32;
+import java.util.List;
+
+/** A request for listing a batch of transactions for proposal. */
```
jakrawcz-rdx, line 73: _(a note from the author)_

This is only a drive-by refactor for cleaner lambda in one place.

core/src/main/java/com/radixdlt/ledger/StateComputerLedger.java
```diff
@@ -154,6 +154,7 @@ void commit(
   private final Metrics metrics;
   private final LedgerAccumulator accumulator;
   private final LedgerAccumulatorVerifier verifier;
+  // TODO(resolve during review): Is it still needed? Was it ever needed?
   private final Object lock = new Object();
```
jakrawcz-rdx, line 158: _(a note from the author)_

Actually @malexandru-rdx brought my attention to this: apparently all my efforts to reduce "locking on too-wide scopes" on Rust side are killed by this lock on Java-side. Does anyone know a specific purpose of this extra lock here? Was there some weird effect going on, or is this just someone being too eager? Or is this a remnant from the days when the "state manager" was happening only on Java side?

Based on https://github.com/radixdlt/babylon-node/blame/29c163a2562b30d52e7db2a43e969a3847b042f5/radixdlt-core/radixdlt/src/main/java/com/radixdlt/ledger/StateComputerLedger.java and https://github.com/radixdlt/babylon-node/commit/e355b3ddebdf487b39da448b963a3ed1fdd5979e I suspect it is the "old sync from java days" case, but I need confirmation before I delete this.

core-rust/core-api-server/src/core_api/server.rs
```diff
@@ -78,13 +78,15 @@ use state_manager::jni::state_manager::ActualStateManager;
...
 use handle_status_network_configuration as handle_provide_info_at_root_path;
+use state_manager::simple_mempool::SimpleMempool;
 use state_manager::store::StateManagerDatabase;
 
 #[derive(Clone)]
 pub(crate) struct CoreApiState {
     pub network: NetworkDefinition,
     pub state_manager: Arc<RwLock<ActualStateManager>>,
     pub database: Arc<RwLock<StateManagerDatabase>>,
+    pub mempool: Arc<RwLock<SimpleMempool>>,
```
jakrawcz-rdx, line 89: _(a note from the author)_

So, @malexandru-rdx , taking https://github.com/radixdlt/babylon-node/pull/411#pullrequestreview-1390194356 into account - is this the moment to start pushing a `CoreApiState` instance across the stack? (other ideas welcome too)

malexandru-rdx: I personally like to see verbose code when setting up different connected "services" and also I'd rather read it as a flatter config file than having multiple layers of abstraction.
In this case, other than passing `CoreApiState` directly to `create_server` I would not do anything else (not sure about this either).

core/src/main/java/com/radixdlt/ledger/StateComputerLedger.java
malexandru-rdx: I would still merge this PR first before figuring out this lock.
Definitely the lock can be split between `prepare` + `commit` and `mempooladd` + `getTransactionsForProposal` and get (some of?) the benefits of your refactors.
I'm wondering if there is any reason not to remove it completely.

malexandru-rdx (approved):
Nice 👍

PR #417, opened 2023-04-18
Core API Typescript SDK

dhedey:
## Changes
* Fixes some issues with the LTS API not being consistent or not aligning with the schema
* Adds a Core API Typescript SDK, a nice wrapper, tests, docs, and a deployment workflow

## Reviewing suggestions

In terms of reviewing, I'd recommend:
* Ignoring any files in `generated` folders
* Checking out the typescript sdk README: https://github.com/radixdlt/babylon-node/blob/faa4150a69eb4d9255d96c84cd727c3ecc4bd7e8/sdk/typescript/README.md
* Check out `lts.ts` and `lts.test.ts`


kofrdx (approved):

.github/workflows/publish-typescript-sdk.yml
```diff
@@ -0,0 +1,46 @@
...
+      package_version_number:
+        description: "Package version number"
+        required: true
+  # For now, let's just deploy it manually, to avoid issues with needing to keep its version numbers
+  # in-sync with the node, which doesn't really make sense
+  # release:
+  #   types: [published]
+
+jobs:
+  publish-gateway-sdk-npmjs:
```
0xOmarA, line 15: Minor: do you mean core-api SDK?

sdk/typescript/lib/generated/models/LtsTransactionStatusRequest.ts
```diff
@@ -0,0 +1,75 @@
...
+ * @export
+ * @interface LtsTransactionStatusRequest
+ */
+export interface LtsTransactionStatusRequest {
+    /**
+     * The logical name of the network
+     * @type {string}
+     * @memberof LtsTransactionStatusRequest
+     */
+    network: string;
```
0xOmarA, line 27: Out of curiosity, why does this endpoint take the logical name of the network? Isn't the state-manager already aware of it?

0xOmarA (approved):
Great work on this PR! Love the `LTS` class and the simple interface that it provides. I've left a number of comments but they're all minor.

sdk/typescript/lib/subapis/lts.ts
```diff
@@ -0,0 +1,198 @@
...
+  | {
+      result: "MempoolFull";
+      details: LtsTransactionSubmitMempoolFullErrorDetails;
+      error: ResponseError;
+    };
+
+/**
+ * Wraps the lower-level `LTSApi` - which can be accessed with `innerClient` for advanced use cases.
+ */
+export class LTS {
```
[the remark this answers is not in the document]
0xOmarA: I see that the expected usage is something like `coreApiClient.LTS` so please ignore the above comment.

sdk/typescript/lib/generated/models/LtsTransactionStatusRequest.ts
malexandru-rdx: It's mainly to ensure explicit intent and that nobody is mistakenly using the wrong network (all endpoints have this guardrail).

malexandru-rdx (approved):

sdk/typescript/lib/generated/models/LtsTransactionStatusRequest.ts
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * The hex-encoded transaction intent hash. This is known as the Intent Hash, Transaction ID or Transaction Identifier for user transactions. This hash is `Blake2b-256(compiled_intent)`
+     * @type {string}
+     * @memberof LtsTransactionStatusRequest
+     */
+    intent_hash: string;
```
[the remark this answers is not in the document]
dhedey: For API endpoints specifically, I personally quite like the models matching the specifications 1-for-1, but it's a personal preference.

I also have a personal preference for JSON objects over classes for data in Typescript, and using a data-split-from-code approach rather than a class/object, domain-driven-development model - but I'm not 100% sure why... Possibly because classes in early typescript were harder to type, and classes in JS have lots of gotchas and pain regarding `this` and just generally are (historically) full of minefields.

sdk/typescript/lib/generated/models/LtsTransactionStatusRequest.ts
dhedey: Right - I had this discussion with Josh wanting to remove it about 9 months ago, but Josh felt very strongly that we should require it on each endpoint, at least for the Core API where you're often querying a node hostname which doesn't specify the network.

Whereas on the Gateway, the network name is normally in the title so it's harder to mess up, so he was happy for us to drop it there.

sdk/typescript/lib/subapis/lts.ts
```diff
@@ -0,0 +1,198 @@
...
+  constructor(public innerClient: LTSApi, public logicalNetworkName: string) {}
```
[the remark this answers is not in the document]
dhedey: This is intended to be created by the CoreApiClient which already has it available.

PR #419, opened 2023-04-19
Simplify and improve local rust-builder performance

dhedey:
By using the volumes instead of mounts we get big local speed-ups and simplifications to what was the rust-builder.

I've also cleared up the gradle build, Dockerfile and other scripts.

But notably the local build is now only used for `./gradlew deb4docker` - which isn't even used by `./docker/scripts/rundocker.sh` which makes this a little pointless :)

But if we could somehow circumvent all of the lib-builder branch of the `Dockerfile` locally and just copy it from the rust build, that would be quite a win for running local docker networks...


shambupujar (approved):

kofrdx:
./gradlew deb4docker is currently failing for me. Let me see if I can fix it.

The Docker build works though.

kofrdx (approved):

shambupujar:
Looks good and the images are getting built

PR #420, opened 2023-04-19
Flakey Test: Using a static max adversary set

jakrawcz-rdx:
A subset of `F=(N-1)/3` adversaries is assumed to be static across hotstuff execution (i.e. across 3 rounds).

This assumption is not met by our `NetworkDroppers.fRandomProposalsPerRoundDropped()` util (which changes the set every round). With sufficiently large network (`N>=5` on CI worker, `N>=10` on a laptop), this easily leads to unfortunate subset selections / timings (further complicated by pipelined nature of hotstuff), which cause timeouts (which are asserted in these tests).

The easiest way to fix is to have the subset `F` static across the entire test execution - and this PR applies it.
Locally, I was also able to handcraft some awkward code which changed the `F'=(N-1)/3 - 2` only by 1 node each round (i.e. there were still `N - (N-1)/3` of the same non-adversary validators in every consecutive 3 rounds) - and this also made the test stable (although I am almost sure I was able to catch a failure once, with `N=20`; possibly due to other reasons?).

Anyways: we still have another test, `OutOfOrderTest`, which uses the too-random `NetworkDroppers.fRandomProposalsPerRoundDropped()` and asserts liveness and safety. (i.e. it passes because it does not assert on timeouts)
So I think we are fine with testing using the static adversary set here.

### DIFF
--- a/core/src/integration/java/com/radixdlt/integration/steady_state/simulation/consensus/FProposalsPerRoundDropperTest.java
+++ b/core/src/integration/java/com/radixdlt/integration/steady_state/simulation/consensus/FProposalsPerRoundDropperTest.java
@@ -104,7 +104,8 @@ import org.junit.runners.Parameterized.Parameters;
 public class FProposalsPerRoundDropperTest {
   @Parameters
   public static Collection<Object[]> testParameters() {
-    return Arrays.asList(new Object[][] {{4}, {10}});
+    // TODO(only for CI testing): bring back the `{4}, {10}`
+    return Arrays.asList(new Object[][] {{4}, {4}, {4}, {4}, {4}, {10}, {10}, {10}, {10}, {10}});
   }
 
   private final Builder bftTestBuilder;
@@ -124,10 +125,7 @@ public class FProposalsPerRoundDropperTest {
                     SafetyRecoveryConfig.MOCKED,
                     ConsensusConfig.of(5000, 0L),
                     LedgerConfig.mocked(numNodes)))
-            .addTestModules(
-                ConsensusMonitors.safety(),
-                ConsensusMonitors.vertexRequestRate(75), // Conservative check
-                ConsensusMonitors.noTimeouts());
+            .addTestModules(ConsensusMonitors.safety(), ConsensusMonitors.noTimeouts());
   }
 
   /**
--- a/core/src/integration/java/com/radixdlt/integration/steady_state/simulation/consensus/FProposalsPerRoundDropperTest.java
+++ b/core/src/integration/java/com/radixdlt/integration/steady_state/simulation/consensus/FProposalsPerRoundDropperTest.java
@@ -104,7 +104,10 @@ import org.junit.runners.Parameterized.Parameters;
 public class FProposalsPerRoundDropperTest {
   @Parameters
   public static Collection<Object[]> testParameters() {
-    return Arrays.asList(new Object[][] {{4}, {10}});
+    return Arrays.asList(
+        new Object[][] { // TODO(wip): bring back just the 4, 10.
+          {4}, {4}, {4}, {4}, {4}, {5}, {5}, {10}, {10}, {10}, {20}
+        });
   }
 
   private final Builder bftTestBuilder;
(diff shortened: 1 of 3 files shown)


jakrawcz-rdx:
Super funny:

The `FProposalsPerRoundDropperTest.given_incorrect_module_where_vertex_sync_is_disabled__then_test_should_fail_against_drop_proposal_adversary` was failing because of `Expecting Optional to contain a value but it was empty.` (see https://rdxworks.slack.com/archives/C01M90Y2448/p1681890434408999).

And now, with that fixed, I see on CI that its sibling fails because of `CONSENSUS_VERTEX_REQUEST_RATE`:
```
FProposalsPerRoundDropperTest dropping_sync_adversary_should_cause_no_timeouts_because_of_sync_retries[0] FAILED (9.1s)

  java.lang.AssertionError: 
  Expecting all elements of:
    {CONSENSUS_SAFETY=Optional.empty, CONSENSUS_NO_TIMEOUTS=Optional.empty, CONSENSUS_NO_EPOCH_TIMEOUTS=Optional.empty, CONSENSUS_VERTEX_REQUEST_RATE=Optional[TestInvariantError{desc=Get Vertices over the rate limit (75/sec) for node: NodeId[0279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798]
```

talekhinezh (approved):

jakrawcz-rdx:
Worth noting: the `[0]` there means that `Get Vertices over the rate limit (75/sec)` is not met for `numNodes=4` (while it works fine for `numNodes=10`).

(OFC locally I was not able to catch it; although I do see a lot of `[WARN/BFTSync/BFT no...tesf6qf6p] (BFTSync.java:465) - RATE_LIMIT: Outbound BFT Sync request GetVerticesRequest{id=c9097e6b1e3a3ff84905579e9c1356b94f005d13dca6fc9afc22f8a0f3575e8b count=1} for round 54 due to QCSync to NodeId[02e493dbf1c10d80f3581e4904930b1404cc6c13900ee0758474fa94abe8c4cd13] was not sent because we're over our 50.0/second rate limit on sync requests`)

Why would switching to static proposal-dropper set cause the too-high vertex request rate?

jakrawcz-rdx:
Ok, so first a noob question: the base simulation test config there introduces `F=(N-1)/3` adversaries (a max theoretically allowed number). And then, the `dropping_sync_adversary_should_cause_no_timeouts_because_of_sync_retries` test adds `NetworkDroppers.bftSyncMessagesDropped(0.1)` (i.e. 10% more "adversity", caused by dropping random BftSync messages). Why would we assume that hotstuff still progresses without any problems?

_(also interesting: why is this effect magnified by the base `F` becoming static...)_

jakrawcz-rdx:
ehh sweet, now yet another `given_incorrect_module_where_vertex_sync_is_disabled__then_test_should_fail_against_drop_proposal_adversary[1]` failed (the other passed this time).

Maybe we assert too hard in these tests?

jakrawcz-rdx:
huh, so even after getting rid of the "rate limit" assert, we have a single `dropping_sync_adversary_should_cause_no_timeouts_because_of_sync_retries[4]` failure (1 fail out of 5 runs for N=4).

This fail is the same as in https://github.com/radixdlt/babylon-node/pull/420#issuecomment-1514887221 (and my question "why would we expect this to be NOT flakey" remains).

jakrawcz-rdx:
the most recent commit has the most stable test set-up that I found; I am just not sure if it is correct 🤷 but ready for review.

jakrawcz-rdx:
Note: I re-run the int-tests on CI a few times (and each time it was [deliberately running 10x](https://github.com/radixdlt/babylon-node/pull/420/files#diff-df670db7ed3ddc957b33c756aca3b3639a6cc476bef21822584a220227317131R108) the problematic tests): it no longer fails (well, it timed out once, but maybe because of the sheer repeated test length).

jakrawcz-rdx:
As [agreed on Slack](https://rdxworks.slack.com/archives/C01M90Y2448/p1682006219100579?thread_ts=1681890434.408999&cid=C01M90Y2448), we can merge this version.

PR #424, opened 2023-04-21
Expanding the range of the Java-side locks (to cover the missed read-modify-write case)

jakrawcz-rdx:
A small fix for an issue introduced by https://github.com/radixdlt/babylon-node/pull/421.

Before this PR, the `FNodesNeverReceiveProposalDropperTest ` can be ~easily caught flakey locally (1/30 runs, see https://github.com/radixdlt/babylon-node/pull/423#issuecomment-1517571044).

After this PR, I can no longer repro locally (>100 runs in a loop). Let's see CI.

PR #429, opened 2023-04-24
Extending the lock span in Java-side prepare (avoid attempts to prepare against old state)

jakrawcz-rdx:
This is yet another fix after https://github.com/radixdlt/babylon-node/pull/421 (the first one was https://github.com/radixdlt/babylon-node/pull/424).

### DIFF
--- a/core/src/main/java/com/radixdlt/ledger/StateComputerLedger.java
+++ b/core/src/main/java/com/radixdlt/ledger/StateComputerLedger.java
@@ -207,92 +207,90 @@ public final class StateComputerLedger implements Ledger, ProposalGenerator {
     final LedgerHeader parentHeader = vertex.getParentHeader().getLedgerHeader();
     final AccumulatorState parentAccumulatorState = parentHeader.getAccumulatorState();
 
-    // The `commit()` operation may concurrently advance the ledger, so let's only read out the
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/radixdlt/ledger/StateComputerLedger.java
```diff
@@ -207,93 +207,91 @@ private Optional<ExecutedVertex> prepareInternal(
...
-    // The `commit()` operation may concurrently advance the ledger, so let's only read out the
-    // (immutable) accumulator state once.
-    // (we do not rely on any other state, and the Rust side of the prepare operation is properly
-    // synchronized there, which means we can carry on without the lock).
-    final AccumulatorState againstAccumulatorState;
+    final StateComputerResult result;
     synchronized (this.commitAndAdvanceLedgerLock) {
-      againstAccumulatorState = this.currentLedgerHeader.getHeader().getAccumulatorState();
-    }
+      final AccumulatorState againstAccumulatorState =
```
jakrawcz-rdx, line 212: _(a note from the author)_

In theory, the java prepare could stay with that short lock, and we could handle the "oops, old state" case in rust (i.e. return such result gracefully). But @malexandru-rdx convinced me that it is not worth it.

PR #434, opened 2023-04-25
Logging the mempool rejections as debug (instead of error) on java side

jakrawcz-rdx:
Going according to this: https://rdxworks.slack.com/archives/C01M90Y2448/p1682432746089539?thread_ts=1682088313.507689&cid=C01M90Y2448

**As a result of this PR**, these **3** rust-originating "problems"...
https://github.com/radixdlt/babylon-node/blob/cbded8dbb21a7063913afb45e3439903be136af0/core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs#L16-L20
...will now be logged as **debug, not error**.

Regarding the other, related reported log line:
`2023-04-21T15:08:13.858238Z ERROR core_api_server::core_api::handlers::transaction_submit: error=ResponseError { status_code: 400, public_error_message: "Mempool is full", trace: None, details: Some(TransactionSubmitMempoolFullErrorDetails { mempool_capacity: 50 }) }`

It ☝️ is not logged directly but our code, but comes from `tracing`, which by default logs the `Err` with a `level=Error` (reasonable TBH).
This behavior is configurable since _literally yesterday_: https://github.com/tokio-rs/tracing/releases/tag/tracing-attributes-0.1.24 - we can bump and then do something like:
image

**However**, I think it's not what we want - typically, I'd expect an HTTP 500 to be logged as error, and HTTP 4xx as info, etc. We cannot configure the `tracing` to be that flexible. Any ideas?

### DIFF
--- a/core/src/main/java/com/radixdlt/rev2/REv2StateComputer.java
+++ b/core/src/main/java/com/radixdlt/rev2/REv2StateComputer.java
@@ -164,7 +164,7 @@ public final class REv2StateComputer implements StateComputerLedger.StateCompute
                 mempoolAddSuccessEventDispatcher.dispatch(success);
               } catch (MempoolFullException | MempoolDuplicateException ignored) {
               } catch (MempoolRejectedException e) {
-                log.error(e);
+                log.debug(e);
               }
             });
   }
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/radixdlt/rev2/REv2StateComputer.java
```diff
@@ -164,7 +164,7 @@ public void addToMempool(MempoolAdd mempoolAdd, NodeId origin) {
                 mempoolAddSuccessEventDispatcher.dispatch(success);
               } catch (MempoolFullException | MempoolDuplicateException ignored) {
```
dhedey, line 165: I know this is unrelated to your change, but confused me a little, maybe a `// Ignored` in this empty bracket might make this clearer :)

dhedey (approved):

PR #435, opened 2023-04-26
Custom HttpStatus-aware replacement for `err(Debug)` in instrumentation

jakrawcz-rdx:
A follow-up to https://github.com/radixdlt/babylon-node/pull/434.

Here we take care of the `"Mempool is full"` log, which was an open question back then.

I removed all `err(Debug)`, and now its job is done by a single `emit_error_response_event` custom axum Layer.
Here is how it looks before (line 1) and after (line 2):
image
_(in short: the same info, but `path=` instead of handler function name)_

Currently I configured the levels as: `WARN` for HTTP 5xx, `DEBUG` for every other HTTP code (according to David's wish at https://rdxworks.slack.com/archives/C01M90Y2448/p1682440513064859?thread_ts=1682436148.845669&cid=C01M90Y2448).

### DIFF
--- a/core-rust/core-api-server/src/core_api/errors.rs
+++ b/core-rust/core-api-server/src/core_api/errors.rs
@@ -2,6 +2,7 @@ use axum::{
     response::{IntoResponse, Response},
     Json,
 };
+
 use hyper::StatusCode;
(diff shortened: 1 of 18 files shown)


malexandru-rdx (approved):
👍

PR #437, opened 2023-04-26
Implement list_substates fn

LukasGasior1:


core-rust/state-manager/src/store/db.rs
```diff
@@ -87,6 +87,32 @@ use crate::{
 use radix_engine::types::{KeyValueStoreId, SubstateId};
 use radix_engine_stores::hash_tree::tree_store::{NodeKey, Payload, ReadableTreeStore, TreeNode};
 
+// TODO(engine-merge): remove (use the impl in radix engine)
+pub fn encode_substate_id(index_id: &Vec<u8>, db_key: &Vec<u8>) -> Vec<u8> {
```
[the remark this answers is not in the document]
jakrawcz-rdx: in https://github.com/radixdlt/radixdlt-scrypto/pull/976 I kinda went with the flow and kept using the `DbIndex` + `DbKey` (and their tuple called `DbId`). I introduced the named typedefs and it suddenly became hard to confuse (I am sure we cannot operate on `Vec<u8>`s for everything).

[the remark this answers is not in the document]
dhedey: Agreed - see my comment on Jakub's PR on the scrypto side.

core-rust/state-manager/src/store/db.rs
```diff
@@ -87,6 +87,32 @@ use crate::{
...
+    }
+    None
+}
+
+// TODO(engine-merge): placeholder trait; remove once radix engine is merged
+pub trait ListableSubstateStore {
+    fn list_substates(
+        &self,
+        index_id: &Vec<u8>,
+    ) -> Box<dyn Iterator<Item = (Vec<u8>, Vec<u8>)> + '_>;
```
malexandru-rdx, line 113: Is `dyn` necessary here?

core-rust/state-manager/src/staging/cache.rs
```diff
@@ -287,7 +316,10 @@ impl<K, N> AccuTreeDiff<K, N> {
 
 #[derive(Clone)]
 pub struct ImmutableStore {
+    // TODO(engine-merge):: remove
```
dhedey, line 319: Can we just use one?

After merging latest scrypto, we can probably just have `ImmutableOrdMap<DbPartitionKey, DbSortKey>` :)

core-rust/state-manager/src/staging/sorted_kv_merge_iterator.rs
```diff
@@ -0,0 +1,116 @@
+use std::cmp::Ordering;
+use std::iter::Peekable;
+
+/// Merges two ordered (by K) iterators of (K, V) pairs.
+/// Preserves the order by interleaving items from both iterators.
+/// If the same key is present in both iterators,
+/// a pair from the left iterator is returned
+/// and a pair from the right iterator is discarded.
+pub struct SortedKvMergeIterator<K, V, L, R>
```
dhedey, line 9: Really nice abstraction :D .

I might slightly prefer them called "overlay" and "base" or something rather than "left" and "right" because l/r doesn't suggest any semantics to me.
Also because I'd naturally put base in left and overlay in right :P

core-rust/state-manager/src/store/db.rs
dhedey, line 91: It's not clear to me what this encoding is for - it looks like we will have different encodings for different things now. I'd suggest talking to @jakrawcz-rdx  to get some help with this - as we've discussed this extensively with Josh and have a clearer idea about what's best here 👍

core-rust/state-manager/src/store/db.rs
dhedey: Either you use Box<dyn Iterator... or you make the iterator an associated type of the trait. Happy with either.

core-rust/state-manager/src/staging/sorted_kv_merge_iterator.rs
```diff
@@ -0,0 +1,116 @@
+use std::cmp::Ordering;
+use std::iter::Peekable;
+
+/// Merges two ordered (by K) iterators of (K, V) pairs.
```
dhedey, line 4: Sadly this doesn't work right - we need to also handle _deletes_ - which means that the staging store (and this merger) will need to operate on DbOperations rather than values IMO.

In particular, eg, if overlay is delete and base has value, then both iterators advance and ignore the value.

dhedey (changes requested):
Nice abstractions, a few things need to be fixed or updated with the new scrypto version :)

core-rust/state-manager/src/store/db.rs
malexandru-rdx: I'm in favour of a GAT (compile-time vs runtime).

LukasGasior1:
superseded by https://github.com/radixdlt/babylon-node/pull/448

PR #442, opened 2023-04-28
rename typescript sdk subapis

dawidsowardx:
Change property names exposed through `CoreApiClient`. It's not general practice in Typescript to uppercase public class properties.


malexandru-rdx (approved):

dhedey (approved):
Nice, thanks!

If you have time, can we add your nice `innerClient` abstraction as well, as per Gateway API, and adjust the docs? :)

dawidsowardx:
> If you have time, can we add your nice `innerClient` abstraction as well, as per Gateway API, and adjust the docs? :)

What do you exactly mean? Is it creating empty, wrapper classes for each sub api? LTS and Status already have access to `innerClient`

PR #445, opened 2023-05-01
Replace database boilerplate code with enum_dispatch proc_macro

malexandru-rdx:
This PR is built on top of [PR 444](https://github.com/radixdlt/babylon-node/pull/444).

The `StateManagerDatabase` uses static dispatch (using enum/match) in order to support multiple implementations. Obviously this code is very repetitive and hence easy to generate. 

In this PR I use `enum_dispatch` to replace most of the boilerplate trait implementations for it.

Note 1: `proc_macro`s don't work across crates, so little to no chance for generating traits from `radixdlt-scrypto` this easily.
Note 2: Generic traits are not yet supported.

### DIFF
--- a/core-rust/Cargo.lock
+++ b/core-rust/Cargo.lock
@@ -583,6 +583,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7fcaabb2fef8c910e7f4c7ce9f67a1283a1715879a7c230ca9d6d1ae31f16d91"
 
 [[package]]
+name = "enum_dispatch"
+version = "0.3.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "11f36e95862220b211a6e2aa5eca09b4fa391b13cd52ceb8035a24bf65a79de2"
+dependencies = [
+ "once_cell",
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+]
+
+[[package]]
 name = "errno"
 version = "0.3.0"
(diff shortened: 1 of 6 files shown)


jakrawcz-rdx (approved):
I think this change is an improvement enough to be merged as-is 👍 

Regarding your 2 notes (on places where this could not be applied): could we workaround it e.g. by defining `trait OurSubstateStore: scrypto::SubstateStore` ?

malexandru-rdx (commented):
@jakrawcz-rdx 
> Regarding your 2 notes (on places where this could not be applied): could we workaround it e.g. by defining trait OurSubstateStore: scrypto::SubstateStore ?

`OurSubstateStore` is a [supertrait](https://doc.rust-lang.org/rust-by-example/trait/supertraits.html) and it does nothing to the implementation of the `scrypto::SubstateStore` (you still have to `impl` it somewhere). The way `enum_dispatch` works is you "register" the traits that you want to autogenerate (with `#[enum_dispatch]`), which captures everything to be implemented later (at `#[enum_dispatch(OurSubstateStore)]`). 
Unfortunately it doesn't walk the chain of trait dependencies and even if it did, `enum_dispatch` doesn't know anything about those traits to impl them.

There is also [enum_delegate](https://crates.io/crates/enum_delegate) which "fixes" this by letting you pass the trait definition into the proc. I tried using it but the published version is still buggy and with more corner cases than `enum_dispatch`.

dhedey (approved):
Looks good - let's have this for now, and then if/when we remove the in memory DB, we can possibly remove enum_dispatch. :)

PR #446, opened 2023-05-01
Removing no-longer-used metric definitions

jakrawcz-rdx:
I only did this because I noticed "remove old metrics" in the scope of https://radixdlt.atlassian.net/browse/NODE-545 (but apparently the mempool metrics were already cleaned up).


dhedey (approved):

jakrawcz-rdx:
(only some windows build got a random HTTP 500)

PR #447, opened 2023-05-03
Limit core api stream response size

malexandru-rdx:
In this PR I: 
1. add `ByteCountWriter` which implements `std::io::Write` in order to "hijack" `serde_json::to_writer` for size calculation instead of rendering `Json`s.
2. add `GetJsonSize` which uses `ByteCountWriter`
3. add a response total size limit for all `/*/stream/*` endpoints

### DIFF
--- a/core-rust/core-api-server/src/core_api/constants.rs
+++ b/core-rust/core-api-server/src/core_api/constants.rs
@@ -1,4 +1,5 @@
 pub(crate) const MAX_STREAM_COUNT_PER_REQUEST: u16 = 10000;
+pub(crate) const MAX_STREAM_TOTAL_SIZE_PER_RESPONSE: usize = 40 * 1024 * 1024;
 
 // TODO - Change this to be slightly larger than the double the max transaction payload size.
 // (We double due to the hex encoding of the payload)
--- a/core-rust/core-api-server/src/core_api/handlers/lts/stream_account_transaction_outcomes.rs
+++ b/core-rust/core-api-server/src/core_api/handlers/lts/stream_account_transaction_outcomes.rs
@@ -59,27 +59,35 @@ pub(crate) async fn handle_lts_stream_account_transaction_outcomes(
         limit,
     );
 
-    let committed_transaction_outcomes = state_versions
-        .iter()
-        .map(|state_version| {
-            Ok(to_api_lts_committed_transaction_outcome(
-                &mapping_context,
-                database
-                    .get_committed_transaction(*state_version)
-                    .expect("Transaction store corrupted"),
-                database
-                    .get_committed_transaction_receipt(*state_version)
-                    .expect("Transaction receipt index corrupted"),
-                database
-                    .get_committed_transaction_identifiers(*state_version)
-                    .expect("Transaction identifiers index corrupted"),
-            )?)
-        })
-        .collect::<Result<Vec<models::LtsCommittedTransactionOutcome>, ResponseError<()>>>()?;
+    // Reserve enough for the "header" fields
+    let mut current_total_size = 2048;
+    let mut committed_transaction_outcomes = Vec::new();
+    for state_version in state_versions {
+        let committed_transaction_outcome = to_api_lts_committed_transaction_outcome(
+            &mapping_context,
+            database
+                .get_committed_transaction(state_version)
+                .expect("Transaction store corrupted"),
+            database
+                .get_committed_transaction_receipt(state_version)
+                .expect("Transaction receipt index corrupted"),
+            database
+                .get_committed_transaction_identifiers(state_version)
+                .expect("Transaction identifiers index corrupted"),
+        )?;
+
+        let committed_transaction_size = committed_transaction_outcome.get_json_size();
+        if current_total_size + committed_transaction_size > MAX_STREAM_TOTAL_SIZE_PER_RESPONSE {
+            break;
+        }
+        current_total_size += committed_transaction_size;
+
+        committed_transaction_outcomes.push(committed_transaction_outcome);
+    }
 
     Ok(models::LtsStreamAccountTransactionOutcomesResponse {
         from_state_version: to_api_state_version(from_state_version)?,
-        count: state_versions
+        count: committed_transaction_outcomes
             .len()
             .try_into()
             .map_err(|_| server_error("Unexpected error mapping small usize to i32"))?,
--- a/core-rust/core-api-server/src/core_api/handlers/lts/stream_transaction_outcomes.rs
+++ b/core-rust/core-api-server/src/core_api/handlers/lts/stream_transaction_outcomes.rs
@@ -31,22 +31,30 @@ pub(crate) async fn handle_lts_stream_transaction_outcomes(
 
     let max_state_version = database.max_state_version();
 
-    let txns = database.get_committed_transaction_bundles(
+    let transactions = database.get_committed_transaction_bundles(
         from_state_version,
         limit.try_into().expect("limit out of usize bounds"),
     );
 
-    let committed_transaction_outcomes = txns
-        .into_iter()
-        .map(|(ledger_transaction, receipt, identifiers)| {
-            Ok(to_api_lts_committed_transaction_outcome(
-                &mapping_context,
-                ledger_transaction,
-                receipt,
-                identifiers,
-            )?)
-        })
-        .collect::<Result<Vec<models::LtsCommittedTransactionOutcome>, ResponseError<()>>>()?;
+    // Reserve enough for the "header" fields
+    let mut current_total_size = 2048;
+    let mut committed_transaction_outcomes = Vec::new();
+    for (ledger_transaction, receipt, identifiers) in transactions {
+        let committed_transaction = to_api_lts_committed_transaction_outcome(
+            &mapping_context,
+            ledger_transaction,
+            receipt,
+            identifiers,
+        )?;
+
+        let committed_transaction_size = committed_transaction.get_json_size();
+        if current_total_size + committed_transaction_size > MAX_STREAM_TOTAL_SIZE_PER_RESPONSE {
+            break;
+        }
+        current_total_size += committed_transaction_size;
+
+        committed_transaction_outcomes.push(committed_transaction);
+    }
 
     let count: i32 = {
         let transaction_count = committed_transaction_outcomes.len();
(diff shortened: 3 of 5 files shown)


core-rust/core-api-server/src/core_api/constants.rs
```diff
@@ -1,4 +1,5 @@
 pub(crate) const MAX_STREAM_COUNT_PER_REQUEST: u16 = 10000;
+pub(crate) const MAX_STREAM_TOTAL_SIZE_PER_RESPONSE: usize = 40 * 1024 * 1024;
```
malexandru-rdx, line 2: Note: I am not 100% on this limit. Can the (JSON) receipt be bigger than this? A bigger limit seems too loose, but obviously we want to be able to return at least 1 entry :)

core-rust/core-api-server/src/core_api/helpers.rs
```diff
@@ -62,3 +64,37 @@ pub(crate) fn wrong_substate_type(substate_offset: SubstateOffset) -> ResponseEr
...
+        Ok(())
+    }
+}
+
+pub trait GetJsonSize: Serialize {
+    fn get_json_size(&self) -> usize {
+        let mut bytes = 0;
+        {
+            let writer = ByteCountWriter::new(&mut bytes);
+            serde_json::to_writer(writer, &self).expect("Failed to serialize JSON");
```
KrzysztofLabus-RDX, line 94: Am I right to assume that we serialize to JSON just to get final string length? That seems super wasteful. I guess we should try serialize some sort of stream that could be backed by some array/collection of transactions that would additionally account for serialized response size as collection elements get serialized.

core-rust/core-api-server/src/core_api/handlers/stream_transactions.rs
```diff
@@ -47,22 +47,30 @@ pub(crate) async fn handle_stream_transactions(
...
-            Ok(to_api_committed_transaction(
-                &mapping_context,
-                ledger_transaction,
-                receipt,
-                identifiers,
-            )?)
-        })
-        .collect::<Result<Vec<models::CommittedTransaction>, ResponseError<()>>>()?;
+    // Reserve enough for the "header" fields
+    let mut current_total_size = 2048;
```
KrzysztofLabus-RDX, line 56: That looks super fragile... I'm quite sure it will break at some point.

core-rust/core-api-server/src/core_api/handlers/stream_transactions.rs
```diff
@@ -47,22 +47,30 @@ pub(crate) async fn handle_stream_transactions(
...
+        let committed_transaction = to_api_committed_transaction(
+            &mapping_context,
+            ledger_transaction,
+            receipt,
+            identifiers,
+        )?;
+
+        let committed_transaction_size = committed_transaction.get_json_size();
+        if current_total_size + committed_transaction_size > MAX_STREAM_TOTAL_SIZE_PER_RESPONSE {
+            break;
```
KrzysztofLabus-RDX, line 68: Guess we should at least log this information (produce some metric)? This feature seems extremely fragile so I'd invest a bit in observability here.

KrzysztofLabus-RDX (changes requested):

core-rust/core-api-server/src/core_api/handlers/stream_transactions.rs
malexandru-rdx: I agree. Let me try another good-enough approximation.

core-rust/core-api-server/src/core_api/helpers.rs
malexandru-rdx: Right. This is what I've already spent way too much time trying to do right and efficient. Unfortunately Rust's `serde` does not support any kind of higher level stream besides the "write stream" where I lose any context of what object I'm writing.

`serde` is just plainly not really performant or meant to be used for things like this (it's interface is `-> Result<String>`), normally when this is a problem, people start using protobufs and gRPC.

Strictly for performance (but really not that accurate) not even `get_size::GetSize` can be used because of `models::SborData` dependency on `serde_json::value` itself, which can't be decorated anymore.

On the bright side though, the json serializer doesn't need to allocate memory for Strings.

KrzysztofLabus-RDX: Quick question: maybe we shouldn't even bother counting bytes of JSON-serialized payload? Can we estimate TX size prior to serialization? Think:

```
val maxSize = 1 * 1024 * 1024;
var calculatedSize = 0;
val txCollection = ...;
val finalTxCollection = new List();

foreach (var tx in txCollection) {
  val txRoughSize = 1024 // some typical overhead 
    + tx.manifest.size() // user tx manifest size 
    + tx.createdSubstates.Sum(s => s.sbor_payload.size() * 1.5) // 1.5 is some constant factor
    + tx.updatedSubstates.Sum(s => s.sbor_payload.size() * 1.25) 
    + tx.events.Sum(e => e.sbor_data.size() * 1.75)

  calculatedSize += txRoughSize;
  finalTxcollection.append(tx);

  if (calculatedSize > maxSize) {
    _logger.warning("bla bla bla, max est. size reached....")

    break;
  }
}
```

Looks less fragile...

core-rust/core-api-server/src/core_api/handlers/stream_transactions.rs
malexandru-rdx: We don't have any real metrics in core-api right now.

Added `warn!`s.

core-rust/core-api-server/src/core_api/helpers.rs
malexandru-rdx: Something like this I tried with the `get_size::GetSize`, get the RAM size and multiply by some factors. But I am stopped by `SborData::data_json` which is already an `serde_json::value` and I can't really touch (in order to add `get_size::GetSize` trait).

For sake of robustness I can still multiply current size value by some factor (just to 100% make sure that the limit is indeed an upper bound).

core-rust/core-api-server/src/core_api/constants.rs
jakrawcz-rdx: I think that special-casing "return >=1" is very valid anyways 👍

dhedey: I agree with Jakub, and relatedly - I don't think this needs to be a hard-and-fast limit, and we should give a unit.

Perhaps a `STREAM_CAP_RESPONSE_WHEN_ABOVE_BYTES` or something :)

core-rust/core-api-server/src/core_api/handlers/lts/stream_account_transaction_outcomes.rs
```diff
@@ -59,32 +60,43 @@ pub(crate) async fn handle_lts_stream_account_transaction_outcomes(
...
+                .expect("Transaction receipt index corrupted"),
+            database
+                .get_committed_transaction_identifiers(state_version)
+                .expect("Transaction identifiers index corrupted"),
+        )?;
+
+        let committed_transaction_size = committed_transaction_outcome.get_json_size();
+        if current_total_size + committed_transaction_size > MAX_STREAM_TOTAL_SIZE_PER_RESPONSE {
+            let account_address = request.account_address;
+            warn!("Query for {account_address} from state version {from_state_version} with count limit of {limit} passed total size limit of {MAX_STREAM_TOTAL_SIZE_PER_RESPONSE}.");
```
dhedey, line 90: I don't think we should even log this. Maybe at debug. I imagine we'll hit this limit quite often - particularly if Gateway specifies a limit of 10000.

core-rust/core-api-server/src/core_api/handlers/lts/stream_account_transaction_outcomes.rs
```diff
@@ -59,32 +60,43 @@ pub(crate) async fn handle_lts_stream_account_transaction_outcomes(
...
+            database
+                .get_committed_transaction_receipt(state_version)
+                .expect("Transaction receipt index corrupted"),
+            database
+                .get_committed_transaction_identifiers(state_version)
+                .expect("Transaction identifiers index corrupted"),
+        )?;
+
+        let committed_transaction_size = committed_transaction_outcome.get_json_size();
+        if current_total_size + committed_transaction_size > MAX_STREAM_TOTAL_SIZE_PER_RESPONSE {
```
dhedey, line 88: To avoid issues where the limit is too small, I suggest we just stop processing if we're already above the limit. So it's more of a soft cap rather than a limit?

core-rust/core-api-server/src/core_api/handlers/lts/stream_transaction_outcomes.rs
```diff
@@ -31,25 +32,44 @@ pub(crate) async fn handle_lts_stream_transaction_outcomes(
...
+        let committed_transaction = to_api_lts_committed_transaction_outcome(
+            &mapping_context,
+            ledger_transaction,
+            receipt,
+            identifiers,
+        )?;
+
+        let committed_transaction_size = committed_transaction.get_json_size();
+        if current_total_size + committed_transaction_size > MAX_STREAM_TOTAL_SIZE_PER_RESPONSE {
+            warn!("Query from state version {from_state_version} with count limit of {limit} passed total size limit of {MAX_STREAM_TOTAL_SIZE_PER_RESPONSE}.");
```
dhedey, line 60: Ditto regarding log level and behaviour

core-rust/core-api-server/src/core_api/handlers/stream_transactions.rs
dhedey: Might be nice to have metrics - we could add CoreApiMetrics (registered with the state manager registry on Core API boot-up) to the `CoreApiState`.

But let's save this for a follow-up PR 👍

core-rust/core-api-server/src/core_api/helpers.rs
dhedey: Some good thoughts @KrzysztofLabus-RDX, but to keep-it-simple, I suggested this approach as a good first-stab.

In reality, the overhead of doing this is likely small in the grand scheme of things, and won't be on any critical paths if we're on an API thread here. It's also more accurate than a memory-measuring approach, particularly if we embed programmatic SBOR with user-provided names.

core-rust/core-api-server/src/core_api/handlers/stream_transactions.rs
dhedey: Given that we don't really have a hard-and-fast limit, more a loose cap to try and avoid massive responses / OOM in node and clients, I think we could also just make this an estimate and it's fine.

dhedey (approved):
Approving assuming the count is fixed in this PR, and the iterator / metrics are in follow up PRs :)

KrzysztofLabus-RDX (approved):

PR #449, opened 2023-05-04
DO-1241 / Develop / replace GH secrets with AWS secrets

kofrdx:
Prove that the secrets can be loaded:
https://github.com/radixdlt/babylon-node/actions/runs/4883966591/jobs/8716240040


kofrdx:
On this one I made a mistake. I pushed develop from local and skipped the rule for review.

PR #453, opened 2023-05-05
Cost saving gcr pull

kofrdx:


core-rust/build.gradle
```diff
@@ -142,8 +142,8 @@ task buildRustForDocker(type: Exec) {
                 --file Dockerfile \
                 --build-arg RUST_PROFILE=$envRustProfile \
                 --build-arg BUILDKIT_INLINE_CACHE=1 \
-                --cache-from type=registry,ref=$cacheRef \
-                --cache-to type=registry,ref=$cacheRef,mode=max \
+                --cache-from type=local,src=/tmp/outputs/cache/docker \
```
dhedey, line 145: Should the `$cacheRef` variable be removed in line 129/134 above?

Might also be worth a comment that we are using a cache restore plugin as it's cheaper than using the GCR registry?

dhedey (approved):

PR #455, opened 2023-05-08
DO-1389 / develop / add workflow to build keygen image

kofrdx:
Build the keygen image in PR's and on Workflow dispatch to not loose the ability to release/update it.


Dockerfile
```diff
@@ -68,6 +68,26 @@ USER root
 RUN SKIP_NATIVE_RUST_BUILD=TRUE gradle clean build -x test -Pci=true -PrustBinaryBuildType=release
 USER nobody
 
+# =================================================================================================
+# LAYER: Keygen
+# An alternative build target that executes the keygeneration
+# =================================================================================================
+FROM --platform=linux/amd64 eclipse-temurin:17-jre-alpine AS keygen
```
shambupujar, line 75: Why `eclipse-temurin:17-jre-alpine`?  is it because it has JRE already installed?

kofrdx: the one we used before had the jdk installed instead of jre. So we would have a build environment instead of a runtime environment. Switching to JRE saves 70MB in space.

PR #456, opened 2023-05-09
Database iterator

malexandru-rdx:
Continuing limits work ([PR 447](https://github.com/radixdlt/babylon-node/pull/447)), I change batch database queries into iterators in order to avoid high memory usage (and also wasted work).


core-rust/state-manager/src/store/traits.rs
```diff
@@ -174,14 +174,19 @@ pub mod transactions {
...
+        where
+            Self: 'a;
+
+        fn get_committed_transaction_bundle_iter(
             &self,
-            start_state_version_inclusive: u64,
-            limit: usize,
-        ) -> Vec<CommittedTransactionBundle>;
+            from_state_version: u64,
+        ) -> Self::CommittedTransactionBundleIterator<'_>;
```
dhedey, line 185: If we put the trait as `Box<dyn Iterator<Item = CommittedTransactionBundle> + '_>` here and dispense with the associated type/GAT then in a few cases (eg db.rs and with the account index trait) we could just return an iterator direct from eg RocksDB rather than create our own iterator struct. EG see `list_entries` in the `SubstateDatabase` trait.

The slightly lower performance would hardly matter in this case compared to the other 10000x larger overheads from the transaction mapping.

core-rust/state-manager/src/store/in_memory.rs
```diff
@@ -303,39 +303,68 @@ impl CommitStore for InMemoryStore {
...
+}
+
+impl IterableTransactionStore for InMemoryStore {
+    type CommittedTransactionBundleIterator<'a> = InMemoryCommittedTransactionBundleIterator<'a>;
+
+    fn get_committed_transaction_bundle_iter(
+        &self,
+        from_state_version: u64,
+    ) -> Self::CommittedTransactionBundleIterator<'_> {
+        debug_assert!(self.is_local_transaction_execution_index_enabled());
```
dhedey, line 361: I assume this assert is a fallback, and we've already implemented returning a HTTP error explaining that they need to enable the endpoint? (rather than crashing if a user attempts to call transaction stream with the local index disabled).

dhedey (commented):
Nice work! Will let you merge in `develop` and fix up any markups then happy to re-review :)

dhedey (commented):
Nice work! Will let you merge in `develop` and fix up any markups then happy to re-review :)

core-rust/state-manager/src/store/in_memory.rs
malexandru-rdx: Right. Current APIs are guarded with an client error. It would crash without the `debug_assert!` anyway (for RocksDB version) since it's panicking already when `LocalTransactionExecution` is missing.

core-rust/state-manager/src/store/traits.rs
malexandru-rdx: Sure. Let's go with `dyn`.

core-rust/state-manager/src/store/in_memory.rs
malexandru-rdx: This basically just makes it more clear why that crash happened.

core-rust/state-manager/src/store/in_memory.rs
```diff
@@ -305,39 +304,69 @@ impl CommitStore for InMemoryStore {
...
+}
+
+impl IterableTransactionStore for InMemoryStore {
+    fn get_committed_transaction_bundle_iter(
+        &self,
+        from_state_version: u64,
+    ) -> Box<dyn Iterator<Item = CommittedTransactionBundle> + '_> {
+        debug_assert!(self.is_local_transaction_execution_index_enabled());
 
+        Box::new(InMemoryCommittedTransactionBundleIterator::new(
```
dhedey, line 362: If you want, I think you can just inline it all now
```
Box::new(
    num::iter::RangeFrom(from_state_version)
       .filter_map(... code from next in InMemoryCommittedTransactionBundleIterator)
)
```

But also happy for it to be left as is if you'd rather 👍

core-rust/state-manager/src/store/in_memory.rs
```diff
@@ -471,20 +500,44 @@ impl AccountChangeIndexExtension for InMemoryStore {
...
-            None => Vec::new(),
-            Some(state_versions) => state_versions
-                .range((Included(start_state_version_inclusive), Unbounded))
-                .take(limit)
-                .cloned()
-                .collect(),
-        }
+        from_state_version: u64,
+    ) -> Box<dyn Iterator<Item = u64> + '_> {
+        Box::new(InMemoryAccountChangeIndexIterator::new(
```
dhedey, line 537: We can inline this as eg
```
    let Some(index) = account_change_index_set.get(&account) else {
        return Box::new(vec![].into_iter());
    }
    return Box::new(index.range((Included(from_state_version), Unbounded)));

PR #458, opened 2023-05-15
feat: add wrappers for each subapi in TS SDK

dawidsowardx:
## Description

Each Sub-API from now on has high-level wrapper (some of them are empty) which includes `innerClient` property.


dhedey (approved):
Thanks @dawidsowardx

PR #459, opened 2023-05-15
fix: LTS account balance endpoint on RCnet

dhedey:
Test on branch: `test/rcnet-lts-account-balance-test` which I'm purposefully not merging because a better test is already on develop (and this bug is not an issue on develop)


LukasGasior1:
LGTM!

PR #464, opened 2023-05-18
TS SDK update

malexandru-rdx:
- Cleanup `CoreApiClient.initialize` interface
- Remove unused fields in `runtime.mustache` and add `agent` directly
- Update README


sdk/typescript/lib/index.ts
```diff
@@ -49,6 +53,21 @@ export class CoreApiClient {
   private static constructConfiguration(
     settings: CoreApiClientSettings
   ): Configuration {
+    // Left for backward compatibility
+    if (settings.httpAgent || settings.httpsAgent) {
+      const agentSelector = (parsedUrl: any) => {
+        console.log(parsedUrl);
```
dhedey, line 59: I think this log line should be removed?

dhedey (approved):
Please merge once the log line is fixed, and then let me know so I can create a new build for NPM 👍, and merge into develop.

PR #467, opened 2023-05-22
Passing actual round leaders' reliability stats to the Engine on round change

jakrawcz-rdx:
Here is the "passed by Node" part of https://radixdlt.atlassian.net/browse/SCRY-435 :)


jakrawcz-rdx:
@dhedey as per our Slack discussion: I have updated the CoreApi to also return this new "proposal history" detail in the validator transaction response, using the indices of validators ordered by their address (which is now the standard ordering of validator set, used throughout Engine, Node and CoreApi).

core/src/main/java/com/radixdlt/rev2/REv2StateComputer.java
```diff
@@ -239,13 +247,22 @@ public StateComputerLedger.StateComputerResult prepare(
                             .toList(),
                         v.getLedgerHeader().getAccumulatorState().getAccumulatorHash()))
             .toList();
+    var gapRoundLeaderKeys =
+        LongStream.range(roundDetails.previousQcRoundNumber() + 1, roundDetails.roundNumber())
+            .mapToObj(Round::of)
+            .map(this.currentProposerElection.get()::getProposer)
+            .map(BFTValidatorId::getKey)
```
dhedey, line 254: So theoretically it's possible for two validators to use the same key. Therefore we should actually use the validator component address here.

core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/PrepareRequest.java
```diff
@@ -75,8 +76,11 @@ public record PrepareRequest(
     HashCode parentAccumulatorHash,
     List<PreviousVertex> previousVertices,
     List<RawNotarizedTransaction> proposed,
+    boolean isFallback,
     UInt64 epoch,
     UInt64 roundNumber,
+    List<ECDSASecp256k1PublicKey> gapRoundLeaderKeys,
```
dhedey, line 82: As per the other comment - we should be using the ComponentAddress rather than the public keys here. (because theoretically two validators in the set could share the same key).

dhedey (changes requested):
Marking as request changes to just have a change to using addresses rather than keys.

I also wonder whether it would be useful to have some kind of end-to-end test of this? Maybe a deterministic test where we drop consensus messages from a particular validator for a single round, and check that validator misses that round in the end of epoch summary event?

jakrawcz-rdx:
> I also wonder whether it would be useful to have some kind of end-to-end test of this?

Maybe on Monday...

dhedey (approved):
Looks good! Thanks :)

The E2E test will have to wait till after we merge birch anyway.

In the time being, will merge this into my `feature/merge-birch` branch

jakrawcz-rdx:
Let me merge this - as agreed on Slack, an e2e test will come in a separate PR after we merge birch.

PR #474, opened 2023-05-31
Dropping the accumulator hash computation/verification from Java side

jakrawcz-rdx:
⚠️ I actually do NOT want to merge this before https://github.com/radixdlt/babylon-node/pull/473 ⚠️ 

After this PR, the Java side will no longer compute accumulator hash nor verify it. Only rusts handles this, which should allow us to move from accumulator hash to transaction root hash in next PR.


jakrawcz-rdx:
Closing in favor of https://github.com/radixdlt/babylon-node/pull/477

PR #478, opened 2023-06-03
Removing all mentions of accumulator hash in favor of transaction root

jakrawcz-rdx:
After this PR, only the LTS Core API contains some retro-fitted accumulator hash.


core-rust/state-manager/src/staging/result.rs
```diff
@@ -264,6 +245,35 @@ impl ProcessedCommitResult {
...
+        system_updates: &SystemUpdates,
+    ) -> Vec<SubstateChange> {
+        let mut substate_changes = Vec::new();
+        for ((node_id, module_id), node_module_updates) in system_updates {
+            for (substate_key, update) in node_module_updates {
+                let partition_key = D::to_db_partition_key(node_id, *module_id);
+                let sort_key = D::to_db_sort_key(substate_key);
+                let change_action = match update {
+                    DatabaseUpdate::Set(value) => {
+                        match store.get_substate(&partition_key, &sort_key) {
```
jakrawcz-rdx, line 259: _(note from author)_

This code is not authored, just moved (out of the `process()` function).
While moving, I noticed that:
- we do this post-tx DB lookups, only because we need to distinguish between an update and create - if the Engine does not distinguish this, maybe we also don't need to?
- we use `system_updates` and have to do `D::to_db_..._key()` - could we use `database_updates` which are already provided by the Engine instead?

core-rust/state-manager/src/query/transaction_identifiers.rs
```diff
@@ -1,13 +1,15 @@
 use enum_dispatch::enum_dispatch;
 
-use crate::{AccumulatorState, CommittedTransactionIdentifiers};
+use crate::{CommittedTransactionIdentifiers, LedgerHashes};
 
 #[enum_dispatch]
 pub trait TransactionIdentifierLoader {
-    fn get_top_transaction_identifiers(&self) -> Option<CommittedTransactionIdentifiers>;
-    fn get_top_accumulator_state(&self) -> AccumulatorState {
+    // TODO(no-accu-hash): it's worth having `pub struct StateVersion(u64);` at this point
```
jakrawcz-rdx, line 7: _(note from author)_

I want to do this in a post-feature-refactor pretty much right after we merge this to develop.

core-rust/state-manager/src/transaction/preview.rs
```diff
@@ -68,24 +62,21 @@ impl<S: ReadableStore + QueryableProofStore + TransactionIdentifierLoader> Trans
             .execution_configurator
             .wrap(validated.get_executable(), ConfigType::Preview)
             .warn_after(PREVIEW_RUNTIME_WARN_THRESHOLD, "preview");
-        let receipt = transaction_logic.execute_on(read_store.deref());
 
-        // Fake a LedgerPayloadHash for the purposes of mapping the receipt as it doesn't matter for preview
-        // TODO - don't do most of this work for preview
```
jakrawcz-rdx, line 74: _(note from author)_

I decided to address this now, since otherwise I would have to fake even more stuff.

core/src/main/java/com/radixdlt/consensus/ConsensusHasher.java
```diff
@@ -83,8 +83,7 @@ public static HashCode toHash(
       outputStream.writeInt(header != null ? 0 : 1); // 4 bytes (Version)
       outputStream.write(opaque.asBytes()); // 32 bytes
       if (header != null) {
-        outputStream.write(header.getAccumulatorState().getAccumulatorHash().asBytes()); // 32 bytes
-        outputStream.writeLong(header.getAccumulatorState().getStateVersion()); // 8 bytes
+        outputStream.writeLong(header.getStateVersion()); // 8 bytes
         // TODO: include the header.getStateHash().asBytes() when we are ready to affect consensus
```
jakrawcz-rdx, line 87: _(note from author)_

@dhedey I think we missed this TODO, and...
1. Right now we could probably include all the `LedgerHashes` here, right?
2. AFAIU, this would effectively render the checks inside Rust `prepare()` useless (i.e. impossible to ever trigger on prod), should we keep them as sanity checks anyways?

core/src/main/java/com/radixdlt/sync/LocalSyncService.java
```diff
@@ -568,10 +548,7 @@ private <T extends SyncState> T updateSyncTargetDiffCounter(T syncState) {
       this.metrics
           .sync()
           .targetStateVersion()
-          .set(
-              Math.max(
-                  syncingState.getTargetHeader().getStateVersion(),
-                  syncingState.getTargetHeader().getAccumulatorState().getStateVersion()));
```
jakrawcz-rdx, line 574: _(note from author)_

Does anyone know what the `.max(X, X)` here was really doing? Should it be doing something else?

core-rust/state-manager/src/state_manager.rs
```diff
@@ -447,11 +439,12 @@ where
             let executable = prepared.get_executable();
 
             let next_epoch = {
```
jakrawcz-rdx, line 441: _(note from author)_

In a follow-up refactor I plan to pull the `next_epoch` tracking into `StateTracker`, it clicks nicely (see https://github.com/radixdlt/babylon-node/pull/474/files#diff-5230880996b770cf628e475ce1143be291e99fd9801f307fd1702a525c351aadR533).

core-rust/state-manager/src/state_manager.rs
```diff
@@ -279,17 +271,17 @@ where
...
         let epoch_identifiers = EpochTransactionIdentifiers::from(&epoch_header);
 
-        if prepare_request.committed_accumulator_state != base_accumulator_state {
+        if prepare_request.committed_ledger_hashes != base_ledger_hashes {
             panic!(
                 "state {:?} from request does not match the current ledger state {:?}",
-                prepare_request.committed_accumulator_state, base_accumulator_state
+                prepare_request.committed_ledger_hashes, base_ledger_hashes
             );
         }
```
jakrawcz-rdx, line 287: _(note from author)_

The pre-computation of `prepared_proposed_results` will go away in a refactor which will introduce `DuplicateIntentHashDetector` (see https://github.com/radixdlt/babylon-node/pull/474/files#diff-5230880996b770cf628e475ce1143be291e99fd9801f307fd1702a525c351aadR1005).

core-rust/state-manager/src/staging/cache.rs
```diff
@@ -93,20 +90,45 @@ use radix_engine_stores::hash_tree::tree_store::{
...
+}
+
 pub struct ExecutionCache {
     stage_tree: StageTree<ProcessedTransactionReceipt, ImmutableStore>,
-    root_accumulator_hash: AccumulatorHash,
-    accumulator_hash_to_key: HashMap<AccumulatorHash, DerivedStageKey>,
-    key_to_accumulator_hash: SecondaryMap<DerivedStageKey, AccumulatorHash>,
+    base_transaction_root: TransactionTreeHash,
+    transaction_placement_to_key: HashMap<TransactionPlacement, DerivedStageKey>,
+    transaction_root_to_key: HashMap<TransactionTreeHash, DerivedStageKey>,
```
jakrawcz-rdx, line 120: @malexandru-rdx This bit may be interesting to you: _(how I got rid of accu hash in the staging tree, while avoiding calculation of a transaction merkle tree update before cache-hit)_

A transaction may be unambiguously identified by 2 different business keys here:

- by a "transaction placement":
  - A "transaction placement" of transaction X is a tuple `{X's parent's transaction root, X's payload hash}`. It can be produced for any "candidate" transaction (i.e. even for ones that would be rejected).

- by a transaction root:
   - Yup, just a regular transaction tree root (i.e. our replacement of accumulator hash). There is a gotcha though: transaction root comes from transaction's `LedgerHashes`, and we do not compute `LedgerHashes` for transactions that are rejected (technically we could produce just the transaction root alone for them, but in our current impl we don't). Yet, we want to reference rejected transactions in our cache too!
   - The transaction root is perfect for referencing every possible `{parent of a candidate transaction}` (because a parent must have been committable). It is also easy to get it from the transaction placement (i.e. `placement.parent_tx_root`).

Hence, all we need is 2 `HashMap`s here instead of one :)

core/src/main/java/com/radixdlt/consensus/ConsensusHasher.java
jakrawcz-rdx: (let's discuss it in a spin-off https://github.com/radixdlt/babylon-node/pull/479)

core-rust/state-manager/src/state_manager.rs
jakrawcz-rdx: ☝️ this happens at https://github.com/radixdlt/babylon-node/pull/480

core-rust/state-manager/src/staging/result.rs
dhedey: > we use system_updates and have to do D::to_db_..._key() - could we use database_updates which are already provided by the Engine instead?

Yes - although in a way, I'm not sure why the engine outputs `database_updates` to start with - I feel like we could almost have an `into` from system updates to database updates?

core-rust/state-manager/src/staging/result.rs
```diff
@@ -264,6 +245,35 @@ impl ProcessedCommitResult {
             .map(|next_epoch_result| NextEpoch::from(next_epoch_result.clone()))
     }
 
+    pub fn compute_substate_changes<S: SubstateDatabase, D: DatabaseKeyMapper>(
+        store: &S,
+        system_updates: &SystemUpdates,
```
dhedey, line 250: Ideally this could take an owned SystemUpdates so we can avoid unnecessary cloning?

core-rust/state-manager/src/staging/result.rs
dhedey: > we do this post-tx DB lookups, only because we need to distinguish between an update and create - if the Engine does not distinguish this, maybe we also don't need to?

Ideally we'd like to distinguish them for the Core API, and we'll need to do further work here soon - eg parent map stuff.

dhedey (approved):
Nice stuff. Again, feel free to resolve markups (where they exist) in follow up PRs at the top of the commit stack.

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -290,19 +291,19 @@ impl RocksDBStore {
                 state_version.to_be_bytes(),
             );
         } else {
-            let maybe_existing_ledger_payload_hash = self
+            let maybe_existing_state_version = self
```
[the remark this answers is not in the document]
jakrawcz-rdx: we do have `StateVersionByTxnLedgerPayloadHash` and I've now replaced it on line 297 (indeed a bug, good eye!) 👍

core-rust/state-manager/src/jni/transaction_store.rs
```diff
@@ -204,10 +203,10 @@ extern "system" fn Java_com_radixdlt_transaction_REv2TransactionAndProofStore_ge
     j_state_manager: JObject,
     request_payload: jbyteArray,
 ) -> jbyteArray {
-    jni_sbor_coded_call(&env, request_payload, |_: ()| -> Option<JavaLedgerProof> {
+    jni_sbor_coded_call(&env, request_payload, |_: ()| -> Option<LedgerProof> {
         let database = JNIStateManager::get_database(&env, j_state_manager);
-        let epoch_proof = database.read().get_first_epoch_proof();
-        epoch_proof.map(JavaLedgerProof::from)
+        let proof = database.read().get_first_epoch_proof();
```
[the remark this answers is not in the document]
jakrawcz-rdx: tell it to the borrow checker and its lovely ascii art ;p
```
error[E0597]: `database` does not live long enough
   --> state-manager/src/jni/transaction_store.rs:208:9
    |
208 |         database.read().get_first_epoch_proof()
    |         ^^^^^^^^^^^^^^^
    |         |
    |         borrowed value does not live long enough
    |         a temporary with access to the borrow is created here ...
209 |     })
    |     -
    |     |
    |     `database` dropped here while still borrowed
    |     ... and the borrow might be used here, when that temporary is dropped and runs the `Drop` code for type `parking_lot::lock_api::RwLockReadGuard`
    |
    = note: the temporary is part of an expression at the end of a block;
            consider forcing this temporary to be dropped sooner, before the block's local variables are dropped
help: for example, you could save the expression's value in a new local variable `x` and then make `x` be the expression at the end of the block
    |
208 |         let x = database.read().get_first_epoch_proof(); x
    |         +++++++                                        +++
```

core-rust/state-manager/src/store/in_memory.rs
```diff
@@ -290,21 +272,23 @@ impl ReadableAccuTreeStore<u64, ReceiptTreeHash> for InMemoryStore {
...
-        for CommittedTransactionBundle {
-            raw,
-            receipt,
-            identifiers,
-        } in commit_bundle.transactions
-        {
-            self.insert_transaction(raw, receipt, identifiers);
+        let commit_ledger_header = &commit_bundle.proof.ledger_header;
+        let commit_state_version = commit_ledger_header.state_version;
+        let base_state_version = commit_state_version - commit_bundle.transactions.len() as u64 + 1;
```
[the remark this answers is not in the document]
jakrawcz-rdx: yeah I did it in the name of "why transfer a thing which is so easily computable".
But when I saw the duplication of the "easy computation" I thought "if anyone else is annoyed by this, I'll store the version in the bundle". So, doing it now.

(regarding "store it in `CommittedTransactionIdentifiers`" - the state version is a kinda primary key of a transaction in our DB, and I rarely put a PK into a value stored under this PK)

core-rust/state-manager/src/staging/result.rs
jakrawcz-rdx: unfortunately, with the current shape of the callers, we wouldn't gain much - the entire `CommitResult` is needed to create a `LocalTransactionReceipt` (after `compute_substate_changes()`), so we wouldn't be able to move out the `system_updates` anyways.

core-rust/state-manager/src/store/in_memory.rs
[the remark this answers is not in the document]
dhedey: Yeah, I agree with the call to remove it from `CommittedTransactionIdentifiers` given that it's the PK :)

dhedey (approved):

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -3779,7 +3779,7 @@ components:
           maxLength: 64
           description: |
             The hex-encoded transaction accumulator hash. This hash captures the order of all transactions on ledger.
-            This hash is `ACC_{N+1} = Blake2b-256(CONCAT(ACC_N, LEDGER_HASH_{N}))`, starting with `ACC_0 = 000..000` the pre-genesis accumulator.
+            This hash is `ACC_{N+1} = combine(ACC_N, LEDGER_HASH_{N}))` (where `combine()` is an arbitrary deterministic function we use).
```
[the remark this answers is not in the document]
malexandru-rdx: Me too, but I wouldn't leave it like this in the docs. Maybe mention it is an MerkleTree root hash.

core-rust/state-manager/src/staging/cache.rs
malexandru-rdx: Nice. Thank you!

malexandru-rdx (approved):
LGTM!

PR #480, opened 2023-06-05
Refactor: moving intent hash deduplication responsibility out of core prepare() impl

jakrawcz-rdx:
This refactoring mostly moves the code around, with just one significant logic change (not affecting externally observable behavior):
Before, the proposed transactions were looped over as the first step, in order to collect their intent hashes.
After this PR, the intent hash deduplication happens simply as part of the per-transaction processing.


core-rust/state-manager/src/state_manager.rs
```diff
@@ -287,73 +287,28 @@ where
...
         };
 
         let mut state_tracker = StateTracker::initial(base_state_version, base_ledger_hashes);
+        let mut duplicate_intent_hash_detector =
+            DuplicateIntentHashDetector::new(read_store.deref());
 
         for raw_ancestor in prepare_request.prepared_uncommitted_transactions {
-            // TODO - By passing through the accumulator / tree hash, avoid doing all this validation if the
-            // transactions are already in the execution cache!
+            // TODO(optimization-only): We could avoid the validations here (mostly: creation of
```
dhedey, line 303: The main savings I imagine are hashing, decoding and signature verification, rather than creating the executable itself. But your point stands :)

core-rust/state-manager/src/state_manager.rs
```diff
@@ -477,25 +432,44 @@ where
...
+                    TransactionValidationError::PrepareError(PrepareError::DecodeError(err))
+                })
+                .and_then(|ledger_transaction| {
+                    ledger_transaction.to_raw().map_err(|err| {
+                        TransactionValidationError::PrepareError(PrepareError::EncodeError(err))
+                    })
+                })
+                .and_then(|raw_ledger_transaction| {
+                    self.ledger_transaction_validator
+                        .prepare_from_raw(&raw_ledger_transaction)
```
dhedey, line 464: You can combine these two `and_then`s by using `.prepare_from_model`

core-rust/state-manager/src/state_manager.rs
```diff
@@ -1115,3 +1090,55 @@ struct PendingTransactionResult {
     pub invalid_at_epoch: Epoch,
     pub rejection_reason: Option<RejectionReason>,
 }
+
+/// An internal implementation delegate dealing with intent hash duplicates.
+// TODO: Remove after this responsibility is implemented by the Engine.
+struct DuplicateIntentHashDetector<'s, S> {
```
dhedey, line 1096: Nice :)

core-rust/state-manager/src/state_manager.rs
```diff
@@ -1115,3 +1090,55 @@ struct PendingTransactionResult {
...
+    store: &'s S,
+    recorded_intent_hashes: NonIterMap<IntentHash, IntentHashDuplicateWith>,
```
dhedey, line 1098: Super minor, but we could move `IntentHashDuplicateWith` down here to be alongside the definition of `DuplicateIntentHashDetector`

dhedey (approved):
Nice simplification / abstraction 👍

core-rust/state-manager/src/state_manager.rs
jakrawcz-rdx: I cannot :(
The `raw_ledger_transaction` (i.e. the mid-step result) needs to be returned, since it is later used (as `CommittableTransaction.raw`).

PR #481, opened 2023-06-05
fix: Round update transaction hashes

dhedey:
Ensures that two different round updates have different hashes.

### DIFF
--- a/core-rust/state-manager/src/transaction/round_update_transaction.rs
+++ b/core-rust/state-manager/src/transaction/round_update_transaction.rs
@@ -29,25 +29,27 @@ impl RoundUpdateTransactionV1 {
     }
 
     pub fn prepare(&self) -> Result<PreparedRoundUpdateTransactionV1, PrepareError> {
+        let prepared_instructions = InstructionsV1(self.create_instructions()).prepare_partial()?;
         let hash = HashAccumulator::new()
             .update([
                 TRANSACTION_HASHABLE_PAYLOAD_PREFIX,
                 TransactionDiscriminator::V1RoundUpdate as u8,
             ])
+            // Ensure we include the epoch and round in the hash, even if they're not included in the instructions
             .update(format!(
                 "RoundChange({},{})",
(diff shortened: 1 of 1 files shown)


core-rust/state-manager/src/transaction/round_update_transaction.rs
```diff
@@ -29,25 +29,27 @@ impl RoundUpdateTransactionV1 {
...
                 TRANSACTION_HASHABLE_PAYLOAD_PREFIX,
                 TransactionDiscriminator::V1RoundUpdate as u8,
             ])
+            // Ensure we include the epoch and round in the hash, even if they're not included in the instructions
             .update(format!(
                 "RoundChange({},{})",
                 self.epoch.number(),
                 self.round.number()
             ))
+            .update(prepared_instructions.summary.hash)
```
LukasGasior1, line 44: I'm a bit surprised we're hashing the instructions(?) here rather than the transaction fields

jakrawcz-rdx: hmm true, maybe the most reliable way would be to de-struct here and insert these fields indeed...?
(after all, these fields are the source of truth, while the instructions are derived)

(if de-struct and manual hasher updates are too cumbersome, then actually going through SBOR-encode() [as in the old days] doesn't seem that bad)

dhedey: True - I considered both...
The benefit of hashing the instructions is that the exact instructions can be proven via the merkle hash.
How about we include both a hash of the body and the instructions?

dhedey: Okay, latest push has done that. Will merge when tests pass

LukasGasior1: :+1:

LukasGasior1 (approved):

jakrawcz-rdx (approved):

PR #484, opened 2023-06-06
Feature: Rotate every validator as leader at the beginning of each epoch

jakrawcz-rdx:
This implements https://whimsical.com/babylon-work-tracker-Am2s7hSa7xHd5yCxgub67a@2bsEvpTYSt1HivsbzwK4vKRmWp4h8Rg5Jre

### DIFF
--- a/core/src/main/java/com/radixdlt/consensus/liveness/RotateOnceDecorator.java
+++ b/core/src/main/java/com/radixdlt/consensus/liveness/RotateOnceDecorator.java
@@ -62,43 +62,70 @@
  * permissions under this License.
  */
 
-package com.radixdlt.rev2.modules;
+package com.radixdlt.consensus.liveness;
 
-import com.google.inject.AbstractModule;
-import com.google.inject.Provides;
-import com.google.inject.Singleton;
-import com.radixdlt.consensus.BFTConfiguration;
-import com.radixdlt.consensus.LedgerProof;
-import com.radixdlt.consensus.bft.*;
-import com.radixdlt.consensus.liveness.ProposerElections;
-import com.radixdlt.consensus.vertexstore.VertexStoreState;
-import com.radixdlt.rev2.LastEpochProof;
+import com.google.common.collect.Ordering;
+import com.google.common.primitives.UnsignedBytes;
+import com.radixdlt.consensus.bft.BFTValidator;
+import com.radixdlt.consensus.bft.BFTValidatorId;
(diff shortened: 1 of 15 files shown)


core/src/test-core/java/com/radixdlt/modules/StateComputerConfig.java
```diff
@@ -154,14 +175,24 @@ record MockedStateComputerConfigWithEpochs(
...
     @Override
     public MockedMempoolConfig mempoolConfig() {
       return mempoolType;
     }
   }
 
+  /** A choice of {@link com.radixdlt.consensus.liveness.ProposerElection} implementation. */
+  enum ProposerElectionMode {
+    /** Use the {@link com.radixdlt.consensus.liveness.WeightedRotatingLeaders} only. */
+    ONLY_WEIGHTED_BY_STAKE,
```
dhedey, line 190: Do we need to keep / expose this?

core/src/main/java/com/radixdlt/consensus/liveness/RotateOnceDecorator.java
```diff
@@ -0,0 +1,131 @@
...
+import com.radixdlt.consensus.bft.Round;
+import com.radixdlt.rev2.ComponentAddress;
+import com.radixdlt.utils.KeyComparator;
+import java.util.*;
+
+/**
+ * A {@link ProposerElection} decorator which first simply iterates over all validators once (in the
+ * "highest stake first" order), and then lets the underlying implementation take over.
+ */
+public final class RotateOnceDecorator implements ProposerElection {
```
dhedey, line 81: This approach is only asymptotically "fair", compared with fixing the first n rounds of WeightedRotatingLeader rotation (and letting the leaders with more stake build up "power" over that time) - or directly using the method of equal proportions.

But given that we're separating proposal count from emissions, this is probably fine.

dhedey (approved):

core/src/main/java/com/radixdlt/consensus/liveness/RotateOnceDecorator.java
jakrawcz-rdx: I agree this is probably less fair than MoEP, but it is definitely a more gradual step (i.e. it affects only `validatorsSet.size()` first rounds of an epoch, so even if a massive bug is there, it does not destroy everything), which I value.

Let me merge this now, and we will still have opportunities to tune it later

PR #486, opened 2023-06-07
Refactor: Pulling the next epoch handling into state tracker

jakrawcz-rdx:
A very small refactoring promised by https://github.com/radixdlt/babylon-node/pull/478#discussion_r1215463555


dhedey (approved):
Nice! It's looking better and better

PR #491, opened 2023-06-08
Refactor: Introducing a type-safe `StateVersion`

jakrawcz-rdx:
This introduces type-safety and checked math operations.

I also propagated the usage of Engine's `Epoch` a bit into our code, as a drive-by.


malexandru-rdx (approved):

jakrawcz-rdx:
thanks for reviewing!
let me merge this (there are no behavior changes here + work piles on top of it)

PR #494, opened 2023-06-09
Refactor: Unifying the "series executor" usage + Moving it to separate module

jakrawcz-rdx:
I also move out some other internal helpers of state manager, so that we can keep that file <1000 lines :p


dhedey (approved):
Really nice! 😍 👍

PR #497, opened 2023-06-10
Improving error handling on commit

jakrawcz-rdx:
I hope I captured the conclusions (on the intended behavior) we had with David on Slack.

The drive-by commit addresses https://github.com/radixdlt/babylon-node/pull/494#discussion_r1224545645

### DIFF
--- a/core-rust/state-manager/src/staging/epoch_handling.rs
+++ b/core-rust/state-manager/src/staging/epoch_handling.rs
@@ -62,47 +62,135 @@
  * permissions under this License.
  */
 
+use crate::accumulator_tree::slice_merger::AccuTreeSliceMerger;
+use crate::accumulator_tree::storage::{
+    AccuTreeStore, ReadableAccuTreeStore, TreeSlice, WriteableAccuTreeStore,
+};
+use crate::accumulator_tree::tree_builder::AccuTree;
+use crate::accumulator_tree::IsMerklizableHash;
 use crate::StateVersion;
 
-/// An extracted logic related to the "accu tree per epoch" approach (where the first leaf of the
-/// next epoch's tree is an auto-inserted root of the previous epoch's tree).
-pub struct AccuTreeEpochHandler {
+/// A factory of accu tree utilities operating under the "accu tree per epoch" approach (where the
+/// first leaf of the next epoch's tree is an auto-inserted root of the previous epoch's tree).
+pub struct EpochAwareAccuTreeFactory {
+    epoch_state_version: StateVersion,
     epoch_version_count: usize,
 }
 
-impl AccuTreeEpochHandler {
-    /// Creates an instance scoped at a particular epoch, based on 2 state versions: the state
-    /// version of the transaction which started that epoch, and the state version of the last
-    /// committed transaction.
+impl EpochAwareAccuTreeFactory {
+    /// Creates a factory scoped at a particular epoch, based on 2 state versions: the state version
+    /// of the transaction which started that epoch, and the state version of the last committed
+    /// transaction.
     pub fn new(epoch_state_version: StateVersion, current_state_version: StateVersion) -> Self {
+        let epoch_version_count =
+            StateVersion::calculate_progress(epoch_state_version, current_state_version)
+                .and_then(usize::try_from)
+                .unwrap();
         Self {
-            epoch_version_count: StateVersion::calculate_progress(
-                epoch_state_version,
-                current_state_version,
-            )
-            .and_then(usize::try_from)
-            .unwrap(),
+            epoch_state_version,
+            epoch_version_count,
         }
+    }
+
+    /// Creates an accu tree builder which can continue the build of the scoped epoch's accu tree
+    /// at the current version (i.e. declared during this factory's construction).
+    /// The [`previous_epoch_root`] will be used in case we are at the very beginning of the epoch.
+    pub fn create_builder<'s, S: AccuTreeStore<StateVersion, N>, N: IsMerklizableHash>(
+        &'s self,
+        previous_epoch_root: N,
+        store: &'s mut S,
+    ) -> EpochAccuTreeBuilder<S, N> {
+        EpochAccuTreeBuilder::new(
+            store,
+            self.epoch_state_version,
+            previous_epoch_root,
+            self.current_accu_tree_len(),
+        )
+    }
+
+    /// Creates an accu tree merger which can merge the next slice(s) of the scoped epoch's accu
+    /// tree at the current version (i.e. declared during this factory's construction).
+    pub fn create_merger<N>(&self) -> AccuTreeSliceMerger<N> {
+        AccuTreeSliceMerger::new(self.current_accu_tree_len())
     }
 
     /// Returns the actual number of leaves in the epoch's accu tree.
     /// This takes into account the extra first leaf (i.e. previous epoch's tree root), and handles
     /// the "fresh epoch" (i.e. empty accu tree) case.
-    pub fn current_accu_tree_len(&self) -> usize {
+    fn current_accu_tree_len(&self) -> usize {
         if self.epoch_version_count == 0 {
             0
         } else {
             self.epoch_version_count + 1
         }
     }
+}
 
-    /// Adjusts the next planned batch of leaves to be appended to the epoch's accu tree.
-    /// The adjustment consists of prepending the previous epoch's root in case the current epoch
-    /// has just been started.
-    pub fn adjust_next_batch<T>(&self, previous_epoch_root: T, mut next_batch: Vec<T>) -> Vec<T> {
-        if self.epoch_version_count == 0 {
-            next_batch.insert(0, previous_epoch_root);
+/// An [`AccuTree`] wrapper which adds the proper "accu tree per epoch" handling.
+pub struct EpochAccuTreeBuilder<'s, S, N> {
+    previous_epoch_root: Option<N>,
+    epoch_tree_len: usize,
+    epoch_scoped_store: EpochScopedAccuTreeStore<'s, S>,
+}
+
+impl<'s, S: AccuTreeStore<StateVersion, N>, N: IsMerklizableHash> EpochAccuTreeBuilder<'s, S, N> {
+    fn new(
+        store: &'s mut S,
+        epoch_state_version: StateVersion,
+        previous_epoch_root: N,
+        epoch_tree_len: usize,
+    ) -> Self {
+        Self {
+            previous_epoch_root: Some(previous_epoch_root),
+            epoch_tree_len,
+            epoch_scoped_store: EpochScopedAccuTreeStore {
+                store,
+                epoch_state_version,
+            },
         }
-        next_batch
+    }
+
+    /// Appends the next leaf to the epoch's accu tree.
+    /// This method will handle a special case of an epoch that was just started (which requires
+    /// inserting the previous epoch's root as the first leaf of this epoch's tree).
+    pub fn append(&mut self, new_leaf_hash: N) {
+        self.append_batch(vec![new_leaf_hash]);
+    }
+
+    /// A batch variant of the [`append()`] method.
+    pub fn append_batch(&mut self, mut new_leaf_hashes: Vec<N>) {
+        if self.epoch_tree_len == 0 {
+            new_leaf_hashes.insert(0, self.previous_epoch_root.take().unwrap());
+        }
+        let appended_len = new_leaf_hashes.len();
+        AccuTree::new(&mut self.epoch_scoped_store, self.epoch_tree_len).append(new_leaf_hashes);
+        self.epoch_tree_len += appended_len;
+    }
+}
+
+struct EpochScopedAccuTreeStore<'s, S> {
+    store: &'s mut S,
+    epoch_state_version: StateVersion,
+}
+
+impl<'s, S: ReadableAccuTreeStore<StateVersion, N>, N> ReadableAccuTreeStore<usize, N>
+    for EpochScopedAccuTreeStore<'s, S>
+{
+    fn get_tree_slice(&self, epoch_tree_size: &usize) -> Option<TreeSlice<N>> {
+        let end_state_version = self
+            .epoch_state_version
+            .relative(*epoch_tree_size as u64 - 1);
+        self.store.get_tree_slice(&end_state_version)
+    }
+}
+
+impl<'s, S: WriteableAccuTreeStore<StateVersion, N>, N> WriteableAccuTreeStore<usize, N>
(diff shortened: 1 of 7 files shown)


core-rust/state-manager/src/state_manager.rs
```diff
@@ -712,123 +711,132 @@ where
...
+                receipt: commit.local_receipt,
                 identifiers: CommittedTransactionIdentifiers {
                     payload: validated.create_identifiers(),
                     resultant_ledger_hashes: *series_executor.latest_ledger_hashes(),
                 },
             });
         }
 
-        let commit_ledger_hashes = &commit_ledger_header.hashes;
+        if series_executor.next_epoch() != commit_ledger_header.next_epoch.as_ref() {
```
jakrawcz-rdx, line 825: _(note from author)_

I realized it's enough to check this only once, after all transactions.
(the `series_executor.state_tracker` makes sure that an epoch change, if any, happens at the last transaction of the batch)

core-rust/state-manager/src/state_manager.rs
```diff
@@ -712,123 +711,132 @@ where
...
+                series_executor.latest_state_version(),
+                commit_state_version,
+                commit_transactions_len,
+                commit_request_start_state_version
             );
         }
 
+        // TODO(optimization-only): we could reach to the `execution_cache` to find the transaction
+        // root chain there (to avoid this "repeated" calculation here).
+        let resultant_transaction_root = Self::calculate_transaction_root(
```
jakrawcz-rdx, line 746: _(note from author)_

Here is some unnecessary computation (for @malexandru-rdx's attention). There are a couple of ways to mitigate it (none of them perfect; actually the most tempting of them would create a "malicious actor can trigger an OOME" risk).
I went for simplicity, and I indicated my best mitigation idea in that TODO.

core-rust/state-manager/src/staging/epoch_handling.rs
```diff
@@ -62,47 +62,135 @@
...
+        if self.epoch_tree_len == 0 {
+            new_leaf_hashes.insert(0, self.previous_epoch_root.take().unwrap());
         }
-        next_batch
+        let appended_len = new_leaf_hashes.len();
+        AccuTree::new(&mut self.epoch_scoped_store, self.epoch_tree_len).append(new_leaf_hashes);
+        self.epoch_tree_len += appended_len;
+    }
+}
+
```
jakrawcz-rdx, line 170: _(note from author)_

I moved some code from `result.rs` to here;
Regarding the new code: I only formalized the raw "extracted logic" (i.e. `AccuTreeEpochHandler`) into a proper "epoch aware accu tree factory (+wrapper)".

core-rust/state-manager/src/state_manager.rs
```diff
@@ -712,123 +711,132 @@ where
...
                 )
             };
 
-            let (
-                state_hash_tree_diff,
-                transaction_tree_slice,
-                receipt_tree_slice,
-                local_receipt,
-                database_updates,
-            ) = {
```
jakrawcz-rdx, line 776: _(note from author)_

I removed this `let ... = {...}` (and a couple of `.clone()`s) since the entire body got simplified recently and I feel it's readable without the "scoping".
Let me know if you'd rather have this back. But I have plans for some further simplifications here too.

core-rust/state-manager/src/staging/result.rs
```diff
@@ -180,24 +180,34 @@ impl ProcessedCommitResult {
...
+
+        let epoch_accu_trees =
+            EpochAwareAccuTreeFactory::new(epoch_identifiers.state_version, parent_state_version);
+
+        let transaction_tree_diff = {
+            let mut collector = CollectingAccuTreeStore::new(store);
+            epoch_accu_trees
+                .create_builder(epoch_identifiers.transaction_hash, &mut collector)
+                .append(TransactionTreeHash::from(
+                    ledger_transaction_hash.into_hash(),
```
dhedey, line 192: I feel like we should create a `From<LedgerTransactionHash> for TransactionTreeHash` impl

core-rust/state-manager/src/staging/result.rs
```diff
@@ -180,24 +180,34 @@ impl ProcessedCommitResult {
...
-            parent_state_version,
-            ReceiptTreeHash::from(consensus_receipt.get_hash().into_hash()),
-        );
+
+        let receipt_tree_diff = {
+            let mut collector = CollectingAccuTreeStore::new(store);
+            epoch_accu_trees
+                .create_builder(epoch_identifiers.receipt_hash, &mut collector)
+                .append(ReceiptTreeHash::from(
+                    consensus_receipt.get_hash().into_hash(),
```
dhedey, line 206: Ditto for `ReceiptHash` and `ReceiptTreeHash`

core-rust/state-manager/src/staging/result.rs
```diff
@@ -180,24 +180,34 @@ impl ProcessedCommitResult {
...
-            epoch_transaction_identifiers.transaction_hash,
-            parent_state_version,
-            TransactionTreeHash::from(ledger_transaction_hash.into_hash()),
-        );
+
+        let epoch_accu_trees =
+            EpochAwareAccuTreeFactory::new(epoch_identifiers.state_version, parent_state_version);
+
+        let transaction_tree_diff = {
+            let mut collector = CollectingAccuTreeStore::new(store);
```
dhedey, line 188: To be honest, I find all of this indirection, with Factories, Stores, CollectingStores and Trees a little bit confusing. But I take it it's all been tested and is okay :).

Not something I'd suggest to change as part of this PR at any rate.

core-rust/state-manager/src/state_manager.rs
```diff
@@ -712,123 +711,132 @@ where
...
+                self.ledger_transaction_validator
+                    .prepare_from_raw(raw_transaction)
             })
-            .collect();
+            .collect::<Result<Vec<_>, _>>();
+
+        let prepared_transactions = match prepare_results {
+            Ok(prepared_transactions) => prepared_transactions,
+            Err(error) => {
+                warn!("cannot parse transaction to be committed {error:?}",);
```
dhedey, line 726: Is it worth clarifying slightly more clearly that this is when validating a commit request?

It might be nice if this log included the node/source of the transaction, or maybe we already log that in the Java layer.

core-rust/state-manager/src/state_manager.rs
```diff
@@ -712,123 +711,132 @@ where
 
         // Whilst we could validate intent hash duplicates here, these are checked by validators on prepare already,
```
dhedey, line 712: By the way - the panic on line 704 - I assume we check in Java that the sync response contains transactions before sending it to commit in Rust?

core-rust/state-manager/src/state_manager.rs
```diff
@@ -712,123 +711,132 @@ where
...
+        };
 
         let mut write_store = self.store.write();
         let mut series_executor = self.start_series_execution(write_store.deref());
 
         if commit_request_start_state_version != series_executor.latest_state_version() {
             panic!(
-                "Mismatched state versions - the commit request claims {} but the database thinks we're at {}",
-                commit_request_start_state_version, series_executor.latest_state_version()
+                "ledger state version differs from the proof ({} vs ({} - {} == {}))",
```
dhedey, line 736: Hmm this log comment doesn't sound right to me. It suggests to me there is a problem with the committed state, but I don't think that's right.

Java should be creating the extension from the sync response that satisfies the invariant, so I believe this is more of an "invariant broken" style panic - rather than a corrupt database or something?

core-rust/state-manager/src/state_manager.rs
```diff
@@ -712,123 +711,132 @@ where
...
+        // root chain there (to avoid this "repeated" calculation here).
+        let resultant_transaction_root = Self::calculate_transaction_root(
+            write_store.deref(),
+            series_executor.epoch_identifiers(),
+            series_executor.latest_state_version(),
+            &prepared_transactions,
+        );
+        if resultant_transaction_root != commit_ledger_header.hashes.transaction_root {
+            warn!(
+                "resultant transaction root at version {} differs from the proof ({} != {})",
```
dhedey, line 754: Is it worth clarifying this is when validating a commit request?

core-rust/state-manager/src/state_manager.rs
dhedey: This is all good :)

dhedey (approved):

core-rust/state-manager/src/staging/epoch_handling.rs
```diff
@@ -62,47 +62,135 @@
...
+    AccuTreeStore, ReadableAccuTreeStore, TreeSlice, WriteableAccuTreeStore,
+};
+use crate::accumulator_tree::tree_builder::AccuTree;
+use crate::accumulator_tree::IsMerklizableHash;
 use crate::StateVersion;
 
-/// An extracted logic related to the "accu tree per epoch" approach (where the first leaf of the
-/// next epoch's tree is an auto-inserted root of the previous epoch's tree).
-pub struct AccuTreeEpochHandler {
+/// A factory of accu tree utilities operating under the "accu tree per epoch" approach (where the
```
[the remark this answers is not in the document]
jakrawcz-rdx: No exactly "our" term (as in: "not invented by me") but definitely "adopted in our codebase" (by me :P) from one of the papers which described incremental build of merkle trees (see https://github.com/radixdlt/babylon-node/pull/347#discussion_r1126923431).

To be specific, my interpretation of the naming is:
- a classic "merkle tree" is the most narrow term, and denotes a tree built on top of some fixed number of leaf hashes (e.g. a bittorrent uses a classic merkle tree for a file);
- an "accu tree" allows the tree to grow (i.e. to be built incrementally) as new leaf hashes are appended (strictly to the end)
- a "hash tree" is the least constrained term, and we use it for a very dynamic tree of our substates (in which leaf hashes can be added/removed arbitrarily); in our codebase we use "JMT" as synonym, although the exact impl is even more exotic (the "two-tier nested JMT something").

core-rust/state-manager/src/state_manager.rs
jakrawcz-rdx: Yes: https://github.com/radixdlt/babylon-node/blob/524ab532b6f65cf615c4460a1fc87548f8896228/core/src/main/java/com/radixdlt/ledger/StateComputerLedger.java#L329-L334

(to be specific: we only _use_ the sync response to compute the actual "ledger extension" that we pass to Rust commit; and the computation requires basically that "there must be a transaction with state version greater than current ledger's version", which means that list of transactions cannot be empty)

core-rust/state-manager/src/state_manager.rs
jakrawcz-rdx: > It might be nice if this log included the node/source of the transaction, or maybe we already log that in the Java layer.

Yes - the `Err`s are translated to exceptions, and caught, and logged with this detail: https://github.com/radixdlt/babylon-node/blob/33a9a5fd56b6e6dd0720e5e2bc54abc2f962f8e2/core/src/main/java/com/radixdlt/sync/LocalSyncService.java#L455-L460

core-rust/state-manager/src/state_manager.rs
jakrawcz-rdx: true; I now adjusted that wording 👍

core-rust/state-manager/src/staging/result.rs
jakrawcz-rdx: I see what you mean...

I imagine that a true rust developer would have a method like:
`current_tree.compute_next_tree_diff_after_adding(leaf_hashes: Vec<N>) -> AccuTreeDiff<N>`

But OFC a "straight outta java" guy had to make all this needlessly flexible, via 7 interfaces and a `AccuTreeProxyStoreManagerFactoryCreator` implementation.
In java this is less INconvenient (and often arguably LESS complicated) because there are no "borrow checker rules" (I mean it).

In our use-case - it turns out - we pay the complexity price, but we never need the flexibility (i.e. we always kinda collect the result diff and pass it down; we never write to the accu store directly). So it seems like the abstractions could be cut down considerably. TBH I don't find these particular abstractions overly-complicated (in general, splitting a single domain into [a store, a factory, a collector, whatever] is valid to me) and I wouldn't remove them just yet.

BUT it is certainly true that we can hide some of this complexity from the usual use-case that we have! So I now added a `epoch_accu_tree.compute_tree_diff(new_leaf_hashes)` convenience method 👍

core-rust/state-manager/src/staging/epoch_handling.rs
```diff
@@ -62,47 +62,135 @@
...
-    /// Adjusts the next planned batch of leaves to be appended to the epoch's accu tree.
-    /// The adjustment consists of prepending the previous epoch's root in case the current epoch
-    /// has just been started.
-    pub fn adjust_next_batch<T>(&self, previous_epoch_root: T, mut next_batch: Vec<T>) -> Vec<T> {
-        if self.epoch_version_count == 0 {
-            next_batch.insert(0, previous_epoch_root);
+/// An [`AccuTree`] wrapper which adds the proper "accu tree per epoch" handling.
+pub struct EpochAccuTreeBuilder<'s, S, N> {
+    previous_epoch_root: Option<N>,
+    epoch_tree_len: usize,
```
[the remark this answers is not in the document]
jakrawcz-rdx: I now gave it an exact name 👍

core-rust/state-manager/src/staging/result.rs
dhedey: I wouldn't say they're necessarily bad, but just from doing code reviews on low sleep, I haven't spent enough time with them enough to "get" how their abstractions differ / what they are each responsible for 👍.

Happy to keep them :)

core-rust/state-manager/src/staging/epoch_handling.rs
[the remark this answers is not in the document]
dhedey: Ahh nice. Makes sense. Would slightly prefere `AccumulatorTree` to `AccuTree` (on the basis of our coding style generally avoiding shortenings in our type names). But don't mind too much. 

As for the concrete trees, I quite like `TransactionTree` and `ReceiptTree`, and separately, `StateTree` for the TwoTierJMT - I think we use these already 👍

core-rust/state-manager/src/staging/result.rs
dhedey: As for imagining "what an overly-exuberant Rust developer might do"  - personally I'd probably go the trait route - have an `EpochAwareAccumulatorTree` trait with a `Hash: IsHash` associated type, and then just put most of the code in the trait methods :D. But I'm obsessed with traits after all my work on SBOR, traits and transaction versioning :D

PR #498, opened 2023-06-12
Splitting the simplified "commit genesis" logic out of the "regular commit"

jakrawcz-rdx:
I spotted this refactoring opportunity during recent work inside state manager.

Basically, I think the difference in behavior needed for "genesis commit" vs "regular commit" is significant enough to warrant 2 separate methods:
- genesis is already constructed "validated" (no need for parsing/validating)
- genesis does not require all the consistency checks (e.g. commit request cannot "be invalid/mismatching", because we literally just fabricated the proof in the caller)
- genesis transactions are not tracked in mempool/cache

Having just a single "commit" feels cool for many reasons (e.g. the logic won't diverge; the bugfixes/refactors are only needed in one place; etc.). But it became less cool after seeing `if genesis { ...`, hence my refactor.

### DIFF
--- a/core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
+++ b/core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
@@ -451,17 +451,13 @@ impl PendingTransactionResultCache {
     pub fn track_committed_transactions(
         &mut self,
         current_timestamp: SystemTime,
-        previous_state_version: StateVersion,
-        hashes: Vec<IntentHash>,
+        hashes: Vec<(StateVersion, IntentHash)>,
     ) {
-        let mut resultant_state_version = previous_state_version;
-        for intent_hash in hashes {
+        for (state_version, intent_hash) in hashes {
             // Note - we keep the relevant statuses of all known payloads for the intent in the cache
             // so that we can still serve status responses for them - we just ensure we mark them as rejected
-            resultant_state_version = resultant_state_version.next();
-
             self.recently_committed_intents
-                .push(intent_hash, (resultant_state_version, current_timestamp));
+                .push(intent_hash, (state_version, current_timestamp));
 
             if let Some(payload_hashes) = self.intent_lookup.get(&intent_hash) {
                 for cached_payload_hash in payload_hashes {
@@ -474,9 +470,7 @@ impl PendingTransactionResultCache {
                     // because this is a cache for pending transactions, and it can't be re-committed
                     record.track_attempt(TransactionAttempt {
                         rejection: Some(RejectionReason::IntentHashCommitted),
-                        against_state: AtState::Committed {
-                            state_version: resultant_state_version,
-                        },
+                        against_state: AtState::Committed { state_version },
                         timestamp: current_timestamp,
                     })
                 }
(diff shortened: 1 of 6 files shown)


core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
```diff
@@ -451,17 +451,13 @@ impl PendingTransactionResultCache {
     pub fn track_committed_transactions(
         &mut self,
         current_timestamp: SystemTime,
-        previous_state_version: StateVersion,
-        hashes: Vec<IntentHash>,
+        hashes: Vec<(StateVersion, IntentHash)>,
     ) {
-        let mut resultant_state_version = previous_state_version;
-        for intent_hash in hashes {
+        for (state_version, intent_hash) in hashes {
```
jakrawcz-rdx, line 456: :warning: (_important note from author_):

I believe that this is a bugfix:
Before this PR, this method might have been populating wrong `AtState::Committed.state_version` (i.e. it was computing `base_state_version + index`, but the `hashes` only contained user transactions [`intent_hash_if_user()`, remember?], which means that "round change" transactions were disrupting the computed versions).
I switched to passing explicit, definitely-right state version for every intent hash.

dhedey: Yup - agreed. This is much better 👍

core-rust/state-manager/src/state_manager.rs
```diff
@@ -586,119 +590,59 @@ where
                 Consider wiping your database dir and trying again.", top_txn_identifiers.0);
         }
 
-        let genesis_data_chunks_len = genesis_data_chunks.len();
-        let num_genesis_txns = genesis_data_chunks_len + 2; // + bootstrap + wrap up
+        let mut named_system_transactions = Vec::new();
+        named_system_transactions.push((
```
dhedey, line 594: So we previously prepared then executed / stored one transaction at a time, which kept the memory footprint low.

Preparing them all in advance is _probably_ fine (given that Lukasz crunched the numbers of mainnet), but this does increase the memory pressure quite a lot...

It looks like we could change it back to execute one transaction at a time? If you feel this is better, we should probably get Lukasz in to double-check re the memory overhead for mainnet migration.

dhedey (approved):
Approving, pending adjustment or Lukasz input in `commit_genesis`

core-rust/state-manager/src/state_manager.rs
jakrawcz-rdx: I agree, let's not put memory pressure where not needed.

I now tried to solve it using a classic, simple, effective and battle-proven "lazy provider" approach (which would take like ~10s in java).
In rust, I got completely lost somewhere between `Vec<(&'static str, Box<dyn FnOnce() -> ...>)>` and `reference "chunk" may not live long enough` 🤷‍♂️ .

So instead, I unrolled the loop, and it is 3 lines less than when rolled 🤯  

I believe this completely addresses the concern, so I am going to merge after CI passes.

jakrawcz-rdx:
some docker-focused builds are failing (due to some libs/deps, AFAICT...), but the int-tests are passing, so let me merge

core-rust/state-manager/src/state_manager.rs
dhedey: Haha yes, I love a good loop unroll in situations like this. Rust often punishes us for doing silly abstractions, which I've come to view as a feature, not a bug :)

jakrawcz-rdx: > silly abstractions

image

You have insulted an entire Java community, and especially **The `LazyProviderFactoryManager` Club** ®️ 

_(conceptually, a loop which goes over a polymorphic collection of things [== different genesis transactions] which are all supposed to be treated the same way [== prepared and committed] is a very valid thing to me, and not necessarily worth unrolling; but yeah, I see cons of this approach in rust)_

dhedey: Haha! I meant silly for rust of course 😇

PR #502, opened 2023-06-14
Updating scrypto dep to cedar-5547be8cf

jakrawcz-rdx:
Preparing for cedar Node release.


core-rust/state-manager/src/transaction/round_update_transaction.rs
```diff
@@ -150,9 +150,13 @@ impl PreparedRoundUpdateTransactionV1 {
                     initial_proofs: btreeset!(AuthAddresses::validator_role()),
                     virtual_resources: BTreeSet::new(),
                 },
-                fee_payment: FeePayment::NoFee,
+                // TODO(during review): I copied that from Engine's "system transaction" handling
+                fee_payment: FeePayment {
+                    tip_percentage: 0,
+                    free_credit_in_xrd: DEFAULT_FREE_CREDIT_IN_XRD,
```
LukasGasior1, line 156: I believe we should just use `0` here if we use `ExecutionConfig::for_system_transaction()` (which disables the costing module)

core-rust/state-manager/src/transaction/executable_logic.rs
```diff
@@ -63,17 +63,20 @@ impl ExecutionConfigurator {
...
                     ConfigType::Genesis,
-                    ExecutionConfig::genesis().with_trace(trace),
+                    ExecutionConfig::for_genesis_transaction().with_kernel_trace(trace),
                 ),
                 (
                     ConfigType::Regular,
-                    ExecutionConfig::standard().with_trace(trace),
+                    ExecutionConfig::for_notarized_transaction().with_kernel_trace(trace),
                 ),
+                // TODO(during review): are the mappings below correct? should we add `ConfigType::System`?
```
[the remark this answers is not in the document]
jakrawcz-rdx: but what do you qualify as "system txn"?
(the categorization I know [on Node's side](https://github.com/radixdlt/babylon-node/blob/dee19ff8fb097e96c79074102cefb8198ad9e971/core-rust/state-manager/src/transaction/ledger_transaction.rs#L268-L275) is `(genesis == system) vs (user == notarized) vs (round update)` and I now see that it is not really aligned with the Engine's [which has e.g. `for_genesis_transaction()` and `for_system_transactions()`] - ❓ )

core-rust/state-manager/src/transaction/round_update_transaction.rs
jakrawcz-rdx: huh, `0` looks good too, but to be sure, I'd first need an answer to https://github.com/radixdlt/babylon-node/pull/502/files/adffd10c8a6edd04edca1dce230c987ecbf45022#r1230033309

core-rust/state-manager/src/transaction/executable_logic.rs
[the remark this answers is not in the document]
iamyulong: - If we're using `SystemTransactionV1`, we should use `for_system_transaction`
   - Except for genesis transaction, we use `for_genesis_transaction` which disables auth
- If we're using `NotarizedTransactionV1`, we should use `for_notarized_transaction`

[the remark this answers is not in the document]
jakrawcz-rdx: @iamyulong thanks for this ☝️ , the first dot is really helpful.

and what do you recommend to use for `RoundUpdateTransactionV1`? (I guess it is a category introduced by Node, at https://github.com/radixdlt/babylon-node/blob/4baae181cde6f1d1d451017828e7057d0265859a/core-rust/state-manager/src/transaction/round_update_transaction.rs#L140)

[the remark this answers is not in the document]
iamyulong: I'm not aware why we need a dedicated transaction format for round update.

I'd suggest we add a new method `for_round_update_transaction()` with identical module setup as system transaction. If there's specialty about round update transaction, we can change it easily.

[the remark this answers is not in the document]
jakrawcz-rdx: Thanks for explaining 👍 
If I understood you correctly, you want the Node to:
- initially (i.e. in this PR) run it with `for_system_transaction()`...
- ...and potentially in future, introduce `for_round_update_transaction()` in the Engine and switch Node to it.

core-rust/state-manager/src/transaction/round_update_transaction.rs
jakrawcz-rdx: Alright, as you said: this should be a system transaction and `0` 👍

core-rust/state-manager/src/transaction/executable_logic.rs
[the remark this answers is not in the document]
iamyulong: I was thinking the other way

But, after looking into how `RoundUpdateTransactionV1` is used in the node, it's not really a transaction format but some structure for the convenience of node. It should/can be translated into a "system transaction".

So, let's run it with `for_system_transaction()`.

core-rust/state-manager/src/transaction/round_update_transaction.rs
jakrawcz-rdx: alright, so after "round update" started using `ExecutionConfig::for_system_transaction()` (as agreed in https://github.com/radixdlt/babylon-node/pull/502#discussion_r1230543028), then:
- the `0` here is giving me:
`thread '<unnamed>' panicked at 'round update rejected: RejectResult { error: ErrorBeforeFeeLoanRepaid(SystemModuleError(CostingError(FeeReserveError(LoanRepaymentFailed)))) }'`
- the `DEFAULT_FREE_CREDIT_IN_XRD` works fine.

Also, in the Engine, a system transaction uses `DEFAULT_FREE_CREDIT_IN_XRD` - see https://github.com/radixdlt/radixdlt-scrypto/blob/5547be8cf663e1a09df3758c0b400bfc25daaecc/transaction/src/model/v1/system_transaction.rs#L98-L116.
Is this expected behavior from the Engine's side?

_(either way, I am afraid we need `DEFAULT_FREE_CREDIT_IN_XRD`, at least at the moment)_

core-rust/state-manager/src/transaction/executable_logic.rs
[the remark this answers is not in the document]
jakrawcz-rdx: done, PTAL

@iamyulong I also believe you can have helpful insights on other aspects which are marked `TODO(during review)` in this PR (I am aware you can't invest time to read the entire `+5K` PR, so I put those markers to be `ctrl+F`'d during review; there is like ~5 of them and I really have no clue there)

core-rust/state-manager/src/transaction/round_update_transaction.rs
LukasGasior1: That's a bit unexpected, but happy to keep `DEFAULT_FREE_CREDIT_IN_XRD` for now :man_shrugging:

iamyulong: Yes, Lukas is right on this.

The costing module should've been disabled for `SystemTransactionV1` but the `enabled_module` flags aren't being respected in the code base (This has been fixed by https://github.com/radixdlt/radixdlt-scrypto/pull/1112).

For the time being, can we use `DEFAULT_FREE_CREDIT_IN_XRD` and add a `FIXME Change to zero`?

core-rust/core-api-server/src/core_api/handlers/state_non_fungible.rs
```diff
@@ -40,21 +41,22 @@ pub(crate) async fn handle_state_non_fungible(
         .into_response_error("non_fungible_id"));
     }
 
-    let non_fungible = read_optional_collection_substate::<Option<ScryptoRawValue>>(
+    // TODO(during review): I believe I changed the behavior here (previously we returned error for
+    // deleted substate, and now we will return `is_deleted=true`); is this the right thing to do?
```
iamyulong, line 45: Will need someone from network team to help with this. :)

core-rust/core-api-server/src/core_api/conversions/substate.rs
```diff
@@ -307,7 +384,8 @@ pub fn to_api_metadata_value_substate(
         scrypto_decode(key_bytes).map_err(|_| MappingError::InvalidMetadataKey {
             message: "Was not a string".to_string(),
         })?;
-    Ok(match substate {
+    // TODO(during review): do we want to expose the `DynSubstate.mutability` here?
+    Ok(match &substate.value {
```
iamyulong, line 388: I think we should, because we use this flag for user-facing features like royalty config and metadata; clients would like to know if they're mutable or not.

iamyulong (commented):
Only reviewed the `TODO(during review)` parts. :)

core-rust/core-api-server/src/core_api/handlers/state_non_fungible.rs
dhedey: This seems reasonable to me 👍 

So we can return "does not exist", "exists" or "is deleted". Sweet 👍

jakrawcz-rdx:
🎉 

All immediate "doubt" `TODO`s were resolved.
The tests are green.

We can merge as soon as someone approves 🟢

dhedey (approved):

jakrawcz-rdx:
As agreed with @dhedey on our call:

Let me merge this one. Some openapi shape/naming/docu adjustments may come in next PRs.
And actually another "scrypto dep bump" is supposed to happen immediately, including https://github.com/radixdlt/radixdlt-scrypto/pull/1119.

PR #505, opened 2023-06-15
Metrics: Reporting "per-leader rounds as seen by ledger" to prometheus

jakrawcz-rdx:
A pending infra item after implementation of "validator emissions & fees".

### DIFF
--- a/core-rust/state-manager/src/mempool/mod.rs
+++ b/core-rust/state-manager/src/mempool/mod.rs
@@ -68,25 +68,12 @@ use transaction::{errors::TransactionValidationError, prelude::NotarizedTransact
 
 use std::string::ToString;
 
-use crate::MetricLabel;
-
 pub use crate::pending_transaction_result_cache::*;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum MempoolAddSource {
     CoreApi,
     MempoolSync,
-}
-
-impl MetricLabel for MempoolAddSource {
-    type StringReturnType = &'static str;
-
-    fn prometheus_label_name(&self) -> Self::StringReturnType {
-        match *self {
-            MempoolAddSource::CoreApi => "CoreApi",
-            MempoolAddSource::MempoolSync => "MempoolSync",
-        }
-    }
 }
 
 #[derive(Debug)]
@@ -139,22 +126,6 @@ impl MempoolAddRejection {
             AtState::Committed { .. } => self.reason.is_rejected_because_intent_already_committed(),
             AtState::PendingPreparingVertices { .. } => false,
             AtState::Static => false,
-        }
-    }
-}
-
-impl MetricLabel for MempoolAddError {
-    type StringReturnType = &'static str;
-
-    fn prometheus_label_name(&self) -> Self::StringReturnType {
-        match self {
-            MempoolAddError::Rejected(rejection) => match &rejection.reason {
-                RejectionReason::FromExecution(_) => "ExecutionError",
-                RejectionReason::ValidationError(_) => "ValidationError",
-                RejectionReason::IntentHashCommitted => "IntentHashCommitted",
-            },
-            MempoolAddError::Full { .. } => "MempoolFull",
-            MempoolAddError::Duplicate(_) => "Duplicate",
         }
     }
 }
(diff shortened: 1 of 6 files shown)


core-rust/state-manager/src/mempool/mod.rs
```diff
@@ -78,17 +76,6 @@ pub enum MempoolAddSource {
     MempoolSync,
 }
 
-impl MetricLabel for MempoolAddSource {
```
jakrawcz-rdx, line 81: _(note from author)_

I moved it to the metric-related file.

core-rust/state-manager/src/state_manager.rs
```diff
@@ -650,28 +650,36 @@ where
...
+                Err(error) => {
+                    warn!(
+                        "invalid commit request: cannot parse transaction at index {}: {:?}",
+                        index, error
+                    );
+                    return Err(InvalidCommitRequestError::TransactionParsingFailed);
+                }
+            };
+
+            if let PreparedLedgerTransactionInner::RoundUpdateV1(_) = &prepared_transaction.inner {
```
jakrawcz-rdx, line 671: _(note from author)_

Yes, this is ugly, since we have to re-parse the transaction that was already parsed. Is there a way to use the already-parsed one? Or to capture the `RoundUpdateV1` representation _before_ it becomes `instructions: Vec<_>`?
(From performance PoV, I am sure we can afford this - it only happens for the single round update transaction, not every transaction)

There are also some completely different approaches possible (e.g. emit this information in an Event from Engine - but parsing an Event is not the easiest either!)

core-rust/state-manager/src/transaction/round_update_transaction.rs
```diff
@@ -169,3 +163,90 @@ impl HasRoundUpdateTransactionHash for PreparedRoundUpdateTransactionV1 {
...
+/// A builder of "successful/timeout/gap rounds by leader" counter update, for metrics purposes.
+#[derive(Default)]
+pub struct LeaderRoundCountersBuilder {
+    /// Counters per validator (of the corresponding [`ValidatorIndex`]).
+    /// Implementation note:
+    /// This structure starts as an empty one (on the builder's initialization via `default()`).
+    /// Then, it is lazily-initialized by any larger-than-previously-observed [`ValidatorIndex`]
+    /// (within `update()`; see `get_counter_mut()`). The size of this vector is thus effectively
+    /// bounded by `u8::MAX` (even if validator indices found in proposal history were invalid).
+    // TODO(during review): actually it's hard to say if this is more performant than `BTreeMap<ValidatorIndex, _>`...?
```
jakrawcz-rdx, line 176: _(note from reviewer)_

I am really curious what reviewers think!

core-rust/state-manager/src/metrics.rs
```diff
@@ -226,3 +237,59 @@ impl<T: MetricLabel> MetricLabel for &T {
...
+    fn prometheus_label_name(&self) -> Self::StringReturnType {
+        self.to_hex()
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ConsensusRoundResolution {
+    Successful,
+    Fallback,
+    Gap,
```
jakrawcz-rdx, line 255: _(note from reviewer)_

I am open to any renames of the label names and values in this PR.

core-rust/state-manager/src/transaction/round_update_transaction.rs
LukasGasior1: I agree, " it's hard to say" :trollface: happy to go with a Vec

core-rust/state-manager/src/metrics.rs
LukasGasior1: We generally call them `missed proposals` so maybe:
- Successful
- MissedAndFallback
- MissedAndTimeout
?

LukasGasior1 (approved):

core-rust/state-manager/src/transaction/round_update_transaction.rs
malexandru-rdx: 100% faster, would be even faster with a preallocation.

PR #510, opened 2023-06-19
Add vertex limits & enforce inside prepare

malexandru-rdx:
As discussed in [RNP-10](https://radixdlt.atlassian.net/wiki/spaces/RPNV1/pages/3067904346/RNP-10+-+Mempool+Vertex+Prepare+Proposal), in this PR I add vertex limits and their enforcement inside of prepare.


core-rust/state-manager/src/state_manager.rs
```diff
@@ -425,21 +470,49 @@ where
...
+                        }
+                        Err(error) => {
+                            rejected_transactions.push(RejectedTransaction {
+                                index: index as u32,
+                                intent_hash: Some(intent_hash),
+                                notarized_transaction_hash: Some(notarized_transaction_hash),
+                                ledger_transaction_hash: Some(ledger_transaction_hash),
+                                error: format!("{:?}", &error),
+                            });
+                            // Note: we are not adding this transaction to [`pending_transaction_results`] because
```
malexandru-rdx, line 512: _(Note from author)_

We will want to revisit mempool eviction based on prepare failure as it's too agressive anyway.

core-rust/state-manager/src/limits.rs
```diff
@@ -0,0 +1,179 @@
...
+        self.total_transactions_count += 1;
+        self.total_transactions_size += transaction_size;
+        self.total_execution_cost_units_consumed += execution_metrics.execution_cost_units_consumed;
+        self.total_substate_read_count += execution_metrics.substate_read_count;
+        self.total_substate_read_size += execution_metrics.substate_read_size;
+        self.total_substate_write_count += execution_metrics.substate_write_count;
+        self.total_substate_write_size += execution_metrics.substate_write_size;
+    }
+
+    pub fn is_full(&self, limits_config: &VertexLimitsConfig) -> bool {
```
jakrawcz-rdx, line 169: The `limits_config` seems to be a param to almost all methods, so (e.g. in java) it would be no-brainer to make it an instance field. Is this difficult due to borrow-checker here?

core-rust/state-manager/src/state_manager.rs
```diff
@@ -425,21 +470,49 @@ where
...
+                Ok(processed_commit_result) => {
+                    match vertex_limits_tracker.check_post_execution(
+                        &self.vertex_limits_config,
+                        &processed_commit_result
+                            .local_receipt
+                            .local_execution
+                            .execution_metrics,
+                    ) {
+                        Ok(_) => {
+                            vertex_limits_tracker.advance(
```
jakrawcz-rdx, line 482: hmm, couldn't `.advance()` just do the "post-execution check" in its beginning, and return "yeah cool, advanced | whoops, some limit exceeded" ?

(if this seems unreasonable, then it's probably me being a psycho-fan of narrow APIs)

core-rust/state-manager/src/state_manager.rs
```diff
@@ -333,6 +362,22 @@ where
...
+            // (most probable by transaction count).
+            if vertex_limits_tracker.is_full(&self.vertex_limits_config) {
+                break;
+            }
+
+            let transaction_size = raw_user_transaction.as_slice().len();
+
+            // Skip validating and executing this transaction if it doesn't fit it in the vertex.
+            if vertex_limits_tracker
+                .check_pre_execution(&self.vertex_limits_config, transaction_size)
```
jakrawcz-rdx, line 375: it feels a bit weird that `.is_full()` and `.check_pre_execution()` co-exist;
even weirder that they are called one after another;
and even weirder that they both check `total_transactions_count` (effective one of those is 0xDEADC0DE, isn't it?)

AFAIU, conceptually we want to distinguish between 2 cases:
A. "some limit was exceeded by this transaction, but it's worth checking another one!"
B. "some limit was exceeded by this transaction and will definitely be exceeded by any other"
(that B. is a bit artificial, because effectively only the tx count can cause this, right?)

anyways, maybe it's possible to indicate this in the result returned from `check_pre_execution()`?
(and drop `is_full()`)

core-rust/node-common/src/config/limits.rs
```diff
@@ -0,0 +1,109 @@
...
+ * permissions under this License.
+ */
+
+use radix_engine_constants::{
+    DEFAULT_MAX_SUBSTATE_READS_PER_TRANSACTION, DEFAULT_MAX_SUBSTATE_WRITES_PER_TRANSACTION,
+};
+
+// TODO: revisit & tune before Babylon
+const DEFAULT_MAX_TOTAL_TRANSACTIONS_COUNT: usize = 10;
+const DEFAULT_MAX_TOTAL_TRANSACTIONS_SIZE: usize = 4 * 1024 * 1024;
```
dhedey, line 71: These two could both do with specifying VERTEX_ like the rest.

core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
```diff
@@ -21,6 +22,7 @@ pub enum RejectionReason {
     FromExecution(Box<RejectionError>),
     ValidationError(TransactionValidationError),
     IntentHashCommitted,
+    VertexLimit(VertexLimitsExceeded),
```
dhedey, line 25: It doesn't feel to me that this should be a rejection - rejections all currently imply "at some state version, this transaction was rejected / uncommittable". And we handle rejections by (eg) booting the transaction from the mempool and adding them to the rejection cache.

I realise you've given it a base delay of 0, but even so, it feels like it shouldn't be a rejection to start with.

core-rust/state-manager/src/receipt.rs
```diff
@@ -136,6 +137,7 @@ pub struct LedgerTransactionReceipt {
 /// It is not verifiable against ledger, but may still be useful for debugging.
 #[derive(Debug, Clone, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
 pub struct LocalTransactionExecution {
+    pub execution_metrics: ExecutionMetrics,
```
dhedey, line 140: Very minor, but in terms of field ordering, I'd suggest it's less important than, and therefore should be ordered below `outcome`

core-rust/node-common/src/config/limits.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub max_total_transactions_count: usize,
+    pub max_total_transactions_size: usize,
+    pub max_total_execution_cost_units_consumed: usize,
+    pub max_total_substate_read_size: usize,
+    pub max_total_substate_read_count: usize,
+    pub max_total_substate_write_size: usize,
+    pub max_total_substate_write_count: usize,
+}
+
+impl VertexLimitsConfig {
```
dhedey, line 90: I'm surprised not to see a separate ProposalLimitsConfig

core-rust/state-manager/src/limits.rs
```diff
@@ -0,0 +1,179 @@
...
+
+        Ok(())
+    }
+
+    pub fn check_post_execution(
+        &self,
+        limits_config: &VertexLimitsConfig,
+        execution_metrics: &ExecutionMetrics,
+    ) -> Result<(), VertexLimitsExceeded> {
+        if self.total_execution_cost_units_consumed
```
dhedey, line 125: I assume we're safe from overflows here due to limits in the engine?

core-rust/state-manager/src/limits.rs
```diff
@@ -0,0 +1,179 @@
...
+            total_transactions_size: 0,
+            total_execution_cost_units_consumed: 0,
+            total_substate_read_size: 0,
+            total_substate_read_count: 0,
+            total_substate_write_size: 0,
+            total_substate_write_count: 0,
+        }
+    }
+
+    pub fn check_pre_execution(
```
dhedey, line 102: This is fine, but if we refactor, I might consider to track the remaining instead of the total, and combine the "advance" and "check_" functions to be `consume_pre_execution(..) -> Result<(), VertexLimitsExceeded>`

core-rust/state-manager/src/state_manager.rs
dhedey: I suggested this independently in the other file, where I said it was optional, but given Jakub also mentioned this, I suggest it's worth doing 👍

core-rust/state-manager/src/state_manager.rs
dhedey: > We will want to revisit mempool eviction based on prepare failure as it's too aggressive anyway.

Maybe, but it's actually quite an effective performance boost in the 99% case where we have a linear vertex chain.

core-rust/node-common/src/config/limits.rs
dhedey: Was there a separate PR for proposal limits that I missed?

dhedey (commented):
Looks good. Happy to re-review quickly when markups are in

core-rust/state-manager/src/limits.rs
jakrawcz-rdx: > track the remaining instead of the total

I like this approach (it would also bypass the "`limits_config` passed everywhere" issue https://github.com/radixdlt/babylon-node/pull/510#discussion_r1234445596).
The only downsides are:
- errors won't surface the nice "got `X`, but max is `Y`" messages _(not a big problem for us, since our `Y` is not very dynamic)_
- some broad "constraints" cannot be tracked in a "countdown" manner _(also not yet our problem, since all our constraints so far are of type "limit")_

@malexandru-rdx (just a suggestion, you decide) I wonder if we would benefit from introducing a proper "math"-like abstractions here...
E.g. a `struct VertexSize { transactions_count, transactions_bytes, substate_reads_count, ... }`.
And then:
- the `VertexLimitsConfig` is gone (i.e. it is implemented as `pub config_max_vertex_size: VertexSize,`)
- the `VertexLimitsTracker` simply has `pub current: VertexSize, pub limit: VertexSize,`
- you can introduce "math" like `(self.current + added) < self.limit` or whatever you find helpful

core-rust/state-manager/src/state_manager.rs
malexandru-rdx: Agree.
For this particular case though, would not be right to remove them (transaction just didn't fit in vertex limits).

core-rust/state-manager/src/limits.rs
malexandru-rdx: Right now, yes.

core-rust/state-manager/src/limits.rs
malexandru-rdx: > This is fine, but if we refactor, I might consider to track the remaining instead of the total, and combine the "advance" and "check_" functions to be consume_pre_execution(..) -> Result<(), VertexLimitsExceeded>

Initially I wrote it with a single `try_advance` method, but I realised it's not ideal to execute transaction if it doesn't fit by size anyway.
I agree merging `check_post_execution` with `advance` though.

core-rust/state-manager/src/limits.rs
malexandru-rdx: It was mere personal preference (pushing data ownership upwards if only read and let parent delegate strictly when needed). I'll capture it by ref.

core-rust/state-manager/src/limits.rs
malexandru-rdx: > (just a suggestion, you decide) I wonder if we would benefit from introducing a proper "math"-like abstractions here...
Nice idea, only thing I don't like it's we still have to add 7 ifs to get the corresponding `VertexLimitExcedeed`.

core-rust/node-common/src/config/limits.rs
malexandru-rdx: It is already on the Java side. I will capture it on the Rust side in a future refactor (and pass to Java).

core-rust/state-manager/src/limits.rs
malexandru-rdx: I've refactor to `remaining_`, so this is not an issue anymore.

core-rust/state-manager/src/limits.rs
```diff
@@ -0,0 +1,177 @@
...
+impl VertexLimitsTracker {
+    pub fn new(config: &VertexLimitsConfig) -> Self {
+        Self {
+            remaining_transactions_count: config.max_total_transactions_count,
+            remaining_transactions_size: config.max_total_transactions_size,
+            remaining_execution_cost_units_consumed: config.max_total_execution_cost_units_consumed,
+            remaining_substate_read_size: config.max_total_substate_read_size,
+            remaining_substate_read_count: config.max_total_substate_read_count,
+            remaining_substate_write_size: config.max_total_substate_write_size,
+            remaining_substate_write_count: config.max_total_substate_write_count,
```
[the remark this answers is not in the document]
malexandru-rdx: I specifically switched to `remaining` to avoid possible overflow issues (since limits are `usize`) - yes, this can be fixed having `i64` for `VertexCounters`. 
I respect the elegant code you provided, but as you already mentioned the ugliness is just hidden. I would happily refactor like this if it would be duplicate code :).

Most of the limits work seems like could be macro autogenerated code to some degree in the future.

Either way, I would like to keep `*Config` inside of `node-common` (without knowledge of `VertexCounters`).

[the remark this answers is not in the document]
jakrawcz-rdx: > seems like could be macro autogenerated code to some degree in the future.

yeah, true, it is an even better target 👍 
let's not try too hard right now, the PR is good to merge as-is

jakrawcz-rdx (approved):

core-rust/state-manager/src/state_manager.rs
```diff
@@ -500,6 +480,10 @@ where
                                 invalid_at_epoch,
                                 rejection_reason: None,
                             });
+                            match success {
+                                VertexLimitsAdvanceSuccess::VertexFilled => break,
+                                VertexLimitsAdvanceSuccess::VertexNotFilled => {}
+                            }
                         }
                         Err(error) => {
                             rejected_transactions.push(RejectedTransaction {
```
dhedey, line 489: We possibly should (EG) consider stopping executing the proposal the first time a limit is breached, which may prove a more efficient use of time for the network. Otherwise a proposal full of slow transactions would still take a long time to execute, even if only (say) the first two could be committed.

dhedey: In any case, I'm looking forward to us having a node metric here so we understand how much wasted work there is executing transactions we can't fit.

dhedey (approved):
Approving so we can merge this PR - but I'd like to pick up the discussion around what to do with lots of rejections on slack.

core-rust/state-manager/src/limits.rs
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }
+
+    pub fn check_pre_execution(&self, transaction_size: usize) -> Result<(), VertexLimitsExceeded> {
+        if self.remaining_transactions_count < 1 {
+            return Err(VertexLimitsExceeded::TransactionsCount);
+        }
+
+        if self.remaining_substate_read_size < transaction_size {
```
jakrawcz-rdx, line 113: hah, a bit late, but I noticed that you probably meant `self.remaining_transactions_size` here :p

I am fixing this during mu update from develop to https://github.com/radixdlt/babylon-node/pull/519.

PR #511, opened 2023-06-19
Indexing a Node's parent and root Substate by post-processing Substate changes on commit()

jakrawcz-rdx:
This is the first PR in the ["parent map" effort](https://whimsical.com/babylon-work-tracker-Am2s7hSa7xHd5yCxgub67a@2bsEvpTYSt1HiyHZkipb9svniAiXqVnUZKo).
In next PR this new table will be used to improve "balance changes" responses (from LTS endpoints).

The PR is rather large:
- many lines are actually only moved:
  - most notably, the JNI calls which were defined only for testing in many places are now moved to a single place (`TestStateReader`
- I added a few low-level tests (for metadata resolver logic), and also a mid-level test (doing a read-out of the index table); a high-level test will come with the feature PR itself


core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -142,6 +148,7 @@ const ALL_COLUMN_FAMILIES: [RocksDBColumnFamily; 17] = [
     LedgerProofByStateVersion,
     LedgerProofByEpoch,
     Substates,
+    SubstateNodeMetadatas,
```
dhedey, line 151: Ditto, `SubstateNodeMetadata`*

core-rust/state-manager/src/store/traits.rs
```diff
@@ -160,9 +160,62 @@ pub mod vertex {
...
+                .map(|node_id| self.get_metadata(node_id))
+                .collect()
+        }
+    }
+
+    /// Relevant metadata of a RE Node (which is not provided by the RE itself).
+    #[derive(Debug, Clone, Eq, PartialEq, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
+    pub struct SubstateNodeMetadata {
+        /// A substate owning the Node (i.e. its immediate parent).
+        /// Note: this will always be present, since we do not track metadata of the root RE Nodes.
```
dhedey, line 207: I quite like the name .... but given that we've called it `Metadata` now, I'm slightly surprised we don't track metadata for global nodes

core-rust/state-manager/src/store/parent_map.rs
```diff
@@ -0,0 +1,315 @@
...
+ * permissions under this License.
+ */
+
+use crate::store::traits::*;
+use crate::{ChangeAction, SubstateChange};
+
+use radix_engine::types::*;
+
+/// A resolver of [`SubstateNodeMetadata`] for [`SubstateChange`]s.
+pub struct NodeMetadataResolver {
```
dhedey, line 71: I like the name. Perhaps the file should be renamed from `parent_map` to match? 👍

core-rust/state-manager/src/store/parent_map.rs
```diff
@@ -0,0 +1,315 @@
...
+    /// A map of created Nodes to their parent Substates, extracted from the [`SubstateChange`]s.
+    changed_parent_by_node: NonIterMap<NodeId, SubstateReference>,
```
dhedey, line 73: Perhaps `parents_for_created_node`?

The engine doesn't support reparenting, and if it did, this approach wouldn't be sufficient anyway - because if A got reparented from B to C, then any of A's descendents wouldn't have their root updated.

core-rust/state-manager/src/store/parent_map.rs
```diff
@@ -0,0 +1,315 @@
...
+                    let parent_reference = SubstateReference(
+                        substate_change.node_id,
+                        substate_change.partition_number,
+                        substate_change.substate_key.clone(),
+                    );
+                    Some((parent_reference, owned_node_ids))
+                }
+                ChangeAction::Delete => None,
+            })
+            .collect::<IndexMap<_, _>>();
```
dhedey, line 110: I'm slightly worried about edge case behaviour here - EG conceivably one substate could have 1000 children and updates some irrelevant data - but we're still querying our DB for all 1000 children each update here.

It may be more efficient to filter to only nodes that have been created here... Which I thought we could easily get from the receipt, but looks like we don't have that information available :'(.

Although actually, I believe we could also resolve the "previous" values for updates in `compute_substate_changes`, right? (and in fact, maybe we should even persist this as part of `ChangeAction`? This would be helpful for Gateway for sure)

If we had that, we could compare "before" and "after", and only need to resolve/add newly created nodes, which would limit the number of lookups we need to do here quite a lot I imagine.

core-rust/state-manager/src/store/parent_map.rs
```diff
@@ -0,0 +1,315 @@
...
+
+        // Load the missing metadata of these parents from the `SubstateNodeMetadataStore`:
+        let existing_metadatas = metadata_store.batch_get_metadata(&top_level_parent_node_ids);
+
+        // Construct a Node -> root map needed for the `NodeMetadataResolver` helper structure:
+        let existing_root_by_node = top_level_parent_node_ids
+            .into_iter()
+            .zip(existing_metadatas)
+            .filter_map(|(parent, existing_metadata)| {
+                existing_metadata.map(|metadata| (parent, metadata.root))
```
dhedey, line 133: If `existing_metadata` doesn't exist for `parent`, it could be because the `parent` is global. But it seems we filter that out instead of mapping to `(parent, parent)` - is this an issue? Or have I misunderstood something?

dhedey: Ahh this is handled by the resolver 👍  nice. ignore this.

core-rust/state-manager/src/store/parent_map.rs
```diff
@@ -0,0 +1,315 @@
...
+
+        let resolver = Self {
+            changed_parent_by_node,
+            existing_root_by_node,
+        };
+
+        // Simply query the fully-preloaded `NodeMetadataResolver` for each Node:
+        changed_nodes_by_parent
+            .into_iter()
+            .map(move |(parent, child_node_ids)| {
```
dhedey, line 145: This is O(N^2) in the worst case without a cache, or more technically O(N * D) for D the max depth, which I think can be D ~ N for nested key value stores.

core-rust/state-manager/src/store/parent_map.rs
```diff
@@ -0,0 +1,315 @@
...
+                let root = resolver.resolve_root(&parent);
+                (child_node_ids, SubstateNodeMetadata { parent, root })
+            })
+    }
+
+    /// Resolves the root Substate of the given one by traversing the internal maps:
+    /// - the given Substate is assumed to originate from the [`SubstateChange`]s, so we first
+    ///   follow the parent links through the [`changed_parent_by_node`];
+    /// - after this walk ends, we check the [`existing_root_by_node`], which may contain a hit
```
dhedey, line 154: Can we explain that if it doesn't contains a hit, then it's global?

core-rust/state-manager/src/store/parent_map.rs
```diff
@@ -0,0 +1,315 @@
...
+
+impl NodeMetadataResolver {
+    /// Resolves the [`SubstateNodeMetadata`] for all newly-created Nodes found in the [`Create`]d
+    /// and [`Update`]d substates. (within the given [`SubstateChange`]s).
+    /// The resolution considers the trees (or rather: potentially-incomplete bottom-up tree
+    /// fragments) constructed within the given [`SubstateChange`]s, and the top fragments of these
+    /// trees already-existing in the given [`SubstateNodeMetadataStore`].
+    /// API note: the results are grouped by [`SubstateNodeMetadata`] (since it may be shared by
+    /// many [`NodeId`]s).
+    pub fn batch_resolve<S: SubstateNodeMetadataStore>(
```
dhedey, line 88: Nice. Didn't quite realise how intrinsically complicated this resolution is, but it looks correct to me, and can't see any obvious simplifications :D

core-rust/state-manager/src/store/parent_map.rs
dhedey: I don't know what your thoughts are / whether you think it's worth adding a cache to reduce it to O(N)? Might not be worthwhile in practice, happy for you to make that call

core-rust/state-manager/src/store/traits.rs
dhedey: Calling it something more like `SubstateNodeAncestorMetadata` would make the fact it's missing for Global nodes feel more reasonable to me

dhedey (approved):
Really nice, love the tests. And +1 on moving the test DB read methods to a separate JNI file.

Approving, although there are a couple of medium-importance reflections on speeding up the parent map resolution in some edge cases that may be worth resolving

Also shared a tangentially related thought that we store the old substate value if it'd help with this (as I think it will have use not just for this, but also for the Core API and other data feeds we might create). Input welcome again 👍

core-rust/state-manager/src/store/rocks_db.rs
jakrawcz-rdx: huh, then it clashes with a struct of the same name :(
I'll figure something out.

core-rust/state-manager/src/store/traits.rs
jakrawcz-rdx: I imagine we may be tempted to put more fields (than just `Ancestor`s) into that table, and then the `SubstateNodeMetadata` name is very nice, and the fields may be applicable to roots too.
In current state, the roots would only be deadweight (i.e. a row of a root `X` is computable: `{parent=None, root=X}`). If you insist though, I can put them there.

core-rust/state-manager/src/store/parent_map.rs
jakrawcz-rdx: Cool idea, I think I'll be able to neatly integrate the cache 👍

core-rust/state-manager/src/store/parent_map.rs
jakrawcz-rdx: > but we're still querying our DB for all 1000 children each update here.

I don't see this.
I definitely query the DB about the _parent_ of these children (i.e. the parent == the substate that was created/updated here), in a 1 batch read, to discover the root of that parent.

The inefficiency I see though:
We will create 1000 Metadata rows, and we will upsert them on commit, which will be completely needless since they already exist :(

I agree that this could be mitigated by adding `previous` to `ChangeAction::Update()`.
It's definitely worthwhile - should we treat it as a performance optimization that happens in next PR?

dhedey: Ahh right, I didn't look far enough down to work out what the inefficiency actually was, but knew we needed to handle these 1000 things somehow. Happy to do it in the next PR.

core-rust/state-manager/src/store/in_memory.rs
```diff
@@ -123,6 +125,7 @@ impl InMemoryStore {
             epoch_proofs: BTreeMap::new(),
             vertex_store: None,
             substate_store: InMemorySubstateDatabase::standard(),
+            node_metadatas: BTreeMap::new(),
```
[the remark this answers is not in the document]
jakrawcz-rdx: As discussed on a call, this gets renamed to `node_ancestry_record`s, to avoid 2 issues (the plural naming and the non-existence of records for roots)

core-rust/state-manager/src/store/traits.rs
jakrawcz-rdx: as discussed on a call: the root records would be useless and would have to be migrated (together with all other records) even if we plan to use this table for other (more relevant for roots) metadata in future. 
And the conceptual issue (i.e. "missing" rows) goes away with the rename to `...Ancestry...`.

core-rust/state-manager/src/store/parent_map.rs
jakrawcz-rdx: As discussed on our call, I went for a more proper "propagate the topmost parents down" approach (which is definitely O(N)).

Achieving it without recursion (and without rewriting `Vec<>`s until it is not O(N) anymore) was a challenge, and resulted in a less-then-perfectly-readable code, so I added heavy comments and extra unit tests. PTAL.

All the other rename requests are already applied.
The other optimization (https://github.com/radixdlt/babylon-node/pull/511#discussion_r1235023152) happens in next PR.

dhedey (approved):
Nice!

core-rust/state-manager/src/store/node_ancestry_resolver.rs
```diff
@@ -0,0 +1,486 @@
...
+        // `SubstateChange`s) and `topmost_parent_roots` (roots of the trees, loaded from DB).
+        let parent_to_root = transitively_owned_parent_sets
+            .iter()
+            .zip(topmost_parent_roots)
+            .flat_map(|((topmost_parent, transitively_owned_parents), root)| {
+                let root_for_topmost = root.clone();
+                transitively_owned_parents
+                    .iter()
+                    .map(move |owned_parent| (*owned_parent, root.clone()))
+                    // interesting gotcha: the topmost parent is also "a parent" (and won't be on any child-list)
```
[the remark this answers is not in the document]
jakrawcz-rdx: > What if the top-most parent is itself a root here?

Absolutely not a problem: the `parent_to_root` map will then simply contain an entry like `topmost_parent=7 -> root=7`. And the iteration at the end will query it out, and it will be a correct answer.

But I agree: if it is itself a root, we can totally skip it from `parent_to_root`, and the iteration at the end will correctly assume it is a root then.

So I applied this change and added extra comments to the gotcha.

I adjusted `newly_created_root_nodes_are_not_recorded_in_index` to show that, as you asked 👍 
(and it is passing for both versions, before and after the change)

jakrawcz-rdx:
Thank you for a detailed review in this hard times 🙇 

Everything addressed, let me merge after seeing 🟢 tests.

PR #512, opened 2023-06-20
Export transaction ExecutionMetrics to Prometheus

malexandru-rdx:
This builds on top of [510](https://github.com/radixdlt/babylon-node/pull/510).

In this PR I export transaction size and execution metrics to Prometheus.


core-rust/state-manager/src/metrics.rs
```diff
@@ -111,6 +130,89 @@ impl LedgerMetrics {
...
+                    "Maximum invoke payload size in bytes per committed transaction.",
+                ),
+                Self::buckets(DEFAULT_MAX_INVOKE_INPUT_SIZE),
+            )
+            .registered_at(registry),
+        }
+    }
+
+    // Given a limit, builds buckets for a Histogram with higher resolution for lower values.
+    fn buckets(limit: usize) -> Vec<f64> {
```
malexandru-rdx, line 207: _(question from author)_

Let me know if this seems reasonable enough (or if hand-picked buckets are preferred).

dhedey (approved):

core-rust/state-manager/src/metrics.rs
dhedey: Although might be worth a quick comment?

core-rust/state-manager/src/metrics.rs
```diff
@@ -111,6 +130,89 @@ impl LedgerMetrics {
     }
 }
 
+// TODO: update buckets limits when default values are overwritten
```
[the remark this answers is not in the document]
malexandru-rdx: Will do. There are still some limits not yet captured in config, so this comment remains true.

PR #513, opened 2023-06-21
Add max size limit in the address book

LukasGasior1:


core/src/main/java/com/radixdlt/p2p/addressbook/AddressBook.java
```diff
@@ -158,49 +166,46 @@ public AddressBook(
...
-    } else if (filteredKnownAddresses.size() != entry.getKnownAddresses().size()) {
-      // some addresses got filtered out, need to persist a new entry
-      final var updatedEntry =
-          new AddressBookEntry(entry.getNodeId(), entry.bannedUntil(), filteredKnownAddresses);
-      cleanedUpEntries.put(entry.getNodeId(), updatedEntry);
-      persistEntry(updatedEntry);
-    } else {
-      cleanedUpEntries.put(entry.getNodeId(), entry);
+    // Reserving up to 1/10 of the address book for high priority peers
+    this.highPriorityPeers = new LRUCache<>(p2pConfig.addressBookMaxSize() / 10);
```
malexandru-rdx, line 173: I would make it `min(p2pConfig.addressBookMaxSize() / 10, MAX_NUM_VALIDATORS)` just to be sure, since it's not that big anyway.

malexandru-rdx (approved):
LGTM!

core/src/main/java/com/radixdlt/p2p/addressbook/AddressBook.java
```diff
@@ -209,10 +214,10 @@ private void insertOrUpdateAddressBookEntryWithUri(RadixNodeUri uri) {
     final var newOrUpdatedEntry =
         maybeExistingEntry == null
             ? AddressBookEntry.create(uri)
-            : maybeExistingEntry.cleanupExpiredBlacklsitedUris().addUriIfNotExists(uri);
+            : maybeExistingEntry.cleanupExpiredBlacklistedUris().addUriIfNotExists(uri);
```
dhedey, line 217: I wonder if this is a good time to `sed/blacklist/blocklist` and `sed/whitelist/allowlist` in the codebase?

core/src/main/java/com/radixdlt/p2p/addressbook/AddressBook.java
```diff
@@ -144,9 +143,18 @@ void resetFailures(RadixNodeUri uri) {
   private final AddressBookPersistence persistence;
   private final Object lock = new Object();
   private final Map<NodeId, AddressBookEntry> knownPeers = new ConcurrentHashMap<>();
+  private int numStoredAddresses;
   private final PeerAddressEntryComparator addressEntryComparator =
       new PeerAddressEntryComparator();
 
+  /**
+   * Stores nodes whose addresses were explicitly requested (via `bestKnownAddressesById`). The
```
dhedey, line 151: I've finished my PR and come back and I'm still not sure I understand this.

I think there's a link I'm missing.
* Why is `bestKnownAddressesById` relevant? What does it mean? When is it called?
* Why should I prioritise these in my address book? What's the motivation?

Is eg a validator id discovered in a proof, so a node tries to sync from it - so the node attempts a connection to this node, which ends up hitting `bestKnownAddressesById`? And because the node was interested in connecting to this other node, it's probably worth remembering (or something?). If so, I'm wondering if putting it in `bestKnownAddressesById` is the right name/abstraction?

LukasGasior1: > Is eg a validator id discovered in a proof, so a node tries to sync from it - so the node attempts a connection to this node, which ends up hitting bestKnownAddressesById? And because the node was interested in connecting to this other node, it's probably worth remembering (or something?).

Yup, exactly that.
PeerManager/PeerDiscovery use two AddressBook's methods to init outbound connections:
`bestCandidatesToConnect` - "I just want to connect to _some_ peers"
 `bestKnownAddressesById` - "I want to connect to this specific peer"
 
And I agree that the abstraction is not quite right :) I'll see what I can do about it (perhaps a separate method/callback that PeerManager can use directly).

core/src/main/java/com/radixdlt/p2p/addressbook/AddressBook.java
```diff
@@ -264,44 +272,38 @@ private boolean isPeerIpAddressValid(RadixNodeUri uri) {
...
+          continue;
+        }
+        final var updatedEntry =
+            entry
+                .getValue()
+                .removeAddressesThatMatchHostAndPort(uriToKeep.getHost(), uriToKeep.getPort());
+        if (entry.getValue().getKnownAddresses().size()
+            != updatedEntry.getKnownAddresses().size()) {
+          // Store a new entry if any addresses were filtered out
+          upsertOrRemoveIfEmpty(updatedEntry);
```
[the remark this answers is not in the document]
LukasGasior1: Ha, nice catch!
It actually only worked thanks to using **Concurrent**HashMap.
Nonetheless, I've updated it to use an iterator (and `iter.remove()` + `entry.setValue(...)`)

core/src/main/java/com/radixdlt/p2p/addressbook/AddressBook.java
LukasGasior1: I think here we could actually rename it to `isInvalid` / `markAsInvalid`.

LukasGasior1: or remove it completely :thinking: we're only blocking the address if handshake fails, so why not remove it immediately?

LukasGasior1: or if we're keeping it, we could name it even more explicitly: `handshakeFailed` or similar

LukasGasior1: I guess it makes _some_ sense to keep an address with this flag in the address book for some time (so that we don't retry if the address is received again in peer discover)

core/src/main/java/com/radixdlt/p2p/addressbook/AddressBook.java
LukasGasior1: Hmh, from address book perspective these aren't necessarily all validators, f.e. in the latest commit I also mark the peers we're syncing with as high priority. Separately, validator set size isn't really a constant

core/src/main/java/com/radixdlt/p2p/addressbook/BerkeleyAddressBookStore.java
```diff
@@ -119,16 +116,14 @@ public BerkeleyAddressBookStore(
...
-          this.dbEnv.getEnvironment().openDatabase(null, "address_book_entries", config);
+      this.entriesDb = this.dbEnv.openDatabase(null, "address_book_entries", config);
       this.highPriorityPeersDb =
-          this.dbEnv
-              .getEnvironment()
-              .openDatabase(null, "address_book_high_priority_peers", config);
+          this.dbEnv.openDatabase(null, "address_book_high_priority_peers", config);
     } catch (DatabaseException | IllegalArgumentException | IllegalStateException ex) {
-      throw new IllegalStateException("while opening database", ex);
+      throw new IllegalStateException("while opening databa se", ex);
```
dhedey, line 126: Very minor typo

dhedey (approved):
Nice - much clearer now :+1:

PR #515, opened 2023-06-22
Previous value in Substate Update

jakrawcz-rdx:
As promised in https://github.com/radixdlt/babylon-node/pull/511#discussion_r1235037977

This PR:
- starts to use the previous value that was read from DB anyway during `substate_changes` construction
- propagates this to Core API (where `CreatedOrUpdatedSubstate` gets split into `CreatedSubstate` vs `UpdatedSubstate`)
   - the `previous` part of `UpdatedSubstate` must be requested in the request flags
- uses it in the parent_map construction (to avoid unnecessary record overwrites)


malexandru-rdx (approved):
LGTM

dhedey (commented):
Super nice :D

PR #517, opened 2023-06-22
Don't verify stale sync responses

LukasGasior1:


core/src/main/java/com/radixdlt/sync/LocalSyncService.java
```diff
@@ -430,10 +430,20 @@ private SyncState processSyncResponse(
     log.trace("LocalSync: Received sync response from {}", sender);
 
     if (!currentState.waitingForResponseFrom(sender)) {
-      log.warn("LocalSync: Received unexpected sync response from {}", sender);
+      log.trace("LocalSync: Received unexpected sync response from {}", sender);
+      metrics.sync().unexpectedSyncResponses().inc();
```
jakrawcz-rdx, line 434: _(if you have time for such nitpicking)_ If the metric had a `sender` label, then the log would not be needed at all.

core/src/main/java/com/radixdlt/sync/LocalSyncService.java
```diff
@@ -430,10 +430,20 @@ private SyncState processSyncResponse(
...
       return currentState;
     }
 
+    final var responseTailStateVersion =
+        syncResponse.getLedgerExtension().getTail().getLedgerHeader().getStateVersion();
+    final var currentStateVersion = currentState.getCurrentHeader().getStateVersion();
+    if (responseTailStateVersion <= currentStateVersion) {
+      log.trace("LocalSync: Received a stale sync response from {}", sender);
+      metrics.sync().staleSyncResponses().inc();
```
jakrawcz-rdx, line 443: _(if you have time for such nitpicking)_ If the metric had a `sender` label, then the log would not be needed at all.

jakrawcz-rdx (approved):

jakrawcz-rdx:
If I understood @dhedey correctly, he suggested to have a condition like `if response.epoch != current_epoch ...`.
Comparing versions instead seems like additional optimization (i.e. filtering out more definitely useless responses).

However, I wonder if we may still see similar issues, but caused by receiving response from _next_ epoch?
_(if that's an unrealistic risk, then comparing versions is perfect)_

core/src/main/java/com/radixdlt/sync/LocalSyncService.java
malexandru-rdx: I would keep the log either way.

core/src/main/java/com/radixdlt/sync/LocalSyncService.java
LukasGasior1: hmh, I'd prefer not to have it as a label, theoretically there isn't any upper limit of peers the node can talk to during its run lifetime

LukasGasior1:
> However, I wonder if we may still see similar issues, but caused by receiving response from next epoch?
(if that's an unrealistic risk, then comparing versions is perfect)

I don't think it's possible. The request contains the start state version and if there's an epoch boundary in the response it must always be the tail header.
Theoretically an adversary could still send us an unsolicited "future" response, but I guess in this case it makes sense to verify it and ban that peer, if needed.

PR #518, opened 2023-06-22
Set Olympia end state http client timeout to 10 minutes

LukasGasior1:


core/src/main/java/com/radixdlt/genesis/olympia/OlympiaGenesisService.java
```diff
@@ -140,7 +140,7 @@ private void poll(CompletableFuture<GenesisData> completableFuture, int counter)
     final var includeTestPayload = counter % 1000 == 1;
     if (counter == 0) {
       log.info(
-          "Querying the Olympia node {} for genesis data{}",
+          "Querying the Olympia node {} for genesis data{} (this may take a few minutes)",
```
dhedey, line 143: Just noticed coincidentally - `counter == 0` and `includeTestPayload` will not be true?

Perhaps we should log if `counter == 0 || includeTestPayload`?

dhedey (approved):

PR #520, opened 2023-06-22
Using parent map for accurate computation of balance changes in LTS endpoints

jakrawcz-rdx:
Finally, a feature enabled by https://github.com/radixdlt/babylon-node/pull/511.

The uncommented asserts in the test started passing immediately after this impl, which is a bit suspicious.

### DIFF
--- a/core-rust/core-api-server/src/core_api/conversions/lts.rs
+++ b/core-rust/core-api-server/src/core_api/conversions/lts.rs
@@ -2,16 +2,20 @@ use radix_engine::{
     transaction::BalanceChange,
     types::{Decimal, GlobalAddress, IndexMap, ResourceAddress, RADIX_TOKEN},
 };
+use state_manager::store::traits::SubstateNodeAncestryStore;
+use state_manager::store::StateManagerDatabase;
 use state_manager::{
     CommittedTransactionIdentifiers, LedgerTransactionOutcome, LocalTransactionReceipt,
     StateVersion, SubstateChange, TransactionTreeHash,
 };
+use std::ops::SubAssign;
 use transaction::prelude::*;
 
 use crate::core_api::*;
 
 #[tracing::instrument(skip_all)]
 pub fn to_api_lts_committed_transaction_outcome(
+    database: &StateManagerDatabase,
     context: &MappingContext,
     state_version: StateVersion,
     receipt: LocalTransactionReceipt,
@@ -40,8 +44,9 @@ pub fn to_api_lts_committed_transaction_outcome(
         }),
         status,
         fungible_entity_balance_changes: to_api_lts_fungible_balance_changes(
+            database,
             context,
-            total_fee,
+            &receipt.local_execution.fee_payments,
             &receipt.local_execution.state_update_summary.balance_changes,
         )?,
         resultant_account_fungible_balances: to_api_lts_resultant_account_fungible_balances(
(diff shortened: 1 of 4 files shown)


core-rust/core-api-server/src/core_api/conversions/lts.rs
```diff
@@ -54,117 +59,59 @@ pub fn to_api_lts_committed_transaction_outcome(
...
-                if biggest_xrd_debit.is_none() || balance_change < biggest_xrd_debit.unwrap().0 {
-                    biggest_xrd_debit = Some((balance_change, *entity_address));
-                }
-                net_xrd_balance_change += balance_change;
-            }
-        }
+    let mut fee_balance_changes = index_map_new();
+    let records = database.batch_get_ancestry(fee_payments.keys());
+    for (paid_fee_amount_xrd, ancestry) in fee_payments.values().zip(records) {
+        let ancestry = ancestry.expect("a vault must be owned by an account");
```
dhedey, line 70: We typically use `MappingError` instead of panics here, so the error gets turned into a 500. (Because a Core API issue/invariant is imo not sufficiently fatal to take down a node).

dhedey: Also, the expect message is incorrect - "a vault must be owned" - but not necessarily by an account! It could be some other global entity.

core-rust/core-api-server/src/core_api/conversions/lts.rs
```diff
@@ -54,117 +59,59 @@ pub fn to_api_lts_committed_transaction_outcome(
...
+        let account_address = GlobalAddress::new_or_panic(ancestry.root.0 .0);
```
dhedey, line 71: Again, not necessarily an account, should probably say `global_ancestor_address`

core-rust/core-api-server/src/core_api/conversions/lts.rs
```diff
@@ -54,117 +59,59 @@ pub fn to_api_lts_committed_transaction_outcome(
...
-                            if *resource_address == RADIX_TOKEN {
-                                let fungible_balance_change =
-                                    get_fungible_balance(balance_change).unwrap();
-                                let non_fee_balance_change =
-                                    fungible_balance_change - assumed_fee_balance_change;
-                                if non_fee_balance_change == Decimal::ZERO {
-                                    return None;
-                                }
-                                return Some(to_api_lts_fungible_resource_balance_change(
+    Ok(balance_changes
```
dhedey, line 78: I don't believe this is quite right. If a fee_balance_change is offset by an equal and opposite transfer in, then the balance maybe won't be in the `balance_changes`.

I would suggest the simplest is to make `non_fee_balance_changes = balance_changes` and add the `paid_fee_amount_xrd` XRD fees to the `non_fee_balance_changes` alongside line 72.

The logic below then also becomes a lot simpler - basically just filtering out 0s and mapping.

dhedey (changes requested):
Nice! Glad the tests passed so simply. A few suggestions for tweaks and a minor edge case that I think could do with fixing up (but I might have misunderstood something!)

core-rust/core-api-server/src/core_api/conversions/lts.rs
dhedey: So I had to refactor all of this as part of pulling through damson/elm, so I suggest probably wanting to close this PR and put these changes on top of the elm PR

jakrawcz-rdx: ok, let me do it 👍 (I assume "the elm PR" == https://github.com/radixdlt/babylon-node/pull/529)

PR #523, opened 2023-06-25
Strict vertex limits

malexandru-rdx:
Vertex prepare stops after first limit exceeded.
Adds VertexPrepareMetrics which tracks proposal size, wasted bandwidth and why prepare stopped (vertex limits).
Sneaks up a fix in limits :).


core-rust/state-manager/src/limits.rs
```diff
@@ -162,16 +175,24 @@ impl VertexLimitsTracker {
...
+        Ok(self.check_filled())
+    }
+}
+
+pub trait HasUnit {
+    fn unit() -> Self;
+}
+
+impl HasUnit for ExecutionMetrics {
+    fn unit() -> Self {
```
dhedey, line 187: I feel like `unit` is a bit of a weird concept. It's usually used for the multiplicative identity - but that doesn't make sense in this context.

Really you're after "minimum_for_transaction" - in which case eg `max_wasm_memory_used` is 0 because a transaction doesn't need that.

But in any case, I'm not sure if this is a particularly useful check in its current form? :)

core-rust/state-manager/src/metrics.rs
```diff
@@ -211,6 +218,31 @@ impl CommittedTransactionsMetrics {
     }
 }
 
+impl VertexPrepareMetrics {
+    pub fn new(registry: &Registry) -> Self {
+        Self {
+            proposal_transactions_size: IntGauge::with_opts(opts(
```
dhedey, line 224: These feel a little weird to be gauges - I imagine you want to use a (basic) Summary instead, or a Histogram?

core-rust/state-manager/src/metrics.rs
```diff
@@ -408,3 +440,30 @@ impl MetricLabel for MempoolAddError {
         }
     }
 }
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum VertexPrepareStopReason {
+    Normal,
```
dhedey, line 446: What does Normal mean? All proposal transactions fitted into the vertex?

dhedey: Maybe `ProposalComplete` might make sense instead of `Normal`?

core-rust/state-manager/src/metrics.rs
```diff
@@ -408,3 +440,30 @@ impl MetricLabel for MempoolAddError {
...
+    type StringReturnType = &'static str;
+
+    fn prometheus_label_name(&self) -> Self::StringReturnType {
+        match self {
+            VertexPrepareStopReason::Normal => "Normal",
+            VertexPrepareStopReason::EpochChange => "EpochChange",
+            VertexPrepareStopReason::LimitExceeded(limit_exceeded) => match limit_exceeded {
+                VertexLimitsExceeded::TransactionsCount => "TransactionsCountExceeded",
+                VertexLimitsExceeded::TransactionsSize => "TransactionsSizeExceeded",
+                VertexLimitsExceeded::ExecutionCostUnitsConsumed => "ExecutionCostUnitsConsumed",
```
dhedey, line 461: Just a note - the suffix here is inconsistent with the rest

core-rust/state-manager/src/metrics.rs
```diff
@@ -408,3 +440,30 @@ impl MetricLabel for MempoolAddError {
...
+                VertexLimitsExceeded::SubstateReadSize => "SubstateReadSizeExceeded",
+                VertexLimitsExceeded::SubstateReadCount => "SubstateReadCountExceeded",
+                VertexLimitsExceeded::SubstateWriteSize => "SubstateWriteSizeExceeded",
+                VertexLimitsExceeded::SubstateWriteCount => "SubstateWriteCountExceeded",
```
dhedey, line 465: On reflection, I'd suggest we phrase these as `_LimitReached` instead of `_Exceeded` - as "Exceeded" sounds like an error to me :)

dhedey (commented):
Nice stuff. Could I request this is rebased / updated to merge into #529 if we want to merge it  for RCnet-V2? :)

core-rust/state-manager/src/metrics.rs
jakrawcz-rdx: I'd go for a single `wasted_proposal_space_size: Histogram` (I am assuming we already have other metrics for count and size distribution of committed transactions).

The gauge would give you a an exact value of a _very_ rare sample - I really believe that knowing an accurate-enough distribution of all samples is better.

jakrawcz-rdx (commented):
Very cool, I love the `StopReason` metric.
Just some minor suggestions.

core-rust/state-manager/src/metrics.rs
malexandru-rdx: We do have for size and count of committed **transactions**. This captures proposals. I'm interested in both how big (or small) the proposals are, as well as the wasted bandwidth.

A Histogram indeed makes more sense.

dhedey: The point is that a Gauge metric isn't useful in this context, as it gives a snapshot of the last quantity when the metric was read, but really you want some kind of summary of things, for two reasons:
* In case multiple things happened between reads
* So you can sensibly aggregate distributions over time

core-rust/state-manager/src/metrics.rs
```diff
@@ -134,62 +145,118 @@ impl CommittedTransactionsMetrics {
...
             )
             .registered_at(registry),
             substate_read_size: new_histogram(
                 opts(
                     "committed_transactions_substate_read_size",
                     "Total (per committed transaction) substate read size in bytes.",
                 ),
-                // TODO(RCnet-V3): update once max substate reads can be limited at execution
-                Self::buckets(DEFAULT_MAX_TOTAL_VERTEX_SUBSTATE_READ_SIZE),
+                // TODO: update once max substate reads can be limited at execution
```
dhedey, line 166: Why have you removed the RCnet-V3 from these comments?

dhedey: In general I'd like the TODOs to capture when we believe they need to be done for

core-rust/state-manager/src/metrics.rs
```diff
@@ -258,6 +325,26 @@ fn equidistant_buckets(number_of_buckets: usize, min: f64, max: f64) -> Vec<f64>
         .collect()
 }
 
+// Given a limit, builds buckets for a Histogram with higher resolution for higher values.
```
dhedey, line 328: Very nice, appreciate the comments! :)

dhedey (approved):
Nice stuff, thanks! :)

core-rust/state-manager/src/metrics.rs
malexandru-rdx: Not sure, probably this was `git merge`. I'll change back and merge.

PR #528, opened 2023-06-27
Update/For Elm

dhedey:
WIP


dhedey:
Closing in favour of https://github.com/radixdlt/babylon-node/pull/529 because it turns out there was some ref called `update` which is breaking git clones

PR #529, opened 2023-06-27
Feature/update for rcnet-v2

dhedey:
The major update PR for `rcnet-v2` - note that there will be a few small updates to Core API and scrypto after we merge this and before release, but I want to get this merged into develop now to decrease merge conflicts


cli-tools/src/integration/java/com/radixdlt/integration/targeted/capabilities/NodeCapabilityTests.java
```diff
@@ -187,7 +187,7 @@ public void ensure_messages_for_an_enabled_capability_are_processed() {
           m -> messagesReceived.add("node1-" + m.message().getClass().getSimpleName()));
 
       // Have the expected messages been received?
-      var result = waitForMessagesReceived(expectedResultMap, messagesReceived, 3);
+      var result = waitForMessagesReceived(expectedResultMap, messagesReceived, 10);
```
dhedey, line 190: This test flaked out on me, probably because genesis is a little slower, so I increased the timeout

core-rust/state-manager/src/transaction/series_execution.rs
```diff
@@ -123,26 +123,32 @@ where
...
             self.state_tracker.state_version,
             &self.state_tracker.ledger_hashes.transaction_root,
-            &transaction.ledger_transaction_hash(),
-            self.execution_configurator
-                .wrap(transaction.get_executable(), config_type)
-                .warn_after(
-                    config_type.get_transaction_runtime_warn_threshold(),
-                    &description,
-                ),
+            &description.ledger_hash,
```
jakrawcz-rdx, line 150: hmm, this argument drives logic, and I tremble in fear when I see that we take its value from a log-only error-surfacing util struct O_O

is there a reason to change this class in this PR anyways?

core-rust/state-manager/src/types.rs
```diff
@@ -88,6 +94,12 @@ impl IsMerklizableHash for SubstateChangeHash {}
 
 define_wrapped_hash!(EventHash);
 
+impl Display for EventHash {
```
jakrawcz-rdx, line 97: in the name of boilerplate reduction, would it be easy enough to make this impl a part of `define_wrapped_hash!()`?

jakrawcz-rdx (approved):
I admire the enormous effort 👍 
I reviewed a couple of files which I was familiar with, and I only have optional remarks (that can all be deferred for later really).

core-rust/state-manager/src/transaction/series_execution.rs
dhedey: Ahh this is a side effect of a refactor to support the Flash transactions (before I changed approach and decided we should just map them into a dummy receipt) - which I then mostly reverted, but this remained!

I didn't realise DescribedTransactionHash was intended just for logging hah - but this is private to this class so I'm not particularly worried :).

I can rename this parameter to `described_hash` if that makes it less scary? 👍

core-rust/state-manager/src/transaction/executable_logic.rs
```diff
@@ -91,81 +93,111 @@ impl ExecutionConfigurator {
...
-        let result = self.underlying.execute_on(store);
-        let elapsed = start.elapsed();
-        if elapsed > self.threshold {
-            warn!(
-                "Execution of {} took {}ms, above warning threshold of {}ms",
-                self.description,
-                elapsed.as_millis(),
-                self.threshold.as_millis(),
-            );
+        match self {
```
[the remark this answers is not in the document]
dhedey: I appreciate this is less composable, but I think that's right actually. I think it is much easier to check/verify correctness having this logic together. And this file is very clearly about "how do I run something against the engine" which is a nice encapsulation.

It was hard to integrate the flash before this refactor because what is 3 lines in the engine is strewn across about 10 files - whereas now we can verify in one place it's configured right :)

The flash is the first transaction on ledger - which basically tells the engine to write substates without executing a transaction - this is to prevent a chicken-and-egg problem with some of the package/system related state that needs to be stored.

core-rust/state-manager/src/transaction/series_execution.rs
jakrawcz-rdx: > I can rename this parameter

Either this (accompanied by update to the rustdoc of DescribedTransactionHash)...
...OR - if the changes to this file really turned out to not be needed - then I'd be even more happy if you simply reverted the file.

PR #532, opened 2023-06-29
Mempool Improvement part 1

malexandru-rdx:
Adds mempool total transactions limit (config field `mempool.maxTotalTransactionsSize`) along transaction count limit (config field `mempool.maxTransactionCount`).
Adds `proposal_priority_index` which is always used to build new proposals as specified in [RNP-10](https://radixdlt.atlassian.net/wiki/spaces/RPNV1/pages/edit-v2/3067904346?draftShareId=91251dd4-1b25-4e74-a608-54148ad1cded).

### CoreAPI changes:
This changes removes the `MempoolFull` in favour of `MempoolPriorityThresholdNotMet`


core-rust/state-manager/src/mempool/simple_mempool.rs
```diff
@@ -120,39 +133,139 @@ impl MempoolData {
     }
 }
 
+/// A wrapper around an [`Arc<MempoolData>`] which implements ordering traits for the proposal priority.
+/// The order is such that a normal iteration over a BTreeSet gives the next best proposal transactions,
+/// i.e. if a < b then a is prioritized over b.
+#[derive(Clone, Eq)]
```
dhedey, line 139: I'm quite scared about the performance of the `Eq` operator here, particularly if it's used `O(N log N)` times - can we derive a more explicit `Eq` on `MempoolData` or `MempoolTransaction` to use the transaction hash?

dhedey (commented):
Really nice work. Let me know when the markups are done and happy to take a re-review.

You may also want to merge the elm branch into this to avoid merge conflicts :)

core-rust/state-manager/src/mempool/mempool_manager.rs
```diff
@@ -327,13 +376,18 @@ impl MempoolManager {
...
-                write_mempool.remove_transaction(intent_hash, user_payload_hash)
+            .filter_map(|(_intent_hash, user_payload_hash)| {
+                write_mempool.remove_by_payload_hash(user_payload_hash)
             })
-            .count();
-        self.metrics.current_transactions.sub(removed_count as i64);
+            .fold(TransactionSizeAccumulator::new(), |acc, evicted| {
+                acc.accumulate(&evicted.transaction)
+            });
+        self.metrics.current_transactions.sub(removed.count as i64);
```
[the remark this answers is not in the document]
malexandru-rdx: Yeah, it felt weird using them in the MempoolManager. I will split (if needed) and move them where it's more fit to hold responsibility.

core-rust/state-manager/src/mempool/mod.rs
```diff
@@ -133,21 +135,13 @@ impl MempoolAddRejection {
 impl ToString for MempoolAddError {
     fn to_string(&self) -> String {
         match self {
-            MempoolAddError::Full {
-                current_size,
-                max_size,
-            } => format!("Mempool Full [{current_size} - {max_size}]"),
+            MempoolAddError::PriorityThresholdNotMet {min_tip_percentage_required, tip_percentage} => format!("Priority Threshold not met: tip is {tip_percentage} while min tip required {min_tip_percentage_required}"),
```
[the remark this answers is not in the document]
malexandru-rdx: Yep. It is `fmt`-approved.

core-rust/state-manager/src/mempool/simple_mempool.rs
```diff
@@ -120,39 +133,139 @@ impl MempoolData {
     }
 }
 
+/// A wrapper around an [`Arc<MempoolData>`] which implements ordering traits for the proposal priority.
+/// The order is such that a normal iteration over a BTreeSet gives the next best proposal transactions,
+/// i.e. if a < b then a is prioritized over b.
```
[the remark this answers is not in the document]
malexandru-rdx: Alright, I'm chaging it. I started by "I want to iterate from transaction in next proposal" - which resulted in this :). It is indeed more clear to have the `Ord` the right way.

core-rust/state-manager/src/mempool/simple_mempool.rs
malexandru-rdx: Thanks for catching this. I wanted to do it but forgot :(

core-rust/state-manager/src/mempool/simple_mempool.rs
```diff
@@ -120,39 +133,139 @@ impl MempoolData {
...
+pub struct MempoolDataProposalPriorityOrdering(pub Arc<MempoolData>);
+
+impl Ord for MempoolDataProposalPriorityOrdering {
+    fn cmp(&self, other: &Self) -> Ordering {
+        match self
+            .0
+            .transaction
+            .tip_percentage()
+            .cmp(&other.0.transaction.tip_percentage())
+        {
```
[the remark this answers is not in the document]
malexandru-rdx: Nice. Thanks :D

core-rust/state-manager/src/mempool/mempool_manager.rs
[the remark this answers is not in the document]
malexandru-rdx: But I would do this in a next PR where I also move/refactor metrics code.

core-rust/state-manager/src/mempool/simple_mempool.rs
```diff
@@ -120,39 +133,139 @@ impl MempoolData {
...
+    fn eq(&self, other: &Self) -> bool {
+        self.0.transaction.notarized_transaction_hash()
+            == other.0.transaction.notarized_transaction_hash()
+    }
+}
+
 pub struct SimpleMempool {
-    max_size: u64,
-    data: HashMap<NotarizedTransactionHash, MempoolData>,
+    remaining_transaction_count: usize,
```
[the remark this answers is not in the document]
malexandru-rdx: Will do in next batch of improvements. 😄

core-rust/state-manager/src/mempool/simple_mempool.rs
```diff
@@ -120,39 +133,139 @@ impl MempoolData {
...
-    ) -> Result<(), MempoolAddError> {
-        self.check_if_mempool_full()?;
-
+    ) -> Result<Vec<Arc<MempoolData>>, MempoolAddError> {
         let payload_hash = transaction.notarized_transaction_hash();
         let intent_hash = transaction.intent_hash();
+        let transaction_size = transaction.raw.0.len();
+
+        let target_remaining_total_transactions_size = transaction_size;
+        let target_remaining_transaction_count = 1;
```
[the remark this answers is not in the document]
malexandru-rdx: I've removed it. Bulk add algorithm requires more steps to make it worth more than iterate and add.

core-rust/state-manager/src/mempool/mempool_manager.rs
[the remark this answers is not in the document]
dhedey: Sounds good

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -4543,11 +4543,16 @@ components:
...
+              description: |
+                Tip percentage of the submitted (and rejected) transaction.
             min_tip_percentage_required:
               type: integer
               format: int32
               minimum: 0
               maximum: 65535
+              description: |
+                A lower bound for tip percentage at current mempool state. Anything lower than this will very likely result in a mempool rejection.
+                A value of 0 means there is no tip that can guarantee submission.
```
dhedey, line 4555: A couple of things:
* I don't think we need to mention the value of 0 / fallback
* When does this occur? If all the mempool is full of max tip transactions? (or the mempool has unrealistic limits) I think we should use 65535 in that case?

core-rust/state-manager/src/mempool/simple_mempool.rs
```diff
@@ -196,52 +187,57 @@ impl SimpleMempool {
...
-            || remaining_transaction_count < target_remaining_transaction_count
+        while total_transaction_size_free_space < transaction_size
+            || total_transaction_count_free_space < 1
         {
-            let order_data = priority_iter.next_back();
-            match order_data {
+            let lowest_priority_transaction = priority_iter.next();
+            match lowest_priority_transaction {
                 None => {
+                    // Even with an empty mempool we are not able to fulfill the request.
```
dhedey, line 222: Hmm I feel we should output something like this in this case, rather than panicking? Some of our tests configure a zero sized mempool, so would be a bit weird to panic.
```
    return Err(MempoolAddError::PriorityThresholdNotMet {
        min_tip_percentage_required: None,
        tip_percentage: transaction_data.transaction.tip_percentage(),
    });
```

core-rust/state-manager/src/mempool/simple_mempool.rs
```diff
@@ -196,52 +187,57 @@ impl SimpleMempool {
...
         }
 
+        // Check the new transaction is better than all to be removed transactions.
         if !to_be_removed.is_empty() {
             let best_to_be_removed = to_be_removed.last().unwrap();
             if new_order_data
-                > MempoolDataProposalPriorityOrdering((*to_be_removed.last().unwrap()).clone())
+                < MempoolDataProposalPriorityOrdering((*to_be_removed.last().unwrap()).clone())
             {
+                // Note: overflow back to 0 is the desired outcome. This means there is no minimum tip guaranteed to add the transaction.
```
dhedey, line 240: Thanks for adding this comment, it draws attention to something I missed :).

But just to say - we should never IMO rely on overflow behaviour - as it panics in debug mode. I would much rather be explicit with eg `wrapping_add`.

Also - I disagree with the premise of using `0` for such a value anyway. We could use an `Option<u16>` where it not being present means there is no valid tip (or even a custom enum) - Rust gives us tools to avoid these magic values, so I feel it's good practice to use them.

core-rust/state-manager/src/mempool/simple_mempool.rs
```diff
@@ -261,7 +258,7 @@ impl SimpleMempool {
 
         // Add new MempoolData
         if self.data.insert(payload_hash, transaction_data).is_some() {
-            panic!("Transaction already inside mempool");
+            panic!("Broken precondition: Transaction already inside mempool");
```
dhedey, line 261: Thanks, this is clearer 👍

core-rust/state-manager/src/mempool/simple_mempool.rs
```diff
@@ -388,6 +385,39 @@ impl SimpleMempool {
...
+    /// Transactions are picked in the order of [`proposal_priority_index`].
+    /// Obeys the given count/size limits and explicit exclusions.
+    pub fn get_proposal_transactions(
+        &self,
+        max_count: usize,
+        max_payload_size_bytes: u64,
+        user_payload_hashes_to_exclude: &HashSet<NotarizedTransactionHash>,
+    ) -> Vec<Arc<MempoolTransaction>> {
+        const MAX_TRANSACTION_TO_TRY: usize = 1000;
+        let max_transaction_to_try = max(max_count, MAX_TRANSACTION_TO_TRY);
```
dhedey, line 399: `max_transactions*_...`

dhedey (changes requested):
Looking much better, thanks! A couple of minor tweaks, then happy for us to merge :).

BTW - we also have a relevant failure:
```
com.radixdlt.integration.targeted.rev2.mempool.REv2MempoolFillAndEmptyTest check_that_full_mempool_empties_itself FAILED (11.3s)

  com.radixdlt.harness.deterministic.DeterministicNodes$EventHandleException: Exception: com.radixdlt.sbor.exceptions.SborDecodeException: Decoded Type ID 9 (TYPE_U32) but expected 4 (TYPE_I32
```

core-rust/state-manager/src/mempool/simple_mempool.rs
```diff
@@ -220,7 +221,11 @@ impl SimpleMempool {
             match lowest_priority_transaction {
                 None => {
                     // Even with an empty mempool we are not able to fulfill the request.
-                    panic!("Impossible to add new transaction. Mempool max size lower than transaction size!");
+                    warn!("Impossible to add new transaction. Mempool max size lower than transaction size!");
```
dhedey, line 224: Nice, thanks for the warn :)

core/src/integration/java/com/radixdlt/integration/targeted/rev2/mempool/REv2MempoolFillAndEmptyTest.java
```diff
@@ -140,7 +140,7 @@ private void fillAndEmptyMempool(DeterministicTest test) {
     var mempoolDispatcher =
         test.getInstance(0, Key.get(new TypeLiteral<EventDispatcher<MempoolAdd>>() {}));
 
-    while (mempoolReader.getCount() < 1000) {
+    while (mempoolReader.getCount() < 100) {
```
dhedey, line 143: This is possibly a less useful test now - could we maybe explicitly set the mempool larger? Happy for you to do this in a follow up PR 👍

dhedey (approved):
Really nice thanks, happy for you to merge

PR #534, opened 2023-07-04
Add option to disable fees; Use correct ledger header for BFTSync; va…

LukasGasior1:
…rious improvements around MultiNodeRebootTest.

Included in this PR:

- add an option to disable fees (and, as a consequence, validator rewards) in tests, e.g. if we want a fixed-stake validator set
- use the latest ledger header when instantiating BFTSync (see comment in EpochManager)
- remove `runOnThreads` from BFTSync - it was a temporary debug info; no longer needed
- update ledger sync request timeout from 5000 ms to 100 in MultiNodeRebootTest - this tests relies on this in some cases (and 5s was enough wait time to trigger a failure)
- don't dispatch BFTHighQCUpdate if we're dispatching BFTCommittedUpdate in VertexStore

### DIFF
--- a/core-rust/state-manager/src/jni/state_manager.rs
+++ b/core-rust/state-manager/src/jni/state_manager.rs
@@ -75,6 +75,7 @@ use node_common::environment::setup_tracing;
 use node_common::java::*;
 use parking_lot::RwLock;
 use prometheus::{Encoder, Registry, TextEncoder};
+use radix_engine::transaction::FeeReserveConfig;
 use radix_engine_interface::network::NetworkDefinition;
 use radix_engine_interface::*;
 use tokio::runtime::Runtime;
@@ -131,6 +132,7 @@ pub struct StateManagerConfig {
     pub database_backend_config: DatabaseBackendConfig,
     pub database_flags: DatabaseFlags,
     pub logging_config: LoggingConfig,
+    pub no_fees: bool,
 }
 
 pub type ActualStateManager = StateManager<StateManagerDatabase>;
@@ -177,7 +179,14 @@ impl JNIStateManager {
             ),
         ));
         let metric_registry = Registry::new();
-        let execution_configurator = Arc::new(ExecutionConfigurator::new(&logging_config));
+        let mut fee_reserve_config = FeeReserveConfig::default();
+        if config.no_fees {
+            fee_reserve_config.cost_unit_price = 0;
+        }
+        let execution_configurator = Arc::new(ExecutionConfigurator::new(
+            &logging_config,
+            fee_reserve_config,
+        ));
         let pending_transaction_result_cache = Arc::new(parking_lot::const_rwlock(
             PendingTransactionResultCache::new(10000, 10000),
         ));
--- a/core/src/integration/java/com/radixdlt/integration/steady_state/deterministic/rev2/consensus_ledger_sync/MultiNodeRebootTest.java
+++ b/core/src/integration/java/com/radixdlt/integration/steady_state/deterministic/rev2/consensus_ledger_sync/MultiNodeRebootTest.java
@@ -86,11 +86,11 @@ import com.radixdlt.rev2.Decimal;
 import com.radixdlt.rev2.REV2TransactionGenerator;
 import com.radixdlt.rev2.modules.MockedVertexStoreModule;
 import com.radixdlt.rev2.modules.REv2StateManagerModule;
+import com.radixdlt.statemanager.DatabaseFlags;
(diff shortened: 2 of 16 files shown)


core-rust/state-manager/src/jni/state_manager.rs
```diff
@@ -131,6 +132,7 @@ pub struct StateManagerConfig {
     pub database_backend_config: DatabaseBackendConfig,
     pub database_flags: DatabaseFlags,
     pub logging_config: LoggingConfig,
+    pub no_fees: bool,
```
dhedey, line 135: I'm not a massive fan of random booleans, but I guess this is okay for now until we have more configurations we might want to combine into a more meaningful struct.

core/src/integration/java/com/radixdlt/integration/steady_state/deterministic/rev2/consensus_ledger_sync/MultiNodeRebootTest.java
```diff
@@ -215,11 +214,15 @@ private DeterministicTest createTest(
...
+                            .totalEmissionXrdPerEpoch(Decimal.of(0))),
                     REv2StateManagerModule.DatabaseType.ROCKS_DB,
+                    new DatabaseFlags(true, false),
                     StateComputerConfig.REV2ProposerConfig.transactionGenerator(
-                        new REV2TransactionGenerator(), 1)),
-                SyncRelayConfig.of(5000, 10, 5000L))));
+                        new REV2TransactionGenerator(), 1),
+                    false,
+                    true),
+                SyncRelayConfig.of(100, 10, 500))));
```
dhedey, line 225: Can we have a comment explaining all of these parameters and how they're important?

I'm worried some of this (eg avoiding stake changing over time) is quite fragile, so any hints to future people regarding where to look if this becomes flaky in future I think would be useful.

core/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
```diff
@@ -171,10 +170,7 @@ public String toString() {
   private final int bftSyncPatienceMillis;
   private final Metrics metrics;
 
-  private LedgerProof currentLedgerHeader;
-
-  // TODO: remove once we figure that out
-  private final Set<String> runOnThreads = Collections.newSetFromMap(new ConcurrentHashMap<>(2));
```
dhedey, line 177: Can you explain a little more why we had this? In DM is fine

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreAdapter.java
```diff
@@ -150,24 +140,31 @@ public List<ExecutedVertex> getPathFromRoot(HashCode vertexHash) {
...
-                                  insertedQc.vertexStoreState())));
-            });
+    result.insertedQcs().forEach(this::dispatchPostQcInsertionEvents);
     result.insertUpdates().forEach(bftUpdateDispatcher::dispatch);
   }
 
+  private void dispatchPostQcInsertionEvents(VertexStore.InsertQcResult.Inserted inserted) {
+    // This is a bit of a hack.
+    // We don't want to persist the vertex store state (via PersistentVertexStore)
+    // if the state is already being persisted alongside commit.
```
dhedey, line 150: Would be useful if this comment explained why? Is this a performance improvement or does it have correctness implications?

dhedey: Does this fix a race condition in the tests? So it sounds like this is a correctness issue? Why though?

Is it that when we commit a vertex chain, we only want to dispatch the most recent vertex store state? Not the historic vertex store state? Perhaps we should make this even more explicit?

But hmm this doesn't even seem to protect against that though? If A, B and C are inserted, maybe they all have the same `committedUpdate`? It's been too long since I looked at this

dhedey (approved):
Nice! A few not super-important markups...

Could I ask you merge / rebase onto the elm PR, and then happy for you to merge into that when you're done.

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreAdapter.java
LukasGasior1: hmh, I think that was specifically needed around node restarts, e.g. when we manage to persist a vertex store state, but not the commit itself

LukasGasior1:
> Could I ask you merge / rebase onto the elm PR, and then happy for you to merge into that when you're done.

:+1: merged to elm

PR #535, opened 2023-07-04
Storing the results of executing "genesis scenarios" in the persistence

jakrawcz-rdx:
This is the first PR for the "expose Genesis Scenarios' results from Core API".
It only stores the results in the DB (and has no observable behavior yet - the API change is coming in another PR).

### DIFF
--- a/core-rust/state-manager/src/state_manager.rs
+++ b/core-rust/state-manager/src/state_manager.rs
@@ -764,9 +764,10 @@ where
         if !scenarios_to_run.is_empty() {
             info!("Running {} scenarios", scenarios_to_run.len());
             let mut next_nonce: u32 = 0;
-            for scenario_name in scenarios_to_run.iter() {
+            for (sequence_number, scenario_name) in scenarios_to_run.iter().enumerate() {
                 next_nonce = self.execute_genesis_scenario(
                     &mut genesis_commit_request_factory,
+                    sequence_number,
                     scenario_name.as_str(),
                     initial_epoch,
                     next_nonce,
@@ -793,6 +794,7 @@ where
     fn execute_genesis_scenario(
         &self,
         genesis_commit_request_factory: &mut GenesisCommitRequestFactory,
+        sequence_number: usize,
         scenario_name: &str,
         epoch: Epoch,
         nonce: u32,
@@ -805,7 +807,6 @@ where
                     scenario_name
                 )
             });
-        let scenario_base_version = genesis_commit_request_factory.state_version;
         let mut previous = None;
         let mut committed_transactions = Vec::new();
         info!("Running scenario: {}", scenario_name);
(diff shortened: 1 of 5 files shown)


dhedey (approved):
Really nice, thanks!

core-rust/state-manager/src/state_manager.rs
```diff
@@ -837,14 +790,93 @@ where
...
+    ) -> u32 {
+        let mut scenario = self
+            .find_scenario(epoch, nonce, scenario_name)
+            .unwrap_or_else(|| {
+                panic!(
+                    "Could not find scenario with logical name: {}",
+                    scenario_name
+                )
+            });
+        let scenario_base_version = genesis_commit_request_factory.state_version;
```
dhedey, line 808: I think this is a bit confusing that it is the state version _before_ the scenario executed.

I'd prefer to just set this to be:
```rust
    if let Some(scenario_start_state_version) = committed_transactions.first() {
        // We only commit a record of the scenario if it commits at least one transaction
        self.store.write().put(scenario_start_state_version, executed_scenario);
    }
```

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -133,12 +133,18 @@ enum RocksDBColumnFamily {
     /// Key: concat(account_address, state_version), value: null
     /// Given fast prefix iterator from RocksDB this emulates a Map<Account, Set<StateVersion>>
     AccountChangeStateVersions,
+    /// Additional details of "Scenarios" (and their transactions) executed as part of Genesis,
+    /// keyed by state version on top of which they are executed (note: this requires an assumption
```
dhedey, line 137: I'm not sure we need this note, or maybe we can say that we explicitly store Scenarios which committed transactions.

dhedey (commented):
Oh actually, just a quick thought

core-rust/state-manager/src/state_manager.rs
jakrawcz-rdx: TBH a "state version after the first transaction that got committed as part of the scenario" is also a bit arbitrary to me :(
And in both cases, "not storing information on attempted scenarios which didn't successfully commit transactions" is also a bit imperfect.

Sooo... why do we want to identify these scenarios by a _transaction's_ ID anyway? Can we do "scenario order number" instead? Like `0 -> radiswap{...}; 1 -> transfer_xrd{...}`? That solves the above problems, and the contained transactions are easily identifiable anyways, and AFAIU we only list all these scenario executions (fully) from the endpoint (i.e. no-one queries them by any ID). Do you see any downsides of "scenario order number"?

dhedey: Sure - that sounds good :)

PR #537, opened 2023-07-05
Exposing the Genesis Scenarios' results from a new Core API endpoint

jakrawcz-rdx:
⚠️ This builds on top of https://github.com/radixdlt/babylon-node/pull/529

This implements a use-case of https://github.com/radixdlt/babylon-node/pull/535.
_(a rather mundane use-case: listing those stored records)_


core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -4428,14 +4458,11 @@ components:
           description: |
             An explanation as to why the intent status is resolved as it is.
         committed_state_version:
-          type: integer
-          format: int64
-          minimum: 1
-          maximum: 100000000000000
+          $ref: "#/components/schemas/StateVersion"
```
jakrawcz-rdx, line 4461: _(note from author)_

This is just my drive-by refactor. (i.e. introduce reusable `StateVersion` openapi type)

core/src/test/java/com/radixdlt/api/core/NetworkConfigurationTest.java
```diff
@@ -67,32 +67,39 @@
...
-      assertThat(response.getNetworkHrpSuffix())
+      assertThat(configurationResponse.getNetworkHrpSuffix())
           .isEqualTo(DeterministicCoreApiTestBase.networkDefinition.hrp_suffix());
 
       // And check the package endpoint whilst we're here, using a well known address...
-      final var faucetPackageAddress = response.getWellKnownAddresses().getFaucetPackage();
+      final var faucetPackageAddress =
+          configurationResponse.getWellKnownAddresses().getFaucetPackage();
 
-      // TODO(wip): it's not so easy to assert on schema anymore; this only tests "no exception":
```
jakrawcz-rdx, line 90: _(note from author)_

This was my fault (a `TODO(wip)` comes from me and my internal rule says that such expression must not be pushed - maybe I should make it a git hook or something).
I now addressed it in the only way I saw viable (still lame). Suggestions (on "what to assert better here") welcome.

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -5208,6 +5203,79 @@ components:
...
+    DescribedAddress:
+      type: object
+      required:
+        - logical_name
+        - address
+      properties:
+        logical_name:
+          type: string
+        address:
+          $ref: '#/components/schemas/GlobalAddress'
```
dhedey, line 5278: Just a heads up that in the next PR, these will become more general addresses - ie include Vault addresses and even non fungible global ids - so perhaps we should just make this a string with a description.

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -5185,11 +5183,8 @@ components:
         - receipt_tree_hash
       properties:
         state_version:
-          type: integer
-          format: int64
-          minimum: 0
-          maximum: 100000000000000
-          description: An integer between `0` and `10^13`, representing the state version. The state version increments with each transaction, starting at `0` pre-genesis.
+          $ref: "#/components/schemas/StateVersion"
+          description: The state version, which increments with each transaction, starting at `0` pre-genesis.
```
dhedey, line 5187: So, _technically_ this is invalid in the Open API schema language - ie you can't override any properties (including an upstream description) here. (Can you tell I'm not a fan of the Open API schema design and ecosystem? :') )

If you notice - the description has now vanished from all the models.

_But_ - importantly - I believe ReDocly _does_ handle this correctly, so this _should_ be fine. Although I'm not 100% sure. It might be worth us double-checking when we deploy the schema file.

dhedey (approved):
Thanks!

core-rust/core-api-server/core-api-schema.yaml
jakrawcz-rdx: whooops, indeed o_O

Let's hope it renders the public-facing file fine - if not, ofc I will bring back the boilerplate 👍

core-rust/core-api-server/core-api-schema.yaml
jakrawcz-rdx: Alright, I migrated to a string with more vague description here 👍

PR #543, opened 2023-07-10
Mempool improvements part 2

malexandru-rdx:
This PR includes: 

- Remove the `get_all_transactions` in favour of `get_k_random_transactions` inside (formerly) `SimpleMempool` to be used by `MempoolManager`. 
- Rename `SimpleMempool` to `PriorityMempool`
- Fix a previously altered test

### DIFF
--- a/core-rust/state-manager/src/mempool/mempool_manager.rs
+++ b/core-rust/state-manager/src/mempool/mempool_manager.rs
@@ -143,6 +143,10 @@ impl MempoolManager {
         max_count: usize,
         max_payload_size_bytes: u64,
     ) -> Vec<Arc<MempoolTransaction>> {
+        // TODO: Definitely a better algorithm could be used here, especially with extra information like:
+        // which peer/peers are we sending this to? or what do we know about said peer to have in it's mempool?
+        // However (NOTE/WARN): changing transactions selection without careful consideration of the peer selection,
+        // can lead to a scenario where we keep sending same transactions to same peer.
         let candidate_transactions = self.mempool.read().get_k_random_transactions(max_count * 2);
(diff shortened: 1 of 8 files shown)


core-rust/state-manager/src/mempool/mempool_manager.rs
```diff
@@ -145,23 +143,36 @@ impl MempoolManager {
...
     }
 
     /// Checks the committability of a random subset of transactions and removes the rejected ones
     /// from the mempool.
     /// Obeys the given limit on the number of actually executed (i.e. not cached) transactions.
     pub fn reevaluate_transaction_committability(&self, max_reevaluated_count: u32) {
-        // TODO: instead of getting whole mempool use a priority queue/round robin.
-        let mut candidate_transactions = self.mempool.read().get_all_transactions();
+        const MIN_TRANSACTIONS_TO_CHECK_FOR_REEVALUATION: u32 = 100;
+        let candidate_transactions = self.mempool.read().get_k_random_transactions(max(
```
dhedey, line 168: It feels like we should actually re-evaluate the ones which have been in here the longest since their last re-evaluation?

dhedey (approved):
Looks good - what do you think about changing the re-evaluation to be done based on time since last re-evaluation?

Or maybe that's hard to do given that we store the mempool transactions separately to the cached results? And best left to a different PR?

core-rust/state-manager/src/mempool/mempool_manager.rs
malexandru-rdx: Adding this to my TODO list for a next PR. :)

PR #547, opened 2023-07-11
Cleanup BFTValidatorId: ValidatorAddress no longer nullable in consensus

LukasGasior1:
- extract SelfValidatorInfo (ValidatorAddress no longer optional/nullable in consensus) 
- more explicit split between epoched and non-epoched consensus modules (e.g. we no longer need an unused non-epoched Pacemaker, BFTSync, etc in the main configuration)
- minor cleanup


core/src/main/java/com/radixdlt/RadixNodeModule.java
```diff
@@ -173,18 +173,17 @@ protected void configure() {
...
     } else {
       // No validator address provided, and use genesis explicitly disabled
       OptionalBinder.newOptionalBinder(binder(), Key.get(ComponentAddress.class, Self.class));
-      install(new BFTValidatorIdModule());
+      install(new SelfValidatorInfoModule());
     }
 
     install(new PersistedBFTKeyModule());
     install(new CryptoModule());
-    install(new ConsensusModule());
```
dhedey, line 187: Yay! Confusing epochless consensus module finally removed :D

core/src/main/java/com/radixdlt/RunningRadixNode.java
```diff
@@ -135,8 +136,8 @@ public static RunningRadixNode run(UnstartedRadixNode unstartedRadixNode) {
     return new RunningRadixNode(injector);
   }
 
-  public BFTValidatorId self() {
-    return this.injector.getInstance(Key.get(BFTValidatorId.class, Self.class));
+  public SelfValidatorInfo self() {
+    return this.injector.getInstance(Key.get(SelfValidatorInfo.class, Self.class));
```
dhedey, line 140: We probably could/should remove the `Self.class` annotation on this? As the clue's in the name now hah

core/src/main/java/com/radixdlt/consensus/bft/BFTValidatorId.java
```diff
@@ -128,32 +115,31 @@ public static BFTValidatorId fromSerializedString(String str) {
...
-  public ComponentAddress getActiveValidatorAddress() {
-    return this.getValidatorAddress()
-        .orElseThrow(
-            () ->
-                new IllegalStateException(
-                    String.format(
-                        "validator (%s) from active validator set must have an address",
-                        this.key)));
+  /** Only for use in tests */
+  public static BFTValidatorId withKeyAndFakeDeterministicAddress(ECDSASecp256k1PublicKey key) {
```
dhedey, line 137: Nice :)

core/src/main/java/com/radixdlt/consensus/epoch/EpochsConsensusModule.java
```diff
@@ -96,19 +93,39 @@
...
     eventBinder.addBinding().toInstance(EpochRoundUpdate.class);
     eventBinder.addBinding().toInstance(EpochProposalRejected.class);
-    eventBinder.addBinding().toInstance(VertexRequestTimeout.class);
-    eventBinder.addBinding().toInstance(LedgerUpdate.class);
+    eventBinder.addBinding().toInstance(EpochLocalTimeoutOccurrence.class);
     eventBinder.addBinding().toInstance(Epoched.class);
   }
 
+  @Provides
+  @Singleton
```
dhedey, line 124: It feels really wrong that we need this in an EpochsConsensusModule (it feels like we should have one per epoch instead?) - could this be removed / fixed?

core/src/main/java/com/radixdlt/consensus/epoch/EpochsConsensusModule.java
```diff
@@ -417,9 +438,9 @@ private BFTSyncFactory bftSyncFactory(
       @BFTSyncPatienceMillis int bftSyncPatienceMillis,
       Metrics metrics,
       Hasher hasher) {
-    return (safetyRules, vertexStore, pacemakerState, currentLedgerHeader) ->
+    return (selfValidaatorId, safetyRules, vertexStore, pacemakerState, currentLedgerHeader) ->
```
dhedey, line 441: Typo on Validator

core/src/main/java/com/radixdlt/keys/SelfValidatorInfoFromGenesisModule.java
```diff
@@ -68,17 +68,17 @@
...
 import com.radixdlt.rev2.modules.REv2LedgerInitializerToken;
 import com.radixdlt.sync.TransactionsAndProofReader;
 
-public final class BFTValidatorIdFromGenesisModule extends AbstractModule {
+public final class SelfValidatorInfoFromGenesisModule extends AbstractModule {
   @Provides
   @Singleton
   @Self
-  private BFTValidatorId self(
+  private SelfValidatorInfo self(
```
dhedey, line 81: Nice. But as mentioned elsewhere, not sure if we need the @Self

core/src/main/java/com/radixdlt/keys/SelfValidatorInfoModule.java
```diff
@@ -68,15 +68,17 @@
...
 
-public final class BFTValidatorIdModule extends AbstractModule {
+public final class SelfValidatorInfoModule extends AbstractModule {
   @Provides
   @Self
-  public BFTValidatorId validatorId(
+  public SelfValidatorInfo selfValidatorInfo(
       @Self Optional<ComponentAddress> validatorAddress, @Self ECDSASecp256k1PublicKey key) {
-    return BFTValidatorId.create(validatorAddress.orElse(null), key);
+    return new SelfValidatorInfo(
```
dhedey, line 81: It feels a little weird to me this takes a @Self Optional<ComponentAddress> rather than computing that here?

Like it's weird to me that `@Self Optional<ComponentAddress>` is bound in the DI at all? SelfValidatorInfo feels like a nicer object to rely on :)

core/src/main/java/com/radixdlt/logger/EventLoggerModule.java
```diff
@@ -219,12 +221,13 @@ private static void processLedgerUpdate(
     */
   }
 
-  private static void logEpochChange(BFTValidatorId self, EpochChange epochChange) {
+  private static void logEpochChange(SelfValidatorInfo self, EpochChange epochChange) {
     var validatorSet = epochChange.getBFTConfiguration().getValidatorSet();
+    final var included = self.bftValidatorId().stream().anyMatch(validatorSet::containsValidator);
```
dhedey, line 226: I dislike the readability of calling `anyMatch` on an Optional, but I guess this is the best in-built Java way of doing this?

core/src/test-core/java/com/radixdlt/environment/NoEpochsConsensusModule.java
```diff
@@ -64,42 +64,256 @@
...
+      Metrics metrics,
+      EventDispatcher<RoundQuorumResolution> roundQuorumResolutionEventDispatcher,
+      ScheduledEventDispatcher<TimeoutQuorumDelayedResolution>
+          timeoutQuorumDelayedResolutionDispatcher,
+      EventDispatcher<ConsensusByzantineEvent> doubleVoteEventDispatcher,
+      EventDispatcher<ProposalRejected> proposalRejectedDispatcher,
+      RoundUpdate roundUpdate,
+      @TimeoutQuorumResolutionDelayMs long timeoutQuorumResolutionDelayMs) {
+    /*
+    TODO: consider cleaning this up (but most probably it's not worth it :))
```
dhedey, line 162: Out of interest, would you support ditching this and instead just using really long epochs? This part of the code always fills me with a sadness that we're testing mocks and things are configured in unrealistic and flaky ways.

Do you think supporting this provides us any benefits worth keeping it around for?

(Not something to address in this PR of course!)

dhedey (approved):
Really, really nice, thanks! A few thoughts I'd be interested to hear your thoughts on.

core/src/main/java/com/radixdlt/RunningRadixNode.java
LukasGasior1: yup, makes sense, done

core/src/main/java/com/radixdlt/consensus/epoch/EpochsConsensusModule.java
LukasGasior1: ohh I've missed it (and the PacemakerState bound in configure()), fixed. I believe we now only need the (initial) BFTConfiguration - which I'd also like to clean up at some point :)

core/src/main/java/com/radixdlt/logger/EventLoggerModule.java
LukasGasior1: I'm afraid so :man_shrugging: . There's good ol' if-else but that's arguably uglier than anyMatch

core/src/main/java/com/radixdlt/keys/SelfValidatorInfoModule.java
LukasGasior1: :+1:

core/src/test-core/java/com/radixdlt/environment/NoEpochsConsensusModule.java
LukasGasior1: That's an interesting idea, definitely something to consider

PR #548, opened 2023-07-11
Add node key / storage location checks to RadixNodeBootstrapper

LukasGasior1:

PR #549, opened 2023-07-11
Don't install BouncyCastleProvider globally, instead use it explicitly where needed

LukasGasior1:


core/src/main/java/com/radixdlt/RadixNodeApplication.java
```diff
@@ -86,32 +84,13 @@ private RadixNodeApplication() {}
...
     try {
       MemoryLeakDetector.start();
-
       logVersion();
       dumpExecutionLocation();
-      // Bouncy Castle is required for loading the node key, so set it up now.
-      setupBouncyCastle();
       final var properties = RuntimeProperties.fromCommandLineArgs(args);
       bootstrapRadixNode(properties);
-      log.info("Node bootstrapping completed");
```
dhedey, line 114: Out of interest, why was this removed?

core/src/main/java/com/radixdlt/RunningRadixNode.java
```diff
@@ -96,8 +96,6 @@ private RunningRadixNode(Injector injector) {
   }
 
   public static RunningRadixNode run(UnstartedRadixNode unstartedRadixNode) {
-    log.info("Starting Radix node");
```
dhedey, line 99: Out of interest, why was this removed?

dhedey (approved):
Approve of the BouncyCastle changes, and state manager log changes.

HostIp changes look okay - assuming they've been sufficiently tested

core/src/main/java/com/radixdlt/RunningRadixNode.java
LukasGasior1: too many open branches :sweat_smile: I've replaced this with `https://github.com/radixdlt/babylon-node/pull/548/files#diff-d14e0db8dc21558566c417273e02c97a19d29b5502fd22260c256bdbb412dbb6R178` in a different PR

core/src/main/java/com/radixdlt/RadixNodeApplication.java
LukasGasior1: we already have a similar message in RadixNodeApplication:
```Radix node {} started successfully in {} ms```

PR #551, opened 2023-07-12
DO-1577 / Keygen arm

kofrdx:
Building the keygen image for the arm architecture. This will move us closer to fully supporting arm for the nodecli.

### DIFF
--- a/.github/workflows/add-artifacts-to-release.yml
+++ b/.github/workflows/add-artifacts-to-release.yml
@@ -125,12 +125,6 @@ jobs:
           ls -lah libcorerust-${{ matrix.target }}
           zip -j babylon-node-rust-${{ matrix.zipname }}-release-${{ github.event.release.tag_name }}.zip libcorerust-${{ matrix.target }}/${{ matrix.artifact }}
       - if: github.event_name == 'release'
-        name: Get release
-        id: get_release
-        uses: bruceadams/get-release@74c3d60f5a28f358ccf241a00c9021ea16f0569f
-        env:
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -72,12 +72,17 @@ USER nobody
 # LAYER: Keygen
 # An alternative build target that executes the keygeneration
 # =================================================================================================
-# hadolint ignore=DL3029
-FROM --platform=linux/amd64 eclipse-temurin:17-jre-alpine AS keygen
+FROM eclipse-temurin:17.0.7_7-jre AS keygen
```
kofrdx, line 75: As you can see [here](https://hub.docker.com/_/eclipse-temurin/tags?page=1&name=alpine) alpine does not have arm releases. Switching to the base release of this.

.github/workflows/add-artifacts-to-release.yml
```diff
@@ -124,12 +124,6 @@ jobs:
         run: |
           ls -lah libcorerust-${{ matrix.target }}
           zip -j babylon-node-rust-${{ matrix.zipname }}-release-${{ github.event.release.tag_name }}.zip libcorerust-${{ matrix.target }}/${{ matrix.artifact }}
-      - if: github.event_name == 'release'
-        name: Get release
-        id: get_release
-        uses: bruceadams/get-release@74c3d60f5a28f358ccf241a00c9021ea16f0569f
```
kofrdx, line 130: This repo is not supported anymore and uses the outdated set-output that was originially gonna be removed starting end of june.

PR #552, opened 2023-07-13
Tame p2p outbound error logs

LukasGasior1:


core/src/main/java/com/radixdlt/messaging/core/MessageDispatcher.java
```diff
@@ -101,6 +106,12 @@ class MessageDispatcher {
   private final PeerManager peerManager;
   private final Addressing addressing;
 
+  private final RateLimiter ttlExpiredLogRateLimiter = RateLimiter.create(0.05);
+  private final AtomicInteger ttlExpiredMessagesCountSinceLastLog = new AtomicInteger(0);
+
+  private final LRUCache<NodeId, Pair<RateLimiter, AtomicInteger>>
```
jakrawcz-rdx, line 112: It looks like a pair `(RateLimiter, AtomicInteger)` actually represents a "rate-limited logger":
1. I vaguely remember that some generic solutions exist for that (for each major logging framework) and it would be cool to use one of them instead.
2. But I also see that you have some more specific needs (e.g. "apply rate-limit per receiver"), so maybe we need to implement it ourselves - but even in such case, I would strongly advice to extract a `class LogLimiter` and then have `private LogLimiter ttlExpired; private LRUCache<NodeId, LogLimiter> sendErrors;`

core/src/main/java/com/radixdlt/messaging/core/MessageDispatcher.java
```diff
@@ -137,26 +143,82 @@ CompletableFuture<Result<Unit, Cause>> send(final OutboundMessageEvent outboundM
...
+              final var cause = t.getCause() != null ? t.getCause().getMessage() : t.getMessage();
+              logSendError(message, receiver, cause);
+              return IO_ERROR.result();
+            });
+  }
+
+  private void logSendError(Message message, NodeId receiver, String cause) {
+    final RateLimiter rateLimiter;
+    final AtomicInteger countSinceLastLog;
+    if (sendErrorLogRateLimitersByReceiver.contains(receiver)) {
```
jakrawcz-rdx, line 162: hmm, I see a mix of thread-safe Atomics and non-thread-safe `LRUCache`, which makes me wonder - can this code be called from multiple threads? should we synchronize the cache access, or demote the AtomicInt to int?

core/src/main/java/com/radixdlt/messaging/core/MessageDispatcher.java
```diff
@@ -137,26 +143,82 @@ CompletableFuture<Result<Unit, Cause>> send(final OutboundMessageEvent outboundM
...
+      final var curr = sendErrorLogRateLimitersByReceiver.get(receiver).orElseThrow();
+      rateLimiter = curr.getFirst();
+      countSinceLastLog = curr.getSecond();
+    } else {
+      rateLimiter = RateLimiter.create(0.05);
+      countSinceLastLog = new AtomicInteger(0);
+      sendErrorLogRateLimitersByReceiver.put(receiver, Pair.of(rateLimiter, countSinceLastLog));
```
jakrawcz-rdx, line 169: BTW is there a reason we have our own `LRUCache` impl?
(e.g. do we want a very strict [deterministic, predictable] "LRU" behavior?)

Or do we rather want a practically-good eviction policy, tuned for hit-rate?
(if so, then we should probably go for guava/caffeine I think?)

LukasGasior1: I wouldn't really call it "our own impl", it's just a thin helper/wrapper around `LinkedHashMap`, no special requirements there, we could as well use Guava.

core/src/main/java/com/radixdlt/messaging/core/MessageDispatcher.java
LukasGasior1: huh, I was pretty sure this only runs on a single thread, the P2P thread (AtomicInteger was merely a mutable container), but as you mentioned it I double checked and it's actually multi (two) threaded. Nice catch :) I'll fix that.

core/src/main/java/com/radixdlt/messaging/core/MessageDispatcher.java
LukasGasior1: yeah, there are some configurable filters but I haven't found anything that matched my requirements :)
I extracted `RejectionCountingRateLimiter` (it doesn't really need to be logger-specific).

core/src/main/java/com/radixdlt/messaging/core/MessageDispatcher.java
LukasGasior1: done

core/src/main/java/com/radixdlt/messaging/core/MessageCentralImpl.java
```diff
@@ -144,7 +144,11 @@ public MessageCentralImpl(
     this.outboundThreadPool =
         new SimpleThreadPool<>(
             "Outbound message processing",
-            1, // Ensure messages sent in-order
+            /* This was meant to ensure messages are sent in order,
```
dhedey, line 147: :D - how did you discover this? And more importantly - what impact does this have?

dhedey: Also - does this mean we only ever send 1 message at any time?

core/src/main/java/com/radixdlt/messaging/core/MessageDispatcher.java
```diff
@@ -137,26 +136,77 @@ CompletableFuture<Result<Unit, Cause>> send(final OutboundMessageEvent outboundM
...
+        rateLimiter = sendErrorLogRateLimitersByReceiver.get(receiver).orElseThrow();
+      } else {
+        rateLimiter = new RejectionCountingRateLimiter(0.05);
+        sendErrorLogRateLimitersByReceiver.put(receiver, rateLimiter);
+      }
+      rateLimiter.tryAcquire(
+          countSinceLastPermit -> {
+            final var baseMsg =
+                String.format(
+                    "An outbound message of type %s couldn't be send to %s because of: \"%s\".",
```
dhedey, line 165: sent* to

core/src/main/java/com/radixdlt/messaging/core/MessageDispatcher.java
```diff
@@ -101,6 +100,11 @@ class MessageDispatcher {
   private final PeerManager peerManager;
   private final Addressing addressing;
 
+  private final RejectionCountingRateLimiter ttlExpiredLogRateLimiter =
+      new RejectionCountingRateLimiter(0.05);
+  private final LRUCache<NodeId, RejectionCountingRateLimiter> sendErrorLogRateLimitersByReceiver =
+      new LRUCache<>(100);
```
dhedey, line 106: Might be worth making it 150 so it's a little larger than the validator set (and can cover validators and seed nodes at Olympia launch)

core/src/main/java/com/radixdlt/messaging/core/MessageDispatcher.java
```diff
@@ -101,6 +100,11 @@ class MessageDispatcher {
   private final PeerManager peerManager;
   private final Addressing addressing;
 
+  private final RejectionCountingRateLimiter ttlExpiredLogRateLimiter =
+      new RejectionCountingRateLimiter(0.05);
```
dhedey, line 104: Could we comment that this translates to one log every 20 seconds?

core/src/main/java/com/radixdlt/messaging/core/MessageDispatcher.java
```diff
@@ -137,26 +136,77 @@ CompletableFuture<Result<Unit, Cause>> send(final OutboundMessageEvent outboundM
...
+            });
+  }
+
+  private void logSendError(Message message, NodeId receiver, String cause) {
+    synchronized (sendErrorLogRateLimitersByReceiver) {
+      final RejectionCountingRateLimiter rateLimiter;
+      if (sendErrorLogRateLimitersByReceiver.contains(receiver)) {
+        rateLimiter = sendErrorLogRateLimitersByReceiver.get(receiver).orElseThrow();
+      } else {
+        rateLimiter = new RejectionCountingRateLimiter(0.05);
```
dhedey, line 158: What does this 0.05 mean? We can have 20 logs per second per validator? Seems excessive still?

Or maybe it's one every 20 seconds? Perhaps we make it  once a minute (eg 1.0 / 60.0) and add a comment that explains it in human terms? OR make the RateLimiter take a TimeSpan instead? (or whatever the Java equivalent is?)

core/src/main/java/com/radixdlt/messaging/core/MessageDispatcher.java
```diff
@@ -137,26 +136,77 @@ CompletableFuture<Result<Unit, Cause>> send(final OutboundMessageEvent outboundM
...
+    }
+  }
+
+  private void logTtlExpired(Message message, NodeId receiver) {
+    synchronized (ttlExpiredLogRateLimiter) {
+      ttlExpiredLogRateLimiter.tryAcquire(
+          countSinceLastPermit -> {
+            final var baseMsg =
+                String.format(
+                    "TTL (of %s ms) has expired for an outbound message of type %s destined to %s.",
```
dhedey, line 188: If I read this message I wouldn't understand:
* Why might that be?
* What is the consequence?

Could we explain EG that the consequence is that the message is dropped? And (if it's true) that the Why is likely that there there is a message sending backlog?

core/src/main/java/com/radixdlt/utils/RejectionCountingRateLimiter.java
```diff
@@ -0,0 +1,101 @@
...
+import com.google.common.util.concurrent.RateLimiter;
+import java.util.function.Consumer;
+import javax.annotation.concurrent.NotThreadSafe;
+
+/**
+ * A wrapper around RateLimiter that counts the number of rejections between permits. This class is
+ * not thread-safe.
+ */
+@SuppressWarnings("UnstableApiUsage")
+@NotThreadSafe
```
dhedey, line 76: Can we not make it thread safe by using an AtomicInteger or something? I feel like RateLimiters are the kind of thing I would expect to be thread safe?

docker/scripts/rundocker.sh
```diff
@@ -23,7 +23,7 @@ export SKIP_NATIVE_RUST_BUILD=TRUE
 
 echo "||> Generating environment variables for running with $validators_count validators"
 # generateDevGenesis outputs a bunch of "export" lines to stdout.
-ENV_TO_LOAD=$(${reporoot}/gradlew -q -P "validators=${validators_count}" -P "networkId=240" ":cli-tools:generateDevGenesis")
+ENV_TO_LOAD=$(${reporoot}/gradlew -q -P "validators=${validators_count}" -P "network=240" ":cli-tools:generateDevGenesis")
```
dhedey, line 26: 👍 thanks for fixing this

core/src/main/java/com/radixdlt/messaging/core/MessageDispatcher.java
dhedey: This seems like quite a scary proposition - out of interest - do we have a metric for the backlog of messages in our outbound queue?

dhedey (approved):
A few relatively important comments, but appreciate you might not be able to get through them today. If not, don't worry.

core/src/main/java/com/radixdlt/utils/RejectionCountingRateLimiter.java
LukasGasior1: :+1:

core/src/main/java/com/radixdlt/messaging/core/MessageDispatcher.java
LukasGasior1: :+1: will change to once a minute

core/src/main/java/com/radixdlt/messaging/core/MessageDispatcher.java
LukasGasior1: we do have a metric: `metrics.messages().outbound().queued()`

core/src/main/java/com/radixdlt/utils/RejectionCountingRateLimiter.java
LukasGasior1: done (though we still need a separate lock on `sendErrorLogRateLimitersByReceiver` in MessageDispatcher)

core/src/main/java/com/radixdlt/utils/RejectionCountingRateLimiter.java
```diff
@@ -91,11 +90,14 @@ public RejectionCountingRateLimiter(double permitsPerSecond) {
    * attempts that have been rejected since the previous permit.
    */
   public void tryAcquire(Consumer<Integer> consumer) {
-    if (baseRateLimiter.tryAcquire()) {
-      consumer.accept(countSinceLastPermit);
-      countSinceLastPermit = 0;
-    } else {
-      countSinceLastPermit += 1;
-    }
+    countSinceLastPermit.getAndUpdate(
```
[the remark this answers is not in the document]
LukasGasior1: ahh, good catch, should have read the docs :facepalm: I purposely avoided your version as I somehow thought it leaves a possibility of +-1 error (if a thread is interrupted right after tryAcquire), but now I see that it's not really the case...the total count is always correct, but sometimes concurrent rejected acquire can be counted in the current `count` and sometimes in the next one, which is fine.

dhedey (approved):

PR #550, opened 2023-07-12
Introducing a common JNI call wrapper to surface all Rust panics as Java exceptions

jakrawcz-rdx:
This is the first part of the ["Handle panics nicer"](https://whimsical.com/babylon-work-tracker-Am2s7hSa7xHd5yCxgub67a@7YNFXnKbZAPNdd1f6K1ae) effort.

Some notes:
- The basic idea is captured in the title, and the impl details are captured in the rustdoc of `jni_call()`
  - Actually that `jni_call()` introduces a new "convention" that we must adhere to in our JNI layer, if we want to stay panic-safe (this PR propagates that convention across the current codebase)
- This PR does not introduce any new error-handling / exception-handling behavior; in particular:
  - Previously, a Rust `panic()` inside JNI was causing the app to shut down (in an "undefined behavior" way - but reliably)
  - After this PR, instead, a Java `RustPanicException` is unhandled in the RX loop, which initiates the hopefully-more-graceful shutdown + gives a nice continuation of the backtrace on Java's side

PR #554, opened 2023-07-14
Catching panics in Core API and returning HTTP 500

jakrawcz-rdx:
A continuation of https://github.com/radixdlt/babylon-node/pull/550 but for Core API.

The implementation: use the `tower-http`'s `CatchPanic` layer to transform panics into error responses.

### DIFF
--- a/core-rust/core-api-server/src/core_api/errors.rs
+++ b/core-rust/core-api-server/src/core_api/errors.rs
@@ -1,10 +1,13 @@
+use axum::body::BoxBody;
 use axum::{
     response::{IntoResponse, Response},
     Json,
 };
+use std::any::Any;
 
 use hyper::StatusCode;
 use radix_engine_interface::network::NetworkDefinition;
+use tower_http::catch_panic::ResponseForPanic;
 
 use super::models;
 use models::{
@@ -66,6 +69,23 @@ impl ErrorDetails for LtsTransactionSubmitErrorDetails {
             trace_id,
             details: details.map(Box::new),
         }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub(crate) struct InternalServerErrorResponseForPanic;
+
+impl ResponseForPanic for InternalServerErrorResponseForPanic {
+    type ResponseBody = BoxBody;
+
+    fn response_for_panic(
+        &mut self,
+        _panic_payload: Box<dyn Any + Send + 'static>,
+    ) -> Response<Self::ResponseBody> {
+        // Please note that we deliberately do *not*:
+        // - log the panic payload (since the default panic handler already does this);
+        // - include the panic payload in the response (it may contain sensitive details).
+        server_error::<()>("Unexpected server error").into_response()
     }
 }
 
--- a/core-rust/Cargo.lock
+++ b/core-rust/Cargo.lock
@@ -44,7 +44,7 @@ checksum = "f8175979259124331c1d7bf6586ee7e0da434155e4b2d48ec2c8386281d8df39"
 dependencies = [
  "async-trait",
  "axum-core",
- "bitflags",
+ "bitflags 1.3.2",
  "bytes",
  "futures-util",
  "http",
@@ -97,7 +97,7 @@ version = "0.64.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c4243e6031260db77ede97ad86c27e501d646a27ab57b59a574f725d98ab1fb4"
 dependencies = [
- "bitflags",
+ "bitflags 1.3.2",
  "cexpr",
  "clang-sys",
  "lazy_static",
@@ -116,6 +116,12 @@ name = "bitflags"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
+
+[[package]]
+name = "bitflags"
+version = "2.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "630be753d4e58660abd17930c71b647fe46c27ea6b63cc59e1e3851406972e42"
 
 [[package]]
 name = "bitmaps"
@@ -333,6 +339,7 @@ dependencies = [
  "serde_json",
  "state-manager",
  "tokio",
+ "tower-http",
  "tracing",
  "transaction",
  "utils",
@@ -741,6 +748,12 @@ dependencies = [
  "http",
  "pin-project-lite",
 ]
+
+[[package]]
+name = "http-range-header"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0bfe8eed0a9285ef776bb792479ea3834e8b94e13d615c2f66d03dd50a435a29"
 
 [[package]]
 name = "httparse"
@@ -1379,7 +1392,7 @@ version = "0.9.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "77a1a2f1f0a7ecff9c31abbe177637be0e97a0aef46cf8738ece09327985d998"
 dependencies = [
- "bitflags",
+ "bitflags 1.3.2",
  "memchr",
  "unicase",
 ]
@@ -1398,7 +1411,7 @@ name = "radix-engine"
 version = "0.10.0"
 source = "git+https://github.com/radixdlt/radixdlt-scrypto?tag=rcnet-v2-638673d16e#638673d16e83e1f04ffd7e44d2a0254a6e54a07e"
 dependencies = [
- "bitflags",
+ "bitflags 1.3.2",
  "colored",
  "hex",
  "lazy_static",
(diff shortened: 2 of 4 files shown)


jakrawcz-rdx:
Based on the above, I can tell that this PR **only** improves the logging and surfacing (HTTP 500) a bit.
The panic itself was always caught by tokio worker pool, and thus the Node was "recovering" from it.

I've been experimenting with the true "recoverability" after panics (i.e. re-trying the same / other requests after catching a panic):

- When caught without any held lock/mutex: Node 100% operational 🟢 _(doh!)_
- When caught while holding a `RwLock.read()`: Node 100% operational 🟢 _(makes sense, nice!)_
- When caught while holding a `RwLock.write()`: Node 100% operational 🟢 _(huh, wut?)_
- When caught while holding an **std::sync** `Mutex`: Other parts of Node (not requiring that `Mutex`) are ~fine 🟡 , but accessing the same `Mutex` causes another panic, this time due to `PoisonError` (caught and surfaced as HTTP 500) 🔴 _(kinda expected, but OFC not perfect)_

So, the `parking_lot`'s lock primitives do _not_ follow the "panic handling" approach of std's `Mutex` (i.e. they do not detect panic on drop; they don't become poisoned) - actually it is documented behavior, see https://amanieu.github.io/parking_lot/parking_lot/struct.Mutex.html#differences-from-the-standard-library-mutex

The question is: what are we going to do about it?

From @malexandru-rdx I've heard that it would be best to kill the app after such HTTP 500. It is true that the app may be left in inconsistent state (half-modified state protected by a lock).
I think that we don't need to be so extreme for read-lock cases (and maybe even for some write-lock cases). But it's hard to say...

malexandru-rdx (approved):
I'm happy to have this merged as is with a TODO/Comment on the ongoing issues since it's an improvement nonetheless.

Obviously it's rather a requirement to have validators not crash because of `core-api`, all I advocate for is: if we can't treat all cases such that `core-api` does not affect the good functioning of consensus (i.e. hidden poisoned lock on state manager) - I'd much rather have it panic than enter a deadlock or worse corrupt data.

Is the `parking_lot` performance really that significant and worth right now? I would much rather have it "safe".

There are other, better paths to optimize this.

dhedey:
Personally, I'm in the "panics should cause node to exit" camp.

If we want a node not to error from an API call, we return a `MappingError` or some other error which can map to the 500s.

However:
* I didn't realise `tokio` was capturing / swallowing panics for us.
* I didn't realise `parking_lot` mutexes don't get poisoned

I wonder if we should instead be calling a callback into Java on panic which exits the node?

malexandru-rdx (commented):
>I wonder if we should instead be calling a callback into Java on panic which exits the node, similar to JNI panics?

I would make at least all locks used by `core-api` behave as the `std::` one and be marked as poisoned after a panic (either use those or wrap `parking_lot` ones), keep changes from this PR and let the panic be raised from outside Tokio when those poisoned locks are tried.

This way it's impossible for `core-api` to crash the node for read-only endpoints.

jakrawcz-rdx:
> If we want a node not to error from an API call, we return a MappingError or some other error which can map to the 500s

That's a nice general rule (in Rust, AFAIU), but there is always gonna be a bug where we do `result.expect("must be successful because we check it 10 lines earlier")` (but a check was changed/moved during a latter refactor).

Is it even realistic to never `.unwrap()` (but propagate `Err` everywhere)? I thought that our panic-catching is supposed to catch this kind of bugs and just "handle" them more gracefully.

Now the question is: what does it mean to "handle" a bug in runtime? To me:
- at a very minimum ❗ : log it in a very clear way, both in tests and on prod.
- nice to have 👍 : surface a clear information to the user (HTTP 500 is a perfect "we have a bug" response).
- valuable, but debatable ❓ : if possible, allow the app to continue running:
  - it seems like this _is_ possible if a bug is on some read-only path of a request-based processing?
  - it also _may be_ possible for certain write paths (assuming transactionality mechanisms can account for that)?

jakrawcz-rdx:
Let me capture the conclusions from our call:

Ideally we would like the panics to shutdown the application in "dangerous" cases (i.e. where state can be corrupted), and HTTP 500 in "safe" cases (e.g. simple ready-only endpoints).
This can be achieved automatically (i.e. driven by poisoning of write locks), or manually (i.e. explicit closures marking dangerous xor safe code).

@dhedey @malexandru-rdx are you ok with addressing the above in next PR?

_(after all, this PR does _not_ make the current `tokio + parking_lot` behavior worse: the CoreApi panics are still 100% recoverable, just returned as HTTP 500 instead of empty bytes)_

core-rust/core-api-server/src/core_api/errors.rs
```diff
@@ -69,6 +72,23 @@ impl ErrorDetails for LtsTransactionSubmitErrorDetails {
...
+    type ResponseBody = BoxBody;
+
+    fn response_for_panic(
+        &mut self,
+        _panic_payload: Box<dyn Any + Send + 'static>,
+    ) -> Response<Self::ResponseBody> {
+        // Please note that we deliberately do *not*:
+        // - log the panic payload (since the default panic handler already does this);
+        // - include the panic payload in the response (it may contain sensitive details).
+        server_error::<()>("Unexpected server error").into_response()
```
dhedey, line 88: We should ideally return a GUID / Trace ID (like we do in the Gateway) so we/they can correlate the error response with our logs of the panic.

(Perhaps on all error responses?)

Happy for us to pursue this in a later PR

dhedey (approved):

core-rust/core-api-server/src/core_api/errors.rs
jakrawcz-rdx: of course this sounds like a good idea

> Happy for us to pursue this in a later PR

...yup, especially if that's a global change, for all existing error responses.

jakrawcz-rdx: Alright, so I see 2 options:

**A)** _[easy]_ Let's simply always generate a `LogTraceId` when `trace: None`. Then, magically, on errors:
- The user will get: `{"error_type":"Basic","code":500,"message":"Unexpected server error","trace_id":"abcdefgh12345678"}`
- Our logs will get:
```
# 2023-07-20T15:32:48,851 [INFO/EventLoggerModule/BFT no...64vp9cx0t] (EventLoggerModule.java:166) - bft_nxtrnd ... // regular logs, and then suddenly a panic:
thread 'tokio-runtime-worker' panicked at 'a bug happened', core-api-server/src/core_api/handlers/mempool_list.rs:9:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/39f2657d1101b50f9b71ae460b762d330cc8426b/library/std/src/panicking.rs:579:5
   1: core::panicking::panic_fmt
             at /rustc/39f2657d1101b50f9b71ae460b762d330cc8426b/library/core/src/panicking.rs:64:14
             at /Users/jakrawcz/repo1/babylon-node/core-rust/core-api-server/src/core_api/handlers/mempool_list.rs:9:9
   ... // snipped
 127: tokio::runtime::blocking::pool::Spawner::spawn_thread::{{closure}}
2023-07-20T13:32:49.742979Z  WARN core_api_server::core_api::server: path="/core/mempool/list" error=BasicErrorResponse { code: 500, message: "Unexpected server error", trace_id: Some("abcdefgh12345678") }
# 2023-07-20T15:32:49,854 [INFO/EventLoggerModule/BFT no...64vp9cx0t] (EventLoggerModule.java:249) - lgr_commit ... // next logs
```
_(this means: `abcdefgh12345678` is not mentioned in the stack trace printed by a panic hook, but only in a log that should be right next **BUT OFC sometimes it's gonna be ambiguous**)_

**B)** _[too clever]_ Let's install our own panic hook, which will generate a trace ID, and will print it together with the rest of panic info. THEN it will do `thread_local.latest_panic_trace_id = generated_trace_id`, and later our http-level logger will do `trace: Some(thread_local.latest_panic_trace_id)`.
- The user will get the same as in **A)**
- Our logs will get:
```
# 2023-07-20T15:32:48,851 [INFO/EventLoggerModule/BFT no...64vp9cx0t] (EventLoggerModule.java:166) - bft_nxtrnd ... // regular logs, and then suddenly a panic with trace_id:
thread 'tokio-runtime-worker' panicked at 'a bug happened' with trace ID 'abcdefgh12345678': , core-api-server/src/core_api/handlers/mempool_list.rs:9:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/39f2657d1101b50f9b71ae460b762d330cc8426b/library/std/src/panicking.rs:579:5
   1: core::panicking::panic_fmt
             at /rustc/39f2657d1101b50f9b71ae460b762d330cc8426b/library/core/src/panicking.rs:64:14
             at /Users/jakrawcz/repo1/babylon-node/core-rust/core-api-server/src/core_api/handlers/mempool_list.rs:9:9
   ... // snipped
 127: tokio::runtime::blocking::pool::Spawner::spawn_thread::{{closure}}
2023-07-20T13:32:49.742979Z  WARN core_api_server::core_api::server: path="/core/mempool/list" error=BasicErrorResponse { code: 500, message: "Unexpected server error", trace_id: Some("abcdefgh12345678") }
# 2023-07-20T15:32:49,854 [INFO/EventLoggerModule/BFT no...64vp9cx0t] (EventLoggerModule.java:249) - lgr_commit ... // next logs
```
_(this means: `abcdefgh12345678` is mentioned in the stack trace printed by a panic hook, and it should be **99.9% unambiguous**, except for crazy "panic overwrite" scenarios [like, panic inside a drop during stack unwinding] - at least that's how I understand rust docs on panic hook and unwinding)_

PR #555, opened 2023-07-14
Fix series executor state update for committed, but later rejected transactions

LukasGasior1:


malexandru-rdx (approved):
Thanks! Shouldn't this be merged into `release/rcnet-v2`?

LukasGasior1:
> Shouldn't this be merged into release/rcnet-v2

hmh, good point ... wdyt @dhedey ? do we reuse the same branch for phase 2? :thinking:

dhedey (approved):
Please merge into `release/rcnet-v2`, and then we can merge that into `develop` and into `release/rcnet-v2-phase2` when we create it.

PR #556, opened 2023-07-17
Make protocol constants configurable

malexandru-rdx:
In first commit I refactor proposal constants and group them in a `ProposalLimitsConfig` struct.
In second commit I add an optional `VertexLimitsConfig` that configures vertex limits. Please note that these should not ever be changed in production and are exposed (temporarily) in order to benchmark different settings and search the final values.

| Config Field | Docker Environment Variable | Default Value | Description |
| --- | --- | --- | --- |
| `protocol.vertex.max_transaction_count` | `RADIXDLT_PROTOCOL_VERTEX_MAX_TRANSACTION_COUNT` | `10` | Maximum number of transactions per vertex |
| `protocol.vertex.max_total_transactions_size` | `RADIXDLT_PROTOCOL_VERTEX_MAX_TOTAL_TRANSACTIONS_SIZE` | `3.8 * 1024 * 1024` | Maximum total size of transactions payload inside of a vertex |
| `protocol.vertex.max_total_execution_cost_units_consumed` | `RADIXDLT_PROTOCOL_VERTEX_MAX_TOTAL_EXECUTION_COST_UNITS_CONSUMED` | `200_000_000` | Maximum execution cost units consumed by all transactions in the vertex |


core/src/main/java/com/radixdlt/rev2/REv2StateComputer.java
```diff
@@ -107,18 +108,7 @@ public final class REv2StateComputer implements StateComputerLedger.StateCompute
...
-  private final int maxProposalTotalTxnsPayloadSize;
-  // Maximum number of transaction payload bytes to include in a proposal and its previous vertices
-  // chain.
-  // Intended to limit the size of a commit batch (i.e. the size of transactions under a single
-  // commit proof).
-  // Note - we can still keep committing round changes, so this is not a guarantee. But should be
-  // reasonably
-  // effective as a limit.
-  private final int maxUncommittedTotalPayloadSize;
+  ProposalLimitsConfig proposalLimitsConfig;
```
jakrawcz-rdx, line 111:
```suggestion
  private final ProposalLimitsConfig proposalLimitsConfig;
```

core/src/main/java/com/radixdlt/consensus/ProposalLimitsConfig.java
```diff
@@ -0,0 +1,128 @@
...
+  public static ProposalLimitsConfig from(VertexLimitsConfig vertexLimitsConfig) {
+    return new ProposalLimitsConfig(
+        // we add an extra transaction to build the vertex from
+        vertexLimitsConfig.maxTransactionCount().toInt() + 1,
+        // even though we would like to fill the vertex transaction count to the maximum possible,
+        // we want to keep
+        // bandwidth strict
+        vertexLimitsConfig.maxTotalTransactionsSize().toInt(),
+        // we multiply by 3 for the 3-chain rule
+        vertexLimitsConfig.maxTotalTransactionsSize().toInt() * 3);
```
jakrawcz-rdx, line 115: these lines look like `ProposalLimitsConfig` is actually computable from `VertexLimitsConfig` in a very reasonable way - do we still need to be able to specify the `ProposalLimitsConfig` independently?

core-rust-bridge/src/main/java/com/radixdlt/statemanager/VertexLimitsConfig.java
```diff
@@ -0,0 +1,95 @@
...
+
+import com.radixdlt.sbor.codec.CodecMap;
+import com.radixdlt.sbor.codec.StructCodec;
+import com.radixdlt.utils.UInt32;
+
+public record VertexLimitsConfig(
+    UInt32 maxTransactionCount,
+    UInt32 maxTotalTransactionsSize,
+    UInt32 maxTotalExecutionCostUnitsConsumed) {
+  // Keep in sync with core-rust/node-common/src/config/limits.rs
```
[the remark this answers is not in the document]
malexandru-rdx: All of them should be tunable, it's just that right now they are not implemented due to RE removing them and I don't see a reason to expose them too. I don't think the naming still holds in this case.

> This point is critical to me since a // Keep in sync with ... comment is the most unreliable way of keeping things in sync (especially when I am a part of the team :D)

Fair enough. What do you think about second option: exposing Rust constants to Java?

core/src/main/java/com/radixdlt/consensus/ProposalLimitsConfig.java
malexandru-rdx: I was also thinking exposing proposal limits as well to tests different cases but this is second priority and we can do after a first round of tuning.

core-rust-bridge/src/main/java/com/radixdlt/statemanager/VertexLimitsConfig.java
```diff
@@ -0,0 +1,96 @@
...
+
+import com.radixdlt.sbor.codec.CodecMap;
+import com.radixdlt.sbor.codec.StructCodec;
+import com.radixdlt.utils.UInt32;
+
+public record VertexLimitsConfig(
+    UInt32 maxTransactionCount,
+    UInt32 maxTotalTransactionsSize,
+    UInt32 maxTotalExecutionCostUnitsConsumed) {
+  // TODO: exposed Rust constants to Java so we keep a single source of truth
```
jakrawcz-rdx, line 75:
```suggestion
  // TODO(follow-up refactor): expose these (and other relevant) Rust constants to Java so we keep a single source of truth
```

jakrawcz-rdx (approved):

PR #558, opened 2023-07-18
Shutdown the Node completely (in Prod, in SimulationTest) on uncaught rx-java exceptions

jakrawcz-rdx:
This fixes the problem overlooked by https://github.com/radixdlt/babylon-node/pull/550, described at https://github.com/radixdlt/babylon-node/pull/550#issuecomment-1638094885.

Here is an example log for a failing `com.radixdlt.integration.steady_state.simulation.rev2.consensus_mempool_ledger_sync.SanityTest` after this PR:

```
... // Node runs
# 2023-07-18T13:35:01,953 [INFO/EventLoggerModule/BFT no...x3x8kavk8] (EventLoggerModule.java:249) - lgr_commit{epoch=2 round=109 version=115 txn_root=579900040f428f89 user_txns=0}
# 2023-07-18T13:35:01,953 [INFO/EventLoggerModule/BFT no...m0jvla7jl] (EventLoggerModule.java:249) - lgr_commit{epoch=2 round=109 version=115 txn_root=579900040f428f89 user_txns=0}
... // one of the rx-java workers gets an uncaught error (here, caused by a rust panic, but that's not a requirement)
# 2023-07-18T13:35:02,601 [ERROR/RunningNetworkImpl/BFT no...0w20kj09v] (SimulationNodes.java:263) - uncaught consensus error; stopping the network
com.radixdlt.exceptions.RustPanicException: random panic in prepare
	at com.radixdlt.statecomputer.RustStateComputer.prepare(Native Method) ~[core-rust-bridge-SNAPSHOT-refactor~complete_shutdown_on_uncaught_rx_error-809ae27093.jar:?]
	at com.radixdlt.sbor.Natives.lambda$builder$0(Natives.java:140) ~[core-rust-bridge-SNAPSHOT-refactor~complete_shutdown_on_uncaught_rx_error-809ae27093.jar:?]
	... // snipped
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) [?:?]
	at java.lang.Thread.run(Thread.java:833) [?:?]


Expecting all elements of:
  {CONSENSUS_SAFETY=Optional.empty, CONSENSUS_LIVENESS=Optional[TestInvariantError{desc=Highest QC hasn't increased from EpochRound{epoch=2 round=116} after 10 SECONDS}], CONSENSUS_NO_TIMEOUTS=Optional.empty, CONSENSUS_NO_EPOCH_TIMEOUTS=Optional.empty, CONSENSUS_DIRECT_PARENTS=Optional.empty, CONSENSUS_TO_LEDGER_PROCESSED=Optional.empty, PROPOSER_TIMESTAMP_CHECK=Optional.empty, LEDGER_IN_ORDER=Optional.empty}
to satisfy given requirements, but these elements did not:

... // the test fails due to liveness break (because of entire network being shut down)
CONSENSUS_LIVENESS=Optional[TestInvariantError{desc=Highest QC hasn't increased from EpochRound{epoch=2 round=116} after 10 SECONDS}]
error: 
Expecting an empty Optional but was containing value: TestInvariantError{desc=Highest QC hasn't increased from EpochRound{epoch=2 round=116} after 10 SECONDS}
java.lang.AssertionError: 
Expecting all elements of:
  {CONSENSUS_SAFETY=Optional.empty, CONSENSUS_LIVENESS=Optional[TestInvariantError{desc=Highest QC hasn't increased from EpochRound{epoch=2 round=116} after 10 SECONDS}], CONSENSUS_NO_TIMEOUTS=Optional.empty, CONSENSUS_NO_EPOCH_TIMEOUTS=Optional.empty, CONSENSUS_DIRECT_PARENTS=Optional.empty, CONSENSUS_TO_LEDGER_PROCESSED=Optional.empty, PROPOSER_TIMESTAMP_CHECK=Optional.empty, LEDGER_IN_ORDER=Optional.empty}
to satisfy given requirements, but these elements did not:

CONSENSUS_LIVENESS=Optional[TestInvariantError{desc=Highest QC hasn't increased from EpochRound{epoch=2 round=116} after 10 SECONDS}]
error: 
Expecting an empty Optional but was containing value: TestInvariantError{desc=Highest QC hasn't increased from EpochRound{epoch=2 round=116} after 10 SECONDS}
	at com.radixdlt.integration.steady_state.simulation.rev2.consensus_mempool_ledger_sync.SanityTest.rev2_consensus_mempool_ledger_sync_cause_no_unexpected_errors(SanityTest.java:141)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	... // usual Gradle log
	at worker.org.gradle.process.internal.worker.GradleWorkerMain.main(GradleWorkerMain.java:74)


OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended
[2m2023-07-18T11:34:46.476326Z[0m [32m INFO[0m [2mstate_manager::state_manager[0m[2m:[0m Committing system flash
[2m2023-07-18T11:34:46.483754Z[0m [32m INFO[0m [2mstate_manager::state_manager[0m[2m:[0m Committing system bootstrap
... // rust logs, printed after java logs for unclear reasons, as they always are
[2m2023-07-18T11:34:47.579982Z[0m [32m INFO[0m [2mstate_manager::state_manager[0m[2m:[0m Genesis transactions successfully executed in 188.245ms
thread '<unnamed>' panicked at 'random panic in prepare', state-manager/src/state_manager.rs:330:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[0K[1mcom.radixdlt.integration.steady_state.simulation.rev2.consensus_mempool_ledger_sync.SanityTest[22m rev2_consensus_mempool_ledger_sync_cause_no_unexpected_errors[31m FAILED[31m (29.4s)[31m
```

### DIFF
--- a/core/src/main/java/com/radixdlt/RunningRadixNode.java
+++ b/core/src/main/java/com/radixdlt/RunningRadixNode.java
@@ -64,14 +64,13 @@
 
 package com.radixdlt;
 
+import com.google.common.base.Preconditions;
 import com.google.inject.Injector;
 import com.google.inject.Key;
 import com.google.inject.TypeLiteral;
 import com.radixdlt.api.CoreApiServer;
 import com.radixdlt.api.prometheus.PrometheusApi;
 import com.radixdlt.api.system.SystemApi;
-import com.radixdlt.consensus.bft.BFTValidatorId;
-import com.radixdlt.consensus.bft.Self;
 import com.radixdlt.consensus.bft.SelfValidatorInfo;
 import com.radixdlt.consensus.safety.PersistentSafetyStateStore;
 import com.radixdlt.environment.Runners;
@@ -84,16 +83,22 @@ import com.radixdlt.statemanager.StateManager;
 import java.time.Duration;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
+import javax.annotation.Nullable;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 public final class RunningRadixNode {
   private static final Logger log = LogManager.getLogger();
+  public static RunningRadixNode ki;
 
   private final Injector injector;
 
+  private final AtomicReference<Thread> shutdownHook;
+
   private RunningRadixNode(Injector injector) {
     this.injector = injector;
+    this.shutdownHook = new AtomicReference<>();
   }
 
   public static RunningRadixNode run(UnstartedRadixNode unstartedRadixNode) {
@@ -101,6 +106,7 @@ public final class RunningRadixNode {
     log.info("Using a genesis of hash {}", unstartedRadixNode.genesisProvider().genesisDataHash());
 
     final var injector = unstartedRadixNode.instantiateRadixNodeModule();
+    final var runningNode = new RunningRadixNode(injector);
 
     final var metrics = injector.getInstance(Metrics.class);
     injector.getInstance(MetricInstaller.class).installAt(metrics);
@@ -118,7 +124,11 @@ public final class RunningRadixNode {
 
     for (var module : moduleStartOrder) {
       final var moduleRunner = moduleRunners.get(module);
-      moduleRunner.start();
+      moduleRunner.start(
+          error -> {
+            log.error("Unhandled exception in runner {}; shutting down the node", module, error);
+            runningNode.shutdown();
+          });
     }
 
     final var peerServer = injector.getInstance(PeerServerBootstrap.class);
@@ -136,7 +146,7 @@ public final class RunningRadixNode {
     final var coreApiServer = injector.getInstance(CoreApiServer.class);
     coreApiServer.start();
 
-    return new RunningRadixNode(injector);
+    return runningNode;
   }
 
   public SelfValidatorInfo self() {
@@ -147,21 +157,36 @@ public final class RunningRadixNode {
     this.injector.getInstance(Metrics.class).misc().nodeStartup().observe(startupTimeMs);
   }
 
+  public void installShutdownHook() {
+    Thread hook = new Thread(this::shutdownFromHook);
+    final var set = this.shutdownHook.compareAndSet(null, hook);
+    Preconditions.checkState(set, "already installed before");
+    Runtime.getRuntime().addShutdownHook(hook);
+  }
+
   public void shutdown() {
+    final @Nullable Thread hook = this.shutdownHook.getAndSet(null);
+    if (hook != null) {
+      Runtime.getRuntime().removeShutdownHook(hook);
+    }
+    doShutdown("internal reasons");
+  }
+
+  private void shutdownFromHook() {
+    doShutdown("a shutdown hook");
+  }
+
+  private void doShutdown(String reason) {
     // using System.out.printf as logger no longer works reliably in a shutdown hook
-    final var self = injector.getInstance(Key.get(BFTValidatorId.class, Self.class));
-    System.out.printf("Node %s is shutting down...\n", self);
+    System.out.printf("Node %s is shutting down due to %s...\n", this.self(), reason);
 
     injector
         .getInstance(Key.get(new TypeLiteral<Map<String, ModuleRunner>>() {}))
+        .values()
         .forEach(
-            (k, moduleRunner) -> {
-              try {
-                moduleRunner.stop();
-              } catch (Exception e) {
-                logShutdownError("ModuleRunner " + moduleRunner.threadName(), e.getMessage());
-              }
-            });
+            moduleRunner ->
+                catchAllAndLogShutdownError(
+                    "ModuleRunner " + moduleRunner.threadName(), moduleRunner::stop));
 
     catchAllAndLogShutdownError(
         "AddressBookPersistence", () -> injector.getInstance(AddressBookPersistence.class).close());
--- a/core/src/main/java/com/radixdlt/RunningRadixNode.java
+++ b/core/src/main/java/com/radixdlt/RunningRadixNode.java
@@ -90,7 +90,6 @@ import org.apache.logging.log4j.Logger;
 
 public final class RunningRadixNode {
   private static final Logger log = LogManager.getLogger();
-  public static RunningRadixNode ki;
 
   private final Injector injector;
 
@@ -126,8 +125,8 @@ public final class RunningRadixNode {
       final var moduleRunner = moduleRunners.get(module);
       moduleRunner.start(
           error -> {
-            log.error("Unhandled exception in runner {}; shutting down the node", module, error);
-            runningNode.shutdown();
+            log.error("Uncaught exception in runner {}; shutting down the node", module, error);
+            runningNode.shutdownFromModule(String.format("an error in runner %s", module));
           });
     }
 
@@ -164,16 +163,16 @@ public final class RunningRadixNode {
     Runtime.getRuntime().addShutdownHook(hook);
   }
 
-  public void shutdown() {
+  private void shutdownFromModule(String reason) {
     final @Nullable Thread hook = this.shutdownHook.getAndSet(null);
     if (hook != null) {
       Runtime.getRuntime().removeShutdownHook(hook);
     }
-    doShutdown("internal reasons");
+    doShutdown(reason);
   }
 
   private void shutdownFromHook() {
-    doShutdown("a shutdown hook");
+    doShutdown("a requested shutdown (via the Java shutdown hook)");
   }
 
   private void doShutdown(String reason) {
--- a/core/src/main/java/com/radixdlt/RunningRadixNode.java
+++ b/core/src/main/java/com/radixdlt/RunningRadixNode.java
@@ -64,7 +64,6 @@
 
 package com.radixdlt;
 
-import com.google.common.base.Preconditions;
 import com.google.inject.Injector;
 import com.google.inject.Key;
 import com.google.inject.TypeLiteral;
@@ -83,8 +82,6 @@ import com.radixdlt.statemanager.StateManager;
 import java.time.Duration;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.atomic.AtomicReference;
-import javax.annotation.Nullable;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -93,11 +90,8 @@ public final class RunningRadixNode {
 
   private final Injector injector;
 
-  private final AtomicReference<Thread> shutdownHook;
-
   private RunningRadixNode(Injector injector) {
     this.injector = injector;
-    this.shutdownHook = new AtomicReference<>();
   }
 
   public static RunningRadixNode run(UnstartedRadixNode unstartedRadixNode) {
@@ -125,8 +119,8 @@ public final class RunningRadixNode {
       final var moduleRunner = moduleRunners.get(module);
       moduleRunner.start(
           error -> {
-            log.error("Uncaught exception in runner {}; shutting down the node", module, error);
-            runningNode.shutdownFromModule(String.format("an error in runner %s", module));
+            log.error("Uncaught exception in runner {}; exiting the process", module, error);
+            System.exit(-1);
           });
     }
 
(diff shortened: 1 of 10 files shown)


jakrawcz-rdx:
For comparison, the `develop`'s log of the same:

```
... // Node runs
# 2023-07-18T13:44:33,272 [INFO/EventLoggerModule/BFT no...tesf6qf6p] (EventLoggerModule.java:249) - lgr_commit{epoch=2 round=112 version=119 txn_root=fa613fca64a3a8f2 user_txns=0}
# 2023-07-18T13:44:33,382 [INFO/EventLoggerModule/BFT no...m0jvla7jl] (EventLoggerModule.java:249) - lgr_commit{epoch=2 round=113 version=120 txn_root=b4e4587054de2b93 user_txns=0}
# 2023-07-18T13:44:33,442 [ERROR/RxModuleRunnerImpl/BFT no...0w20kj09v] (RxModuleRunnerImpl.java:112) - Unhandled exception in the event processing loop. Shutting down the node.
com.radixdlt.exceptions.RustPanicException: random panic in prepare
	at com.radixdlt.statecomputer.RustStateComputer.prepare(Native Method) ~[core-rust-bridge-SNAPSHOT-develop-30119380a1.jar:?]
	... // snipped
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) [?:?]
	at java.lang.Thread.run(Thread.java:833) [?:?]
OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended
[2m2023-07-18T11:44:16.757363Z[0m [32m INFO[0m [2mstate_manager::state_manager[0m[2m:[0m Committing system flash
[2m2023-07-18T11:44:16.764849Z[0m [32m INFO[0m [2mstate_manager::state_manager[0m[2m:[0m Committing system bootstrap
 ... // same rust logs at the end
[2m2023-07-18T11:44:17.923588Z[0m [32m INFO[0m [2mstate_manager::state_manager[0m[2m:[0m Genesis transactions successfully executed in 199.619667ms
thread '<unnamed>' panicked at 'random panic in prepare', state-manager/src/state_manager.rs:330:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[0K[1mcom.radixdlt.integration.steady_state.simulation.rev2.consensus_mempool_ledger_sync.SanityTest[22m rev2_consensus_mempool_ledger_sync_cause_no_unexpected_errors[33m SKIPPED[m
[0K[1;32mSUCCESS: [39mExecuted 1 tests in 20.9s (1 skipped)[m
> Task :core:runSteadyStateIntegrationTests FAILED
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':core:runSteadyStateIntegrationTests'.
... // the hated line, which no longer occurs after this PR:
> Process 'Gradle Test Executor 12' finished with non-zero exit value 255
  This problem might be caused by incorrect test process configuration.
  Please refer to the test execution section in the User Manual at https://docs.gradle.org/7.2/userguide/java_testing.html#sec:test_execution
```

core/src/main/java/com/radixdlt/RunningRadixNode.java
```diff
@@ -147,21 +157,36 @@ public void reportSelfStartupTime(Duration startupTimeMs) {
...
+
+  private void shutdownFromHook() {
+    doShutdown("a shutdown hook");
+  }
+
+  private void doShutdown(String reason) {
     // using System.out.printf as logger no longer works reliably in a shutdown hook
-    final var self = injector.getInstance(Key.get(BFTValidatorId.class, Self.class));
-    System.out.printf("Node %s is shutting down...\n", self);
+    System.out.printf("Node %s is shutting down due to %s...\n", this.self(), reason);
```
jakrawcz-rdx, line 181: _(note from author)_

This is a drive-by fix: for many tests, this line was throwing Guice errors (since bean of `Key.get(BFTValidatorId.class, Self.class)` is not bound).

core/src/main/java/com/radixdlt/p2p/transport/PeerServerBootstrap.java
```diff
@@ -160,15 +161,17 @@ public void start() {
...
     log.info("P2P server started. Node URI is: {}", self);
   }
 
   public void stop() {
-    if (serverBind != null) {
-      serverBind.channel().close().syncUninterruptibly();
+    @Nullable var runningServerBootstrap = this.serverBootstrap.getAndSet(null);
+    if (runningServerBootstrap != null) {
+      runningServerBootstrap.config().childGroup().shutdownGracefully().syncUninterruptibly();
+      runningServerBootstrap.config().group().shutdownGracefully().syncUninterruptibly();
```
jakrawcz-rdx, line 174: _(note from author)_

This fixes the runaway Netty pool problem found at https://github.com/radixdlt/babylon-node/pull/550#issuecomment-1638293010

core/src/main/java/com/radixdlt/modules/ModuleRunner.java
```diff
@@ -64,10 +64,15 @@
...
+
 /** Manages a module */
 public interface ModuleRunner {
-  /** Start running the module */
-  void start();
+  /**
+   * Start running the module. The given error handler will be notified each time an unhandled
+   * exception escapes the module's worker thread.
+   */
+  void start(Consumer<Throwable> errorHandler);
```
jakrawcz-rdx, line 75: _(note from author)_

This is a place which I consider hacky - basically this provides an "external" `appKiller` dependency to each worker.

More elegant ideas I had:
- Have an internal, well-defined "App" bean, which holds a registry of all running workers, and a shutdown method, and can be injected (i.e. build-in the app framework inside Guice).
- Keep the appKiller external, but migrate the worker interface from `start()+stop()` to `runAsync() -> CompletableFuture<Void>` (and operate on proper `.cancel()/.await()/whatever` primitives, in a generic way, outside of Guice)

Both of these approaches seemed like "too much work" in the light of "we want to ditch Java" fact though :(

core/src/main/java/com/radixdlt/RunningRadixNode.java
dhedey, line 181: So, in the `RadixNodeApplication`, we instead handle this by doing:
```
      log.fatal("Unable to start", ex);
      LogManager.shutdown(); // Flush any async logs
```

Before then killing with `java.lang.System.exit(-1);`

Perhaps we should do something similar to improve our logging here?

core/src/main/java/com/radixdlt/RunningRadixNode.java
```diff
@@ -118,7 +124,11 @@ public static RunningRadixNode run(UnstartedRadixNode unstartedRadixNode) {
 
     for (var module : moduleStartOrder) {
       final var moduleRunner = moduleRunners.get(module);
-      moduleRunner.start();
+      moduleRunner.start(
+          error -> {
+            log.error("Unhandled exception in runner {}; shutting down the node", module, error);
+            runningNode.shutdown();
```
dhedey, line 130: Hmm this no longer actually exits the process? But potentially leaves the main thread around doing nothing?

Perhaps we should actually do an exit in the `shutdown()` method? Or else just trigger `Runtime.shutdown()` in there or something? (which I imagine would then run the shutdown hook?

dhedey (approved):
Awesome! Thanks for deep diving on this - it's looking much better!

A few comments of varying importance

core/src/main/java/com/radixdlt/RunningRadixNode.java
jakrawcz-rdx: > Hmm this no longer actually exits the process? 

This exits the process (I tested it from IntelliJ and terminal in some local experiments), by ensuring that all threads started (by `RunningRadixNode.run()`) are stopped (by `RunningRadixNode.shutdown()`), which is a correct way of spinning down a service (contrary to `System.exit()`, which was causing us problems).

> But potentially leaves the main thread around doing nothing?

The main thread is probably long dead when all this happens.
(because `bootstrapRadixNode()` is effectively async - it starts a few threads belonging to the Node and returns)

dhedey: > This exits the process (I tested it from IntelliJ and terminal in some local experiments), by ensuring that all threads started (by RunningRadixNode.run()) are stopped (by RunningRadixNode.shutdown()), which is a correct way of spinning down a service (contrary to System.exit(), which was causing us problems).

But this assumes that we have properly shut down all threads. Otherwise if we introduce a regression in future, we could have a zombie node which doesn't restart automatically, which seems scary to me?

It feels weird that we use the same method for "on shutdown" hook as we do for "please shutdown" - I feel like we should instead tell the JVM "please shut me down", which will then try to gracefully shut down - but also fall back to  non-graceful shutdown if it has to - (which I believe is `System.Exit()`?)

core/src/main/java/com/radixdlt/RunningRadixNode.java
dhedey: Ahh okay, understood, fair enough

core/src/main/java/com/radixdlt/environment/rx/RxModuleRunnerImpl.java
```diff
@@ -195,7 +183,7 @@ public String threadName() {
   }
 
   @Override
-  public void start() {
+  public void start(Consumer<Throwable> errorHandler) {
```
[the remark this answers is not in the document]
dhedey: Interesting... yay for bad online docs :D. Thanks for improving our docs :)

dhedey (approved):

core/src/main/java/com/radixdlt/RunningRadixNode.java
jakrawcz-rdx: hmm, a `System.exit()` called by the "on exception" business callback was the cause of other problems, so I wanted to avoid it...

But I totally see what you want to ensure. Is it feasible to have an integration test that starts an actual process (rigged so that some fatal exception happens) and asserts on the process terminating itself? Would it make this PR less scary?
_(TBH a positive test counterpart [i.e. assert the process starts and stabilizes fine] would also be nice)_

But if not feasible (or the required infra would not fit in this PR, or within priorities), then I guess I can switch to the "request System.exit() and spin yourself down in a shutdown hook" approach.

jakrawcz-rdx: Alright, I realized it is very much ok to go this way, since `RunningRadixNode` is not used by any deterministic/simulation int-tests (only by prod).

So, we no longer risk `... finished with non-zero exit value 255` from any tests (in the most recent commit I also remove this warnings from comments).

PTAL.

core/src/main/java/com/radixdlt/RunningRadixNode.java
```diff
@@ -156,28 +150,9 @@ public void reportSelfStartupTime(Duration startupTimeMs) {
...
-    }
-    doShutdown(reason);
-  }
-
-  private void shutdownFromHook() {
-    doShutdown("a requested shutdown (via the Java shutdown hook)");
-  }
-
-  private void doShutdown(String reason) {
+  public void shutdown() {
```
dhedey, line 153: Now that this is only called by the shutdown hook, I have a slight preference to call this `onShutdown` or stick a JavaDoc to explain this is registered as the shutdown hook

dhedey (approved):
Nice :)

core/src/main/java/com/radixdlt/RunningRadixNode.java
dhedey: Also happy with creating an integration test to ensure the shutdown hook does shut down all threads, but yes - this is only used in prod

dhedey: Do the tests also use the shutdown hook? Or do their own thing?

jakrawcz-rdx: > integration test to ensure the shutdown hook does shut down all threads

Actually not shutdown hook, but JVM is supposed to halt all threads after all shutdown hooks run. I'm gonna trust the JVM :p

(please note how JVM can still hang if e.g. some shutdown hook does NOT return :D ahh the good old Halting Problem)

> Do the tests also use the shutdown hook? Or do their own thing?

Own thing: 
https://github.com/radixdlt/babylon-node/blob/df358cd19debff098855e8a031828ed1f678b7a3/core/src/test-core/java/com/radixdlt/harness/simulation/network/SimulationNodes.java#L262-L263
(that's why I could fulfill your wish without breaking the elegant behavior of tests)

PR #560, opened 2023-07-20
Refactor metrics

malexandru-rdx:
In this PR:
- Move metrics infra into node-common.
- Move mempool metrics under mempool/
- Move metrics update logic out of state_manager.rs
- Split MempoolMetrics and move part of it in the mempool itself
- Fix missed mempool submission rejection metrics

### DIFF
--- a/core-rust/state-manager/src/mempool/priority_mempool.rs
+++ b/core-rust/state-manager/src/mempool/priority_mempool.rs
@@ -63,6 +63,8 @@
  */
 
 use node_common::config::MempoolConfig;
+use node_common::metrics::TakesMetricLabels;
+use prometheus::Registry;
 use rand::seq::index::sample;
 use tracing::warn;
 use transaction::model::*;
@@ -74,6 +76,8 @@ use std::cmp::{max, min, Ordering};
 use std::collections::{BTreeSet, HashMap, HashSet};
 use std::sync::Arc;
 use std::time::Instant;
+
+use super::metrics::MempoolMetrics;
 
 #[derive(Clone, PartialEq, Eq)]
 pub struct MempoolData {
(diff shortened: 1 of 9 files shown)


core-rust/state-manager/src/mempool/priority_mempool.rs
```diff
@@ -299,8 +313,15 @@ impl PriorityMempool {
         let payload_hash = &data.transaction.notarized_transaction_hash();
         let intent_hash = &data.transaction.intent_hash();
 
+        let transaction_size = data.transaction.raw.0.len();
         self.remaining_transaction_count += 1;
-        self.remaining_total_transactions_size += data.transaction.raw.0.len() as u64;
+        self.remaining_total_transactions_size += transaction_size as u64;
+
+        // Update metrics
```
jakrawcz-rdx, line 320: I looked at the entire priority_mempool.rs and I am pretty convinced that we track all (2) add/remove operations.

But ofc, as always, I worry that a future refactor/feature will introduce another add/remove call, and we would miss a metric update.

Actually the same problem is with `proposal_priority_index` (well, probably harder to miss, since it would cause weird behavior right away, unlike metrics).

What do you think about introducing a `ListenableMap<K, V>`, which would allow for installing a `MapListener<K, V>`, which would get notified like `.onEntryAdded(K, V) / .onEntryRemoved(K, V)`?
(and using it for maintaining the `proposal_priority_index` and for the metric-bumping)

core-rust/state-manager/src/transaction/preview.rs
```diff
@@ -150,7 +150,9 @@ mod tests {
             &logging_config,
             FeeReserveConfig::default(),
         ));
-        let pending_transaction_result_cache = Arc::new(parking_lot::const_rwlock(
+        let pending_transaction_result_cache: Arc<
+            parking_lot::lock_api::RwLock<parking_lot::RawRwLock, PendingTransactionResultCache>,
+        > = Arc::new(parking_lot::const_rwlock(
```
jakrawcz-rdx, line 155: unintentional change?

jakrawcz-rdx (commented):
This is a very good cleanup 👍 

I have one suggestion regarding future-proofing.

Can I also ask for pasting here some prometheus endpoint dump showing the [preferably non-0] values of the touched metrics? _(just as a kind of one-shot manual test, since metrics are rarely tested programmatically)_

core-rust/state-manager/src/transaction/preview.rs
malexandru-rdx: Totally unintentional, `spotless` decided it should do it.

core-rust/state-manager/src/mempool/priority_mempool.rs
malexandru-rdx: I think it makes more sense to keep it as is and in future PRs touching this remind people to use only `remove_data` and `add_transaction`. There is also `intent_lookup` that needs maintaining.
If another way of adding transactions (without checking for priorities, sizes, etc.) will make sense we can split `add_transaction` at that point and keep only 2 "primitive/internal" methods that deal with all aux data/metrics.

malexandru-rdx:
This is how the metrics look:
Screenshot 2023-07-26 at 15 18 32

jakrawcz-rdx (approved):
Cool refactor, and metrics seem to still work, so 🟢 :D

PR #563, opened 2023-07-21
Do not recover from Core API panic when any lock is held for writing

jakrawcz-rdx:
As promised in https://github.com/radixdlt/babylon-node/pull/554#issuecomment-1640525523.

From this PR on, we have a locks facade that we should use instead of parking_lot.
(to make it easier, I removed the direct dependency to parking_lot from most projects)


jakrawcz-rdx:
I introduced a `mempool.write() + panic!" to my favorite endpoint. Here is what logs look like:

```
...
# 2023-07-21T08:15:33,725 [INFO/EventLoggerModule/BFT no...64vp9cx0t] (EventLoggerModule.java:166) - bft_nxtrnd{epoch=34 round=593 leader=no...64vp9cx0t nextLeader=no...64vp9cx0t}
thread 'tokio-runtime-worker' panicked at 'jeb!', core-api-server/src/core_api/handlers/mempool_list.rs:11:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/39f2657d1101b50f9b71ae460b762d330cc8426b/library/std/src/panicking.rs:579:5
   1: core::panicking::panic_fmt
             at /rustc/39f2657d1101b50f9b71ae460b762d330cc8426b/library/core/src/panicking.rs:64:14
             at /Users/jakrawcz/repo1/babylon-node/core-rust/core-api-server/src/core_api/handlers/mempool_list.rs:11:9
  ... // snipped
 127: tokio::runtime::blocking::pool::Spawner::spawn_thread::{{closure}}
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
2023-07-21T06:15:34.524189Z ERROR node_common::locks: a node_common::locks::RwLockWriteGuard<state_manager::mempool::priority_mempool::PriorityMempool> was released while panicking; aborting the process

> Task :core:run FAILED
```

The user unfortunately gets `curl: (52) Empty reply from server` and it's not so easy to improve it.

core-rust/node-common/src/locks.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+}
+
+fn abort_if_panicking<T>(_guard: &T) {
+    if thread::panicking() {
+        error!(
+            "a {} was released while panicking; aborting the process",
+            type_name::<T>()
+        );
+        abort();
```
malexandru-rdx, line 210: Can't the abort happen while writing a fresh commit to db?

I think it's a bit extreme, we should at least wait for any ongoing write.

malexandru-rdx (commented):
This is "transactional" only for 1 lock - a better mechanism is needed for more complicated flows. Until now, core-api was not possible to corrupt ledger data (at least this way, which seems reasonable and probable).

core-rust/node-common/src/locks.rs
jakrawcz-rdx: As discussed on Slack:

Looks like the best alternative to the (indeed extreme) `abort()` is to call a JNI `System.exit()`. (it will relatively-gracefully clean up java and rust resources via our shutdown-hook).
Let me try to achieve it.

jakrawcz-rdx: Alright, I am able to call a `System.exit(-1)` from a `rwGuard.drop()` (instead of `abort()`).

However, this is what happens:
```
thread 'tokio-runtime-worker' panicked at 'some bug', core-api-server/src/core_api/handlers/mempool_list.rs:11:13
stack backtrace:
   0: rust_begin_unwind
             at /rustc/39f2657d1101b50f9b71ae460b762d330cc8426b/library/std/src/panicking.rs:579:5
   1: core::panicking::panic_fmt
             at /rustc/39f2657d1101b50f9b71ae460b762d330cc8426b/library/core/src/panicking.rs:64:14
             at /Users/jakrawcz/repo1/babylon-node/core-rust/core-api-server/src/core_api/handlers/mempool_list.rs:11:13
   ... // snipped
 127: tokio::runtime::blocking::pool::Spawner::spawn_thread::{{closure}}
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
2023-07-26T11:30:35.900642Z ERROR node_common::locks: a node_common::locks::RwLockWriteGuard<state_manager::mempool::priority_mempool::PriorityMempool> was released while panicking; invoking System.exit()
Node 833d72:03fff9 is shutting down... // our shutdown hook, nice
Pool BFT no...64vp9cx0t did not terminate // whoops
```

... and nothing more, the process hangs:
image

Indeed, the `BFT no...64vp9cx0t` did not terminate:
image

(the java `.shutdownNow()` could not interrupt it, since it hangs inside the Rust commit)
Only the `kill -9` works.

Is there a way to investigate where exactly in `commit()` we hang?
(so far I've learnt that `rust-gdb` is not available on m1)

I suspect waiting for a lock that is held by some thread which did not release it while unwinding... (although it is supposed to). I know it is not the lock on which I triggered the `System.exit(-1)` (my bug uses a `RwLock::new()`).

jakrawcz-rdx: ahhh it's more complicated: depending on timing, sometimes the BFT pool shutdowns fine.

However, in such case, Rust will hang here:

image

(that's our shutdown hook)

That cleanup method only does:
```
let jni_state_manager: JNIStateManager = env
            .take_rust_field(j_state_manager, POINTER_JNI_FIELD_NAME)
            .unwrap();
```
... so it's hanging inside the `jni` lib :(

jakrawcz-rdx: ok, my mad debugging skillz ("add lots of logging") led me to believe that the thread hangs when dropping the `JNIStateManager` instance (it prints a log from inside its `drop()`, but does not print the log from the very next line, consistently).

Still, I have no idea how to progress.

jakrawcz-rdx: For curious: here is my current state, for easy repro of the "app hangs during cleanup after Rust-induced System.exit()". I welcome any help.

The branch:
https://github.com/radixdlt/babylon-node/compare/refactor/abort_on_broken_synchro...experiment/access_jvm_statically?expand=1

1. Run Single Validator.
2. `curl localhost:3333/core/mempool/list -H 'Content-Type: application/json' -d '{"network":"bomb"}' -v`
3. Observe the process shutting down very orderly, but then hanging just before `println!("field dropped");`
4. `jps` + `kill -9 ...` to kill it

jakrawcz-rdx: The "shutdown hook's thread hangs" problem mentioned above can be fixed by triggering an async shutdown; SOLVED 👍 

However, there is another problem: after extensive experimenting with @malexandru-rdx I must conclude that there is no reliable way of statically obtaining a reference to a running JVM across different JVM versions, JNI versions and rustc versions - it can always happen that someone has a set-up on which the JNI native `JNI_GetCreatedJavaVMs()` returns invalid pointer, which results in SIGBUS or SIGSEGV :(

This means that we no longer can have static lock constructors, but rather a managed `LockFactory`, which is configured with an "app-killer" callback (which will do the most regular `JNIEnv.call(async_exit)`).
Maybe it's even more elegant, since e.g. in tests we do NOT want to do `System.exit()` (i.e. we would prefer to have a "test-fail'er" callback there, not "app-killer" - in order to capture proper test results).

Anyways, that's ☝️ my design and I know how to progress (it's just a bit more code changes).

malexandru-rdx: > I must conclude that there is no reliable way of statically obtaining a reference to a running JVM

I have managed to fix it, on the `experiment/` branch. I think the problem was a mismatch between multiple JVM libraries.

> This means that we no longer can have static lock constructors, but rather a managed LockFactory, which is configured with an "app-killer" callback (which will do the most regular JNIEnv.call(async_exit)).
Maybe it's even more elegant, since e.g. in tests we do NOT want to do System.exit() (i.e. we would prefer to have a "test-fail'er" callback there, not "app-killer" - in order to capture proper test results).

As discussed on Slack, this is a bit too hacky (or too much unnecessary complication) for Rust side - I'd prefer to not let testing from Java interfere with production/arhitecture in Rust at the expense of Java code's quality.

jakrawcz-rdx: Alright, the newest commit introduced the change requested above - invoking java `System.exit()` instead of `abort`.
(credits to @malexandru-rdx for helping with rust build to include jni-sys properly).

Worth noting: after this change, the "shutdown" is reaaaaaallly asynchronous. (and it's not even because of our async `System.exit()` call - the shutdown hooks themselves are always async by themselves)
This means that theoretically, a couple of BFT rounds can go through before the relevant Java/Rust services are stopped. And a couple of Core API calls too.

jakrawcz-rdx:
@malexandru-rdx despite running fine locally (on your machine and mine), the build has problems on CI:

```
thread 'main' panicked at 'Could not locate JVM library: JavaLocatorError { description: "Java is not installed or not added in the system PATH" }', jni-export/build.rs:67:10
```

And:
```
/opt/hostedtoolcache/Java_Zulu_jdk/17.0.8-7/x64/bin/java: symbol lookup error: /runner/_work/babylon-node/babylon-node/core-rust/target/debug/libcorerust.so: undefined symbol: JNI_GetCreatedJavaVMs
```

malexandru-rdx (commented):
Nice! This is better than `abort()` or continue business as usual so an improvement IMO.
A future (possible) improvement would be to also signal Rust state_manager to not write anything further right away (since database corruption is what we don't want) - but I don't know how nicely that will play out with other Java components relying on it.

Anyway, we need to fix CI build first - I think the rust-build docker does not install JDK.

jakrawcz-rdx:
> also signal Rust state_manager to not write anything further right away (since database corruption is what we don't want)

I've been thinking about a very related question earlier:
"How will this `if (panicking) jni.systemExit()` code have to change WHEN we ditch Java and become Rust-only?"

And in light of this, I actually liked the `LockFactory::new(killerCallback)` (https://github.com/radixdlt/babylon-node/pull/563#discussion_r1276521354). Today, `killerCallback = jni.systemExit()`, and tomorrow `killerCallback = stateManager.shutdown()`. And in tests, `killerCallback = {}` (that's also needed today) - it all kinda "clicks".
Are you sure we shouldn't pursue this approach?

jakrawcz-rdx:
Closing this in favor of https://github.com/radixdlt/babylon-node/pull/577.

PR #566, opened 2023-07-21
Comparing only the stateVersion in a pre-check of ledger extension to commit

jakrawcz-rdx:
This avoids the "Silently re-request useless LedgerSync from the same non-matching-ledger Node" (see https://rdxworks.slack.com/archives/C01M90Y2448/p1689936487103519?thread_ts=1689864950.471299&cid=C01M90Y2448)


jakrawcz-rdx:
That comparator impl is probably not needed at all right now (used only in tests), but let's address this later, in develop.

core/src/main/java/com/radixdlt/ledger/StateComputerLedger.java
```diff
@@ -326,7 +327,11 @@ private void commit(LedgerExtension ledgerExtension, VertexStoreState vertexStor
     synchronized (this.commitAndAdvanceLedgerLock) {
       final LedgerProof againstLedgerHeader = this.currentLedgerHeader;
 
-      if (this.headerComparator.compare(nextHeader, againstLedgerHeader) <= 0) {
+      if (nextHeader.getStateVersion() <= againstLedgerHeader.getStateVersion()) {
+        log.trace(
+            "ignoring the ledger extension {} which would not progress the current ledger {}",
```
dhedey, line 332: Ignoring*

dhedey (approved):
Want to check that the tests pass ideally, just in case...

core/src/main/java/com/radixdlt/ledger/StateComputerLedger.java
jakrawcz-rdx: ok, let me wait with this push (let's not cancel the CI)

jakrawcz-rdx:
Only the `MultiNodeRebootTest` failed.
I pushed the typo change.
Feel free to merge whenever.

PR #569, opened 2023-07-23
Synchronizing access to the LRUCache (high-priority peers) in hopes of avoiding NPEs

jakrawcz-rdx:
Addresses the **Fix this P2P race condition causing node crash** item of https://radixdlt.atlassian.net/browse/NODE-561.

### DIFF
--- a/core/src/main/java/com/radixdlt/p2p/addressbook/HighPriorityPeerList.java
+++ b/core/src/main/java/com/radixdlt/p2p/addressbook/HighPriorityPeerList.java
@@ -1,0 +1,126 @@
+/* Copyright 2021 Radix Publishing Ltd incorporated in Jersey (Channel Islands).
+ *
+ * Licensed under the Radix License, Version 1.0 (the "License"); you may not use this
+ * file except in compliance with the License. You may obtain a copy of the License at:
+ *
+ * radixfoundation.org/licenses/LICENSE-v1
+ *
+ * The Licensor hereby grants permission for the Canonical version of the Work to be
+ * published, distributed and used under or by reference to the Licensor’s trademark
+ * Radix ® and use of any unregistered trade names, logos or get-up.
+ *
+ * The Licensor provides the Work (and each Contributor provides its Contributions) on an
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/radixdlt/p2p/addressbook/HighPriorityPeerList.java
```diff
@@ -0,0 +1,126 @@
...
+  private final LRUCache<NodeId, Unit> nodeIds;
+
+  /**
+   * Creates a list with the given maximum size and initial contents. Note: the initial contents may
+   * get truncated (from the beginning) if larger than maximum size.
+   */
+  public HighPriorityPeerList(int maxSize, List<NodeId> initialNodeIds) {
+    this.nodeIds = new LRUCache<>(maxSize);
+    for (var nodeId : initialNodeIds) {
+      Preconditions.checkNotNull(nodeId);
```
jakrawcz-rdx, line 94: _(note from author)_

I'd be very surprised if the `null`s came from persistence; yet I figured this check won't harm us (and it will narrow down the analysis if NPEs happen again).

core/src/main/java/com/radixdlt/p2p/addressbook/HighPriorityPeerList.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+  }
+
+  /**
+   * Reports the usage of the given peer and returns a new (LRU-ordered) list of priority peers (or
+   * empty if the contents of this list was not changed by this call - i.e. if the given peer was
+   * already exactly the least-recently-used one).
+   */
+  public Optional<List<NodeId>> reportUsageAndGetNewHighPriorityPeers(NodeId nodeId) {
+    Preconditions.checkNotNull(nodeId);
```
jakrawcz-rdx, line 105: _(note from author)_

I'd be very surprised if the `null`s came from the outside; yet I figured this check won't harm us (and it will narrow down the analysis if NPEs happen again).

malexandru-rdx (approved):
LGTM

dhedey (approved):
Really nice 👍

PR #570, opened 2023-07-25
Refactor: Driving invalid response sync handling by explicit exceptions

jakrawcz-rdx:
These are refactors and cleanups mentioned in https://radixdlt.atlassian.net/browse/NODE-561 (excluding the [concurrency issue](https://github.com/radixdlt/babylon-node/pull/569)).


core/src/main/java/com/radixdlt/sync/LocalSyncService.java
```diff
@@ -428,68 +418,137 @@ private boolean isFullySynced(SyncState.SyncingState syncingState) {
...
+      this.peerControl.reportHighPriorityPeer(sender);
+      this.syncLedgerUpdateTimeoutDispatcher.dispatch(
+          SyncLedgerUpdateTimeout.create(currentState.getCurrentHeader().getStateVersion()), 1000L);
+      return currentState.clearPendingRequest();
+    } catch (InvalidSyncResponseException isre) {
+      // Implementation note:
+      // Java 17 still cannot compile a "complete catch" of all subclasses of a sealed exception.
+      // But it can compile a "complete switch", and we want to leverage this in order to ensure
+      // that all future corner-cases of response processing are handled here as well.
+      return switch (isre) {
```
jakrawcz-rdx, line 434: _(note from author)_

As you can see below, I went for a rich and explicit exception hierarchy to handle different outcomes of sync response processing.
The `LocalSyncService` cares about maintaining the sync state machine and its infra (e.g. rewarding/penalizing peers, bumping metrics).
The extracted `SyncResponseHandler` cares only about validating and committing the transactions from the response.

You may ask "why validation vs commit cannot be separate responsibilities?".
That's because a critical part of validation (e.g. transaction parsing, ledger hashes verification) is done by a downstream (Rust) state manager and would be rather hard to pull out from there. So I kinda treat "commit or throw if anything was invalid" as a single operation, which can either return nothing (== success) or throw some very specific exception (so that the caller can precisely drive the sync state's machine reaction).

You may also ask "do we need exceptions? can't we return something like `SyncResponseHandlingResult` enum which summarizes the outcome?".
Currently, for technical reasons (i.e. the `EventDispatcher<>` interface, which we use everywhere, even for 100% sync calls), it wouldn't be so easy (we would need a refactor: migrate the "committer" to a plain sync service).
I also considered a hardcore checked `Exception` (since in a BFT world, invalid response is more of a "contingency" than "unexpected error"), but I know many people hate them.

core/src/main/java/com/radixdlt/sync/LocalSyncService.java
```diff
@@ -428,68 +418,137 @@ private boolean isFullySynced(SyncState.SyncingState syncingState) {
...
+              "LocalSync: Received mismatched signatures in sync response {} from {}",
+              syncResponse,
+              sender);
+          metrics
+              .sync()
+              .invalidResponsesReceived()
+              .label(new InvalidSyncResponse(InvalidSyncResponseReason.MISMATCHED_SIGNATURES))
+              .inc();
+          peerControl.banPeer(sender, PEER_BAN_DURATION, "mismatched signatures received");
+          yield this.processSync(currentState.clearPendingRequest().removeCandidate(sender));
```
jakrawcz-rdx, line 548: _(note from author)_

The listing above is lengthy and treats every possible problem separately.
In theory, this could be rolled into 2 major groups: "problems that are probably only caused by a random timeout" vs "problems that strongly indicate a malicious/faulty peer, which should be banned".
I tried this approach, but logging and metric-bumping became a bit cumbersome then.
Hence, the explicit handling seen above.

malexandru-rdx (approved):
Nice!

core/src/main/java/com/radixdlt/sync/SyncResponseHandler.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    final var pendingRequestPeer = optPendingRequestPeer.get();
+    if (!pendingRequestPeer.equals(sender)) {
+      throw new InvalidSyncResponseException.SyncRequestSenderMismatch();
+    }
+
+    final var currentHeader = currentState.getCurrentHeader();
+    if (!ledgerExtensionStart.equals(currentHeader)) {
+      throw new InvalidSyncResponseException.LedgerExtensionStartMismatch();
```
dhedey, line 129: I believe this particular check is more specific than present - we used to accept through requests with a start not matching our current state. Out of interest, what motivated this addition? I guess you felt that this was missing before and needed adding? (It's not wrong, I'm just curious)

In fact I don't think the start proof actually matters (beyond saying "this is where you requested and where I start from").

There are probably some cases where ignoring this start makes things more efficient (eg BFT sync in the mean time) - but, that said, I can't think of any which would be too hurt by this check... 

I think on balance it's better to keep the check in, but interested to hear your thoughts.

dhedey (approved):
Nice! Some minor comments

core/src/main/java/com/radixdlt/sync/SyncResponseHandler.java
jakrawcz-rdx: Yup, I added this on purpose and here is my reasoning:

The whole "ledger sync" feature looks like an RPC use-case: there is a Request, there is a Response.
In fact, we have like ~80% impl of a poor-man's RPC infra here: there is tracking of pending exchanges (request-response pairs), there is rejection of unsolicited responses, there is even "request ID" (but used only for timeout-tracking).

A true RPC infra would actually transmit that request ID to the peer (and expect it in the matching response). Alternatively, it could expect a `hash(request)` in the matching response. Ooooor, alternatively (and a bit wastefully), it could expect a verbatim `request` in the matching response.

So I thought "well, we have the `request` already echoed back in the response, let's expect it to match".
There is no reason (for a well-behaving peer) to return any older start header - if it happens, then most likely our state changed in the meantime and we are no longer expecting this particular sync response.
I prefer to match (or fail-fast) and later in the code have a guarantee that "yes, this is the response to exactly my request, it arrived in time, and I am still in the perfect state to process it as-is".

> There are probably some cases where ignoring this start makes things more efficient

I agree, totally - but I don't think it's worth optimizing for "making maximum use of corner-cased/timed-out/unsolicited sync responses".

And an argument from a different angle:
Please note that the same is true about "ignoring the sender of the response" (I mean: if some other peer just sends you a nice, useful, correctly-signed ledger extension out of nowhere, then why not apply it? Yet, we always match on the sender [even before this PR]).

core/src/main/java/com/radixdlt/ledger/InvalidCommitRequestException.java
```diff
@@ -71,7 +71,14 @@
  * consistent (e.g. non-parseable or non-committable transactions, ledger proof mismatches).
  */
 public class InvalidCommitRequestException extends RuntimeException {
+
+  private final InvalidCommitRequestError error;
+
   public InvalidCommitRequestException(InvalidCommitRequestError error) {
-    super(error.toString());
+    this.error = error;
```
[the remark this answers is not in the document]
jakrawcz-rdx: No magic is that powerful...
I probably deleted this accidentally (and never noticed, since we don't log this exception - just translate its `error` to a higher-level exception). Lemme bring it back, it won't hurt 👍

jakrawcz-rdx:
Huh, this repeats on CI:
```
com.radixdlt.api.core.NetworkConfigurationTest

  Test test_network_configuration FAILED

  com.radixdlt.api.core.generated.client.ApiException: java.net.ConnectException
      at app//com.radixdlt.api.core.NetworkConfigurationTest.test_network_configuration(NetworkConfigurationTest.java:81)
  Caused by: java.net.ConnectException
      at com.radixdlt.api.core.NetworkConfigurationTest.test_network_configuration(NetworkConfigurationTest.java:81)
  Caused by: java.net.ConnectException

  Caused by: java.nio.channels.ClosedChannelException
```
...but never on my machine O_o

core/src/main/java/com/radixdlt/sync/SyncResponseHandler.java
dhedey: Sounds good to me

PR #572, opened 2023-07-27
New metrics: Node's wallclock and proposer timestamp from last commit

jakrawcz-rdx:
The 2 most obvious metrics that will be used for future higher-level health metrics (see https://whimsical.com/node-status-and-health-metrics-QXjDREqeRDvGDu6xg3J8wx)

### DIFF
--- a/common/src/main/java/com/radixdlt/monitoring/Metrics.java
+++ b/common/src/main/java/com/radixdlt/monitoring/Metrics.java
@@ -262,6 +262,7 @@ public record Metrics(
   public record Misc(
       TypedInfo<Config> config,
       Timer nodeStartup,
+      GetterGauge wallclockEpochSecond,
       Counter vertexStoreSaved,
       GetterGauge peerCount) {}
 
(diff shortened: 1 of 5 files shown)


jakrawcz-rdx:
The new prometheus endpoint excerpt:
```
# HELP rn_misc_wallclock_epoch_second rn_misc_wallclock_epoch_second
# TYPE rn_misc_wallclock_epoch_second gauge
rn_misc_wallclock_epoch_second 1.690489632233E9
...
# HELP rn_ledger_last_update_proposer_epoch_second Proposer timestamp from the last proof written to the ledger.
# TYPE rn_ledger_last_update_proposer_epoch_second gauge
rn_ledger_last_update_proposer_epoch_second 1690489632.195
```
(note the different floating point renderings between java and rust metrics - they are both valid and accepted by Prometheus, which uses Go's syntax: https://go.dev/ref/spec#Floating-point_literals)

malexandru-rdx (approved):
LGTM!

PR #574, opened 2023-07-31
New item in info metric: Post-genesis epoch state hash

jakrawcz-rdx:
Another small metric change (after https://github.com/radixdlt/babylon-node/pull/572): Post-genesis epoch state hash (see https://whimsical.com/node-status-and-health-metrics-QXjDREqeRDvGDu6xg3J8wx)


core/src/main/java/com/radixdlt/monitoring/InMemorySystemInfo.java
```diff
@@ -66,48 +66,22 @@
...
-  // private final RadixEngine<LedgerAndBFTProof> radixEngine;
 
   @Inject
-  public InMemorySystemInfo(
-      @LastProof LedgerProof lastProof,
-      @LastEpochProof LedgerProof lastEpochProof,
-      SelfValidatorInfo self
-      /*RadixEngine<LedgerAndBFTProof> radixEngine*/ ) {
-    this.ledgerProof = new AtomicReference<>(lastProof);
+  public InMemorySystemInfo(@LastEpochProof LedgerProof lastEpochProof) {
```
jakrawcz-rdx, line 83: _(note from author)_

As a drive-by refactor, I have trimmed down this `SystemInfo` thing a bit.
Is anyone else under the impression that it should really vanish completely?

dhedey: A lot of the stuff here is related to the old Olympia version of the "own missedProposals" metric which I think you've handled elsewhere?

I don't have a strong feeling about whether this should still exist or not. Feels like an okay abstraction to keep, or move into something else.

core/src/main/java/com/radixdlt/monitoring/MetricInstaller.java
```diff
@@ -90,10 +94,12 @@ public final class MetricInstaller {
   @Inject
   public MetricInstaller(
       final SelfValidatorInfo self,
+      final REv2LedgerInitializerToken ledgerInitialization,
       final InMemorySystemInfo inMemorySystemInfo,
       final PeersView peersView,
       final TimeSupplier timeSupplier) {
     this.self = self;
+    this.ledgerInitialization = ledgerInitialization;
```
dhedey, line 102: I wonder if this means we initialize metrics later than we used to? Maybe this isn't a problem? or previously it was kinda random if it initialized before genesis executed or not?

Or maybe the prometheus server didn't start until after genesis anyway?

dhedey (approved):

core/src/main/java/com/radixdlt/monitoring/MetricInstaller.java
jakrawcz-rdx: > Or maybe the prometheus server didn't start until after genesis anyway?

hah, the prometheus server is started literally as a last-but-one thing that happens: https://github.com/radixdlt/babylon-node/blob/5c88d9489af5527dd1684f97f309e1e11ac81be1/core/src/main/java/com/radixdlt/RunningRadixNode.java#L136

So yeah, the moment at which the (trivial) metrics' constructor is run does not matter here.

dhedey: Haha love it.

Although, just to point out, we somehow reliably observe the `BOOTING` state if something crashes between bootup and first DB read / commit (eg with the failing foundation validator last week) - but that would possibly also be true under this PR I imagine :+1: or maybe not, but the Babylon node corrupts less, so it's not a problem

PR #576, opened 2023-07-31
Adding metric for other validators' proposal timestamp difference

jakrawcz-rdx:
As part of https://whimsical.com/node-status-and-health-metrics-QXjDREqeRDvGDu6xg3J8wx, this PR adds:
```
# TYPE rn_bft_proposal_timestamp_difference_seconds gauge
rn_bft_proposal_timestamp_difference_seconds{key="03fff97bd5755eeea420453a14355235d382f6472f8568a18b2f057a1460297556",component_address="validator_loc1sv7h9eggskmgf7eat8jfd5pvsfx6tf728nwres7zae8694cp5xewdk",} -0.001
```

The interpretation of the above is of course:
"When that validator was creating the proposal, its wallclock was `-0.001` seconds away from my current wallclock."
We are aware that this "clock skew" actually includes the network delay too - but that's all we care about, since we accept/reject proposals based on this clock difference.

### DIFF
--- a/core/src/main/java/com/radixdlt/consensus/bft/processor/ProposalTimestampVerifier.java
+++ b/core/src/main/java/com/radixdlt/consensus/bft/processor/ProposalTimestampVerifier.java
@@ -64,6 +64,7 @@
 
 package com.radixdlt.consensus.bft.processor;
 
+import com.radixdlt.addressing.Addressing;
 import com.radixdlt.consensus.Proposal;
 import com.radixdlt.consensus.bft.*;
 import com.radixdlt.environment.EventDispatcher;
@@ -72,6 +73,7 @@ import com.radixdlt.monitoring.Metrics.RejectedConsensusEvent;
 import com.radixdlt.monitoring.Metrics.RejectedConsensusEvent.TimestampIssue;
 import com.radixdlt.monitoring.Metrics.RejectedConsensusEvent.Type;
 import com.radixdlt.utils.TimeSupplier;
+import java.time.Duration;
 import java.util.Objects;
 import java.util.Optional;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/radixdlt/consensus/bft/processor/ProposalTimestampVerifier.java
```diff
@@ -132,6 +139,15 @@ public void processProposal(Proposal proposal) {
     final var prevTimestamp = proposal.getVertex().parentLedgerHeader().proposerTimestamp();
     final var proposalTimestamp = proposal.getVertex().proposerTimestamp();
 
+    metrics
```
LukasGasior1, line 142: I wonder if we should filter out `self` here? It should always be near 0 and could skew the metrics if we e.g. average the diffs

LukasGasior1 (approved):
just a minor comment, LGTM overall

core/src/main/java/com/radixdlt/consensus/bft/processor/ProposalTimestampVerifier.java
LukasGasior1: but I guess we could as well just filter it out when we actually avg :+1:

jakrawcz-rdx: I could.

But we can also calculate `avg_diff_without_self = avg_diff * N/(N-1)`.
(where `N` = number of validators; assuming `self` is near 0 indeed)

Another interpretation of the above fact: the inclusion of `self` here only skews the `avg` metric by a factor of ~`1/N`. It does not skew other interesting metrics, like `max/min`.

And IF the `self`, for some reason, becomes significantly-non-`0`? Then I will appreciate it is in the metrics very much (and will probably go and investigate what causes the delay of proposal processing in the Node internally).

PR #577, opened 2023-08-01
Introducing a LockFactory with a configurable "fatal panic handler" callback

jakrawcz-rdx:
This is a successor of https://github.com/radixdlt/babylon-node/pull/563 (spinning off at an intermediate commit https://github.com/radixdlt/babylon-node/pull/563/commits/1e87af09c733fdb1d9fbc635ed353262f0151da2, after making a decision on the approach at https://github.com/radixdlt/babylon-node/pull/563#issuecomment-1657849799).

A "fatal panic" is a panic during which some write lock is held. We cannot allow e.g. tokio to catch such panic and recycle the thread (i.e. recover), since the state protected by the lock may already be inconsistent (i.e. aborted in the middle of some operation). I tried to capture this explanation in rustdocs and comments too.

The configurable callback allows us to:
- invoke a graceful shutdown in production
- call programmatic `shutdownNode()` variants in "integration" tests (which spin-up independent in-process Node instances)
- do nothing in unit tests (which will fail due to a `RustPanicException` propagated through JNI for any panic [not only a fatal panic, BTW])


jakrawcz-rdx:
Some "features" coming from this PR:

A **simulation** test report for "failure because of fatal panic":
image
_(valid failure, no process killed, java stack gives panic message)_

A **unit** test report for "failure because of any panic":
image
_(valid failure, no process killed, java stack gives panic message -  but this is unchanged, worked before this PR)_

A **Run Single Validator** logs on "failure because of fatal panic in Core API handler":
```
# 2023-08-01T23:02:24,002 [INFO/EventLoggerModule/BFT no...64vp9cx0t] (EventLoggerModule.java:166) - bft_nxtrnd{epoch=192 round=233 leader=no...64vp9cx0t nextLeader=no...64vp9cx0t}
thread 'tokio-runtime-worker' panicked at 'whoooooooooops!', core-api-server/src/core_api/handlers/mempool_list.rs:11:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/panicking.rs:67:14
             at /Users/jakrawcz/repo1/babylon-node/core-rust/core-api-server/src/core_api/handlers/mempool_list.rs:11:9
   ... // snipped
 147: tokio::runtime::blocking::pool::Spawner::spawn_thread::{{closure}}
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
2023-08-01T21:02:24.737464Z ERROR node_common::locks: a write guard of mempool was dropped while panicking; stopping
# 2023-08-01T23:02:24,738 [INFO/AsynchronousSystem/Thread-8] (AsynchronousSystem.java:81) - Starting a thread for asynchronous System.exit(-1)
2023-08-01T21:02:24.738523Z  WARN jni::wrapper::objects::global_ref: Dropping a GlobalRef in a detached thread. Fix your code if this message appears frequently (see the GlobalRef docs).    
# 2023-08-01T23:02:24,738 [INFO/AsynchronousSystem/Asynchronous-Exit] (AsynchronousSystem.java:86) - Invoking a System.exit(-1)
2023-08-01T21:02:24.738687Z  WARN core_api_server::core_api::server: path="/core/mempool/list" error=BasicErrorResponse { code: 500, message: "Unexpected server error", trace_id: None }
Node 833d72:03fff9 is shutting down...
```
(aaaaand the client gets `HTTP 500`, which is probably better than "connection closed")

jakrawcz-rdx:
One more thing:

As noted in https://github.com/radixdlt/babylon-node/pull/563#discussion_r1276861166, the behavior implemented here is still only a "best effort" avoidance of malformed state:

> after this change, the "shutdown" is reaaaaaallly asynchronous. (and it's not even because of our async `System.exit()` call - the shutdown hooks themselves are always async anyways)
This means that theoretically, a couple of BFT rounds can go through before the relevant Java/Rust services are stopped. And a couple of Core API calls too.

In this case, I wonder if we should _additionally_ poison the "fatally-panicked" lock?
Or maybe even all locks? (which is achievable when you have a central `LockFactory`)
Would this actually prevent _all_ possible "async access of malformed state", or only slightly improve the "best effort"?

@malexandru-rdx @dhedey Let's discuss ☝️ (and whatever the answer - I can address it in next PR)

malexandru-rdx (approved):
LGTM!

Next improvements could be: 

- call core-api shutdown as well
- guard all state manager entry points (panic at jni call if we already panicked with write-lock)

Right now an endpoint that crashes in write-lock is acting more like a "/core/shutdown" endpoint - trying to shutdown everything gracefully.

jakrawcz-rdx:
@malexandru-rdx 

> call core-api shutdown as well

But this will (eventually) happen with this PR's change already. Do you think it's any better if we trigger it synchronously right away? (I mean: it will still be async in nature, since there are already multiple threads which possibly just started processing some other requests)

> guard all state manager entry points (panic at jni call if we already panicked with write-lock)

yes, I assume we want it, and I thought about achieving (effectively) this via poisoning (see https://github.com/radixdlt/babylon-node/pull/577#issuecomment-1661586159), wdyt?

malexandru-rdx:
@jakrawcz-rdx I'm aware but not sure about the solution. I proposed this too (but probably not clear enough) in the first PR [here](https://github.com/radixdlt/babylon-node/pull/554#pullrequestreview-1530858082). However in the meantime I came to realise that it's still possible to cause weird corruption behaviour as it is not transactional - the poisoned lock might happen in the middle of an complex operation (that is _considered_ transactional, as `commit`).

I thought that this approach here implies not to take this route since lock poisoning means harder control of the "exit point" - also kind of diminishes the returns of this PR (why do we add extra Java-Rust infra just so that we crash on poisoning directly?).

I am not 100% about either method (weird stuff can happen in both cases if not taken into account), we just need to pick our poison :).

And to be completely honest, I feel like we are overthinking/overengineering at this point :) - currently only write lock in `core-api` is for mempool.

jakrawcz-rdx:
> And to be completely honest, I feel like we are overthinking/overengineering at this point

You are right, it cannot be fixed very reliably, especially with all the async-ness, indirection and java layer.

So let's think about the "best possible" solution in the future Rust-only impl:

We still won't get 100% safe, but I think we will be as close as possible to safety with the following policy:
- on "dropped lock while panicking", mark the entire application as "stopping"
  - in this state, the in-flight handlers of http requests and bft messages are allowed to finish just fine, as if nothing happened;
  - but any new http requests and bft messages are ignored; _(so, a classic "drain" behavior)_
  - and after all in-flights are handled, some cleaner thread closes all "resources".

This is achievable with the infra brought by this PR (i.e. swap the app-killer callback to another one, which implements the above policy). And poisoning is not required for this.

Hence, I can now answer [my own "should we poison?" question](https://github.com/radixdlt/babylon-node/pull/577#issuecomment-1661586159): No, not worth it, since it does not guarantee anything, and it is totally not needed in the longer run.

PR #578, opened 2023-08-02
Add the option to run the node using docker run instead of compose

balda-rdx:
This change puts the environment variables in a file and it also adds the option to run the node using docker run instead of docker compose


testnet-node/README.md
```diff
@@ -22,6 +22,13 @@ curl \
   -d '{ "network": "ansharnet" }'
 ```
 
+## Running without compose
+If you wanted to run without using compose you can do so like 
+```
+docker run -p 3333:3333 -v ledger-data:/home/radixdlt/RADIXDB -v key-data:/home/radixdlt/key --env-file radix-node.env radixdlt/babylon-node:rcnet-v2-phase2-r4
```
malexandru-rdx, line 28: While port `3334` (System API) and `3335` (Prometheus API) are not critical (and left to the user discretion), gossip port (`30000`) should be exposed (as a good practice mostly).
Of course, this doesn't make a difference for local development environments which are most probably behind a router with no port-forward configured :).

balda-rdx: @malexandru-rdx yeah good point

dhedey: Given this is recommended setup for developers, we should be careful to only bind on localhost, not all interfaces.

I believe we need to preface the binding with `127.0.0.1` to achieve that, although possibly `docker run` does this differently to `docker-compose`.

Also, I think it's worth us binding `3334` and `3335` at least, as suggested above. As developers may want to look at System API and Prometheus APIs.

PR #580, opened 2023-08-03
Adding metrics: Sync target proposer timestamp (untrusted) and Self validator component address (info tag)

jakrawcz-rdx:
These are 2 not-critical-but-useful metrics related to health that we discovered when discussing with Łukasz.

```
# TYPE rn_sync_target_proposer_epoch_second gauge
rn_sync_target_proposer_epoch_second 1.691069949372E9
…
# TYPE rn_misc_config_info gauge
rn_misc_config_info{…,key="03fff97bd5755eeea420453a14355235d382f6472f8568a18b2f057a1460297556",component_address="validator_loc1sv7h9eggskmgf7eat8jfd5pvsfx6tf728nwres7zae8694cp5xewdk",} 1.0
```

The `rn_sync_target_proposer_epoch_second` (when compared against wallclock) is a good indication of "is the Network that I am syncing to alive?". It accompanies the Sync target state version, and similarly cannot be trusted.

The `component_address=...` tag inside our info metric answers the question "is this Node configured to be a validator?" (i.e. empty when not validator). This is accompanied by the existing `rn_bft_in_validator_set` which kinda narrows down the "yes, validator" status.


common/src/main/java/com/radixdlt/monitoring/Metrics.java
```diff
@@ -372,5 +373,9 @@ public enum Direction {
     }
   }
 
-  public record Config(String branchAndCommit, String key, HashCode postGenesisEpochStateHash) {}
+  public record Config(
+      String branchAndCommit,
+      String key,
+      @Nullable String componentAddress, // null when not configured as validator
```
dhedey, line 379: Can we call this validatorAddress instead?

dhedey: Is this the statically configured (or genesis-resolved) validator address?

IE presence of this address doesn't actually apply the node is currently validating, right? 

Perhaps we should have another boolean metric for that, to help validators planning node maintenance?

common/src/main/java/com/radixdlt/monitoring/Metrics.java
```diff
@@ -234,6 +234,7 @@ public record LedgerSync(
       Counter remoteRequestsReceived,
       Gauge currentStateVersion,
       Gauge targetStateVersion,
+      Gauge targetProposerEpochSecond,
```
dhedey, line 237: I didn't understand what this was from just reading the name - I thought it was some diff between current and "target proposer" or something.

We call the thing being set the proposer timestamp. So perhaps we could call this metric `targetProposerTimestampEpochSecond`?

dhedey: Do you think we should add a comment that these two are untrusted?

dhedey (approved):

common/src/main/java/com/radixdlt/monitoring/Metrics.java
jakrawcz-rdx: > We call the thing being set the proposer timestamp. So perhaps we could call this metric `targetProposerTimestampEpochSecond`?

I wonder if anything else than a `Timestamp` can have a unit of `EpochSecond` :p
nevertheless, lemme apply this rename suggestion 👍 

> Do you think we should add a comment that these two are untrusted?

For external readers? I've put it in https://github.com/radixdlt/babylon-node/pull/582/files#diff-28b98263fd5ec3641cf14cb342ed3dd3077c88c19fad7692a35b4ec8f8e680e3R92 (and now I see it might be nice to link to that doc from `Metrics.java` too)
For internal readers? I think all metrics in this file could use descriptions... TBH it would be best defined as Prometheus `HELP` string. We do have a nice support for defining it in Rust, but not in Java :( for now, let me put this single code comment here 👍

common/src/main/java/com/radixdlt/monitoring/Metrics.java
jakrawcz-rdx: > Is this the statically configured (or genesis-resolved) validator address?

Yes.

> Perhaps we should call this configuredValidatorAddress?

Ok, done 👍 

> Perhaps we should have another boolean metric for "currently validating in epoch" under the consensus section, to help validators planning node maintenance? (if we don't already)

Yup, we have it as `rn_bft_in_validator_set`: https://github.com/radixdlt/babylon-node/blob/develop/common/src/main/java/com/radixdlt/monitoring/Metrics.java#L181

malexandru-rdx (approved):

common/src/main/java/com/radixdlt/monitoring/Metrics.java
dhedey: > I wonder if anything else than a Timestamp can have a unit of EpochSecond :p
I read it as TargetProposerEpoch (as in Radix Epoch) Second (as in some time difference). So I got confused because "Epoch" typically means "Ledger Epoch" not "Unix Epoch" to me when working in the node codebase :D 

RE other stuff - code comments all sound good.

PR #582, opened 2023-08-03
Docs: Guidelines for monitoring health of the Node

jakrawcz-rdx:
Based on https://whimsical.com/node-status-and-health-metrics-QXjDREqeRDvGDu6xg3J8wx

In this doc, I try to explain what metrics should be looked at when determining a (mutli-dimensional) "health status" of a Node, and how to interpret them.

I did not implement any convenience derived health-determining / dimension-aggregating Prometheus metrics. I generally frown upon computing a derived metric application-side - I think that PromQL is for deriving, unless the computation requires extensive business knowledge. And OFC this line is blurry, so: suggestions welcome :D 

(e.g. I know Łukasz liked the idea of 2-3 high-level computed "status metrics" + 1 single super-high-level "aggregated health": if you have more specific idea on how to define them, please share - and we will wonder whether it's easier to express it in PromQL or in Java :p)


core-rust/state-manager/src/mempool/priority_mempool.rs
```diff
@@ -254,9 +254,7 @@ impl PriorityMempool {
         // Check the new transaction is better than all to be removed transactions.
         if !to_be_removed.is_empty() {
             let best_to_be_removed = to_be_removed.last().unwrap();
-            if new_order_data
-                < MempoolDataProposalPriorityOrdering((*to_be_removed.last().unwrap()).clone())
-            {
+            if new_order_data < MempoolDataProposalPriorityOrdering(best_to_be_removed.clone()) {
```
jakrawcz-rdx, line 257: _(note from author)_

This is just a drive-by I spotted while reading lots of code related to our metrics.

docs/logging-and-monitoring/health-metrics.md
```diff
@@ -0,0 +1,170 @@
...
+  _Please note that such problems would be hard to unambiguously capture in the metrics anyway._
+
+### Ledger sync / commit status
+
+After the spin-up, one of the Node's primary goals is to keep its local Ledger "synced". Nodes
+constantly exchange their Ledger status information with one another and compare the following two
+metrics:
+
+rn_sync_current_state_version
+: The state version of the last transaction written to the local Ledger (either coming from a BFT
```
jakrawcz-rdx, line 38: _(note from author)_

This is a syntax for a "definition block" in markdown.
I now see that GH does not render it in any special way, but e.g. IntelliJ does.

docs/logging-and-monitoring/health-metrics.md
```diff
@@ -0,0 +1,170 @@
...
+  A state hash captured after the execution of all the Genesis Transactions. This should be constant
+  throughout the Node's lifetime, and the same across the entire Network. 
+- `configured_validator_address`
+  A logical Validator component address (within the Engine). Only present (i.e. non-empty) if it 
+  was statically configured or resolved from Genesis.
+
+rn_bft_in_validator_set
+: A "boolean" gauge, showing `1.0` when this Node belongs to an active Validator Set of the
+_current_ epoch (or `0.0` otherwise).
+The `1.0` here is only possible when `rn_misc_config_info.configured_validator_address` is
```
malexandru-rdx, line 149: AFAIU from here, you can not have a full node and create the validator & "promote" it while it is running, right?

malexandru-rdx (approved):
Looks good, just some minor questions.

I'm wondering if we should provide the health dashboard too. It seems a bit too technical & if not properly "tuned" the health status will still be noisy (basically now we provide the same raw signals as we were using before (+time) but we leave the interpretation to the reader :) ).

docs/logging-and-monitoring/health-metrics.md
jakrawcz-rdx: That is my understanding as well (I've once heard that you have to restart a full node for your validator settings to be applied - is this still the case @LukasGasior1 ?)

jakrawcz-rdx:
> I'm wondering if we should provide the health dashboard too. It seems a bit too technical & if not properly "tuned" the health status will still be noisy (basically now we provide the same raw signals as we were using before (+time) but we leave the interpretation to the reader :) ).

@malexandru-rdx 

I am not sure about providing an "official dashboard", but I [already mentioned](https://github.com/radixdlt/babylon-node/pull/582#issue-1835594468) something that deals with ~80% of this concern:

> I know Łukasz liked the idea of 2-3 high-level computed "status metrics" + 1 single super-high-level "aggregated health": if you have more specific idea on how to define them, please share

So: in order to leave the interpretation to the reader, we can define some computed higher-level indicator metric (e.g. `rn_health_is_ledger_progressing = true|false`). And then creating a nice dashboard with a couple of green/yellow/red LEDs (+ a master LED?) would be easy and unambiguous for node runners.

I just need good ideas for high-level health indicators. Preferably split by "domain". These can be booleans (enums don't really exist in Prometheus, they are 1-hot encoded) or maybe unit-less values (like: "something is 72% healthy", if it even makes sense). Please suggest :)

docs/logging-and-monitoring/health-metrics.md
```diff
@@ -0,0 +1,170 @@
...
+non-empty.
+Please note that the _current_ here means "according to the last epoch header written to the Node's
+local Ledger" - which makes this reading stale if the Node is not synced.
+
+rn_bft_proposal_timestamp_difference_seconds { key=..., component_address=... }
+: The difference between the specified Validator's wallclock and this Node's wallclock, as captured
+during the proposal creation / proposal handling.
+This "clock skew" actually includes the network delay too - but this is the number we care about,
+since we accept/reject proposals based on this clock difference. The protocol defines the acceptable
```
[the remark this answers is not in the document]
jakrawcz-rdx: alright, I now referenced java FQCN and field names from this file 👍 
_(they will definitely go out of sync, but what can we do...)_

jakrawcz-rdx:
Alright, so here are my ideas for higher-level indicators:

### `rn_health_local_ledger_progressing`
(boolean, i.e. `0.0` or `1.0`)

Computed as:
 `rn_ledger_last_update_epoch_second > now() - MAX_HEALTHY_LEDGER_UPDATE_INTERVAL`

This indicator has to be always green, since regardless whether the Node is syncing a backlog, or fully synced, or doing bft, it should be updating the ledger constantly.

❓ We need to define `MAX_HEALTHY_LEDGER_UPDATE_INTERVAL` - I am thinking `10 sec`?


### `rn_health_untrusted_network_alive`
(boolean, i.e. `0.0` or `1.0`)
⚠️ it is coming from a single peer, unverified, and we reflect it in the name

Computed as:
 `rn_sync_target_proposer_timestamp_epoch_second > now() - MAX_HEALTHY_BFT_DELAY`

This should be always green, since we expect network liveness - so, a current wallclock should appear on the ledger after some bounded delay introduced by BFT process.

❓ We need to define `MAX_HEALTHY_BFT_DELAY` - probably coupled with some BFT "timeout" constants that we have?  @LukasGasior1 hints?


### `rn_health_synced_with_alive_network`
(boolean, i.e. `0.0` or `1.0`)

Computed as:
 `rn_ledger_last_update_proposer_epoch_second > now() - (MAX_HEALTHY_BFT_DELAY + MAX_HEALTHY_LEDGER_UPDATE_INTERVAL)`

This is green `if` the network is alive `&&` we are fully synced with it. So:
- when `rn_health_untrusted_network_alive` is red, then this is not likely to be green (unless the untrusted information is actually a lie!)
- when this is green, then it is ALSO a trusted source of "network is alive" information!

❓ I figured we should allow a threshold equal to "time needed for the network to even commit the wallclock timestamp + time needed to sync this change from the network to this Node" - rightfully?


### `rn_health_untrusted_synced_with_network`
(boolean, i.e. `0.0` or `1.0`)
⚠️ it is coming from a single peer, unverified, and we reflect it in the name

Computed as:
 `rn_sync_current_state_version > rn_sync_target_state_version - MAX_HEALTHY_STATE_VERSION_LAG`

When this is green, then this Node is fully-synced with the network, regardless whether the network itself is alive or not.

❓ We need to define `MAX_HEALTHY_STATE_VERSION_LAG` - currently `HealthInfoServiceImpl` says "3", isn't this too low?


### `rn_health_validator_setting_effective`
(boolean, i.e. `0.0` or `1.0`)

Computed as:
 `(rn_misc_config_info.configured_validator_address != null) == rn_bft_in_validator_set`

This is green when the Node's "I want to be validator" desire is accordingly satisfied.
i.e. `(` the Node is configured to be a validator `&&` it is in validator set `) || (` not configured `&&` not in validator set `)`


### `rn_health_max_clock_amplitude_of_other_validator`
(continuous `[-inf, +inf]`)

Computed as:
Find the largest difference among all validators in `rn_bft_proposal_timestamp_difference_seconds`, and scale it relative to the appropriate MAX_DELAY/MAX_RUSH threshold.
(When the Node is NOT in validator set, then `0.0`)

So: the green range is `[-1.0, 1.0]`.
Meaning:
- if `0.03`, then the clocks of all validators in the set are super-nicely aligned with this Node's.
- if `-0.92`, then some fellow-validator has a very delayed clock, but still fine.
- if `0.87`, then some fellow-validator has a very rushing clock, but still fine.
- if `-2.62`, then some fellow-validator has a clock delayed so much that its proposals are probably being rejected.
- if `1.3`, then some fellow-validator has a clock rushing so much that its proposals are probably being rejected.

jakrawcz-rdx:
I think one more high-level readout is super useful for a "syncing" Node: namely the syncing rate (I explain it in the `State: Syncing` section of the `.md` in this PR)
Or maybe even an estimate on "remaining time until this Node is fully synced"?
However, it is hard to do this computation on Java side (compared to PromQL side, which can do `rate(rn_ledger_last_update_proposer_epoch_second{...}[$rate_interval])`).

Still, if you think that `rn_health_time_until_fully_synced_seconds` is a killer feature during syncing, then let me know and I'll see what can be done on Java side.

docs/logging-and-monitoring/monitoring-in-rust.md
```diff
@@ -5,15 +5,15 @@ The goal of recording tracing information is twofold:
 - aid the debugging by recording low-severity events and spans, like recording
   function calls, their parameters, and so on
 
-## Summary
```
LukasGasior1, line 8: In case anyone else is intrigued by this change :D it was an UTF-8 non-breaking space (`C2 A0`).

docs/logging-and-monitoring/health-metrics.md
LukasGasior1: that's correct

PR #585, opened 2023-08-04
Include `state_version` in all Core API `/state/...` responses

jakrawcz-rdx:
The easy part of https://radixdlt.atlassian.net/browse/NODE-563.

### DIFF
--- a/core-rust/core-api-server/core-api-schema.yaml
+++ b/core-rust/core-api-server/core-api-schema.yaml
@@ -6253,6 +6253,7 @@ components:
     StateConsensusManagerResponse:
       type: object
       required:
+        - state_version
         - config
         - state
         - current_proposal_statistic
@@ -6260,6 +6261,9 @@ components:
         - current_time
         - current_time_rounded_to_minutes
       properties:
+        state_version:
+          $ref: "#/components/schemas/StateVersion"
+          description: The state version at which the query was performed.
         config:
           $ref: "#/components/schemas/Substate"
         state:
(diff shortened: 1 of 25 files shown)


core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -6253,13 +6253,17 @@ components:
     StateConsensusManagerResponse:
       type: object
       required:
+        - state_version
```
dhedey, line 6256: For consistency / clarity, I'd suggest we put this in some `ledger_version` wrapper object (or `at_ledger_state` [like the Gateway API](https://radix-rcnet-v2-gateway.redoc.ly/#operation/StateEntityDetails)) which contains the state version and ledger hash stuff - that way, it's a little more separate from the rest of the response.

I'd even be up for if you'd like to put the rest of the response in some `content` or `value` or `data` wrapper, for clarity.

jakrawcz-rdx: > For consistency / clarity, I'd suggest we put this in some `ledger_version` wrapper object (or `at_ledger_state` ...

I am all in - and actually then it would make sense to pull the `state_version` into the `LedgerHeaderSummary(ledger_hashes, proposer_ts)`, as described at https://github.com/radixdlt/babylon-node/pull/586#issue-1837218388. The only awkward question left would be "how do we shape it in the existing LTS endpoints?".

> I'd even be up for if you'd like to put the rest of the response in some content or value or data wrapper

If we are adding the dedicated `at_ledger_state`/`ledger_header_summary`, then wrapping the rest too seems like an overkill.

jakrawcz-rdx:
Closing this in favor of https://github.com/radixdlt/babylon-node/pull/586#issuecomment-1667620833

PR #587, opened 2023-08-06
feature: Update to fig

dhedey:
Updates to the latest version from https://github.com/radixdlt/radixdlt-scrypto/pull/1315 / `release/fig` on scrypto.

There aren't any big changes:
* Mostly just renames or slight restructures in the Core API.
  * Notably the AccessRules module has been renamed to the RoleAssignment module.
  * A few of the Account substates have been renamed a little
  * There's a new Account partition for authorized depositors. This likely will not be exposed on the Gateway. 
* Also some new scenarios.
* A few behind-the-scenes changes (eg the three tier JMT integration)


core-rust/state-manager/src/staging/mod.rs
```diff
@@ -71,19 +71,13 @@ mod substate_overlay_iterator;
...
-pub trait ReadableStateTreeStore:
-    ReadableTreeStore<PartitionPayload> + ReadableTreeStore<()>
-{
-}
-impl<T> ReadableStateTreeStore for T where
-    T: ReadableTreeStore<PartitionPayload> + ReadableTreeStore<()>
-{
-}
+pub trait ReadableStateTreeStore: ReadableTreeStore {}
+impl<T> ReadableStateTreeStore for T where T: ReadableTreeStore {}
```
jakrawcz-rdx, line 80: We can just get rid of it now.
(But it's also fine in some next refactor, possibly drive-by)

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -1065,22 +1058,32 @@ impl RecoverableVertexStore for RocksDBStore {
 }
 
 fn encode_to_rocksdb_bytes(partition_key: &DbPartitionKey, sort_key: &DbSortKey) -> Vec<u8> {
-    let mut buffer = Vec::with_capacity(1 + partition_key.0.len() + sort_key.0.len());
+    let mut buffer = Vec::with_capacity(2 + partition_key.node_key.len() + sort_key.0.len());
```
jakrawcz-rdx, line 1061:
```suggestion
    let mut buffer = Vec::with_capacity(1 + partition_key.node_key.len() + 1 + sort_key.0.len());
```

(or some `size_of()`, or some well-named constants - and the same numbers can be used inside `decode_from_rocksdb_bytes()`)

But this is nitpicking only for my own readability; can be done in some next PR.

jakrawcz-rdx (approved):
I focused on the non-trivial JMT/DB-related changes.
I left some comments which can be treated as my own notes-to-self for later.

The remaining changes seem like renames-only (easy enough) + Some stuff that I don't understand ("depositor") - so it would be cool if someone else had a look there too.

LukasGasior1 (approved):

jakrawcz-rdx:
We want to merge https://github.com/radixdlt/babylon-node/pull/594 into this branch first.

... unless this change was unintentional:

image

?

PR #590, opened 2023-08-08
Filter out substate updates that do not change the value

LukasGasior1:


jakrawcz-rdx (approved):
The change is ofc valid 👍 

My only doubt is whether we are sure we want it. It probably doesn't hurt, but https://rdxworks.slack.com/archives/C01M90Y2448/p1691514783799429?thread_ts=1691494005.277249&cid=C01M90Y2448 doesn't really give definitive answer.
Anyway, if you are more sure than me (that we want this `if`), feel free to merge 🟢

dhedey (approved):

malexandru-rdx (approved):
LGTM!

PR #591, opened 2023-08-08
decouple keygen from cli-tools

balda-rdx:
this is taking keygen out of cli-tools so that we can easily build and distribute it as an executable without relying on rust to be present at build time.


malexandru-rdx (commented):
It seems fine and I'll approve after fixing the CI tests.

As a general rule before committing you should always run `./gradlew spotlessApply` to run the linter format.
Also you can run the tests locally:
- `./gradlew clean check jacocoTestReport --stacktrace --scan` for unit tests
- `./gradlew clean runSteadyStateIntegrationTests --refresh-dependencies` for integration tests

kofrdx (approved):

PR #595, opened 2023-08-09
Returning more details in the transaction submit response in case of "intent already committed"

jakrawcz-rdx:
This PR will help GW unambiguously detect a race condition (between async re-submit and async transaction streaming).
It should then resolve the "Core API transaction status bug" (https://whimsical.com/babylon-work-tracker-Am2s7hSa7xHd5yCxgub67a@2bsEvpTYSt1HitothizJCSfrasRGs4kqUdn).


core-rust/core-api-server/src/core_api/handlers/transaction_status.rs
```diff
@@ -193,6 +193,7 @@ fn map_rejected_payloads_due_to_known_commit(
                 .most_applicable_status()
                 .map(|reason| reason.to_string())
                 .unwrap_or_else(|| {
+                    // TODO(resolve during review): do we ever encounter this?
```
jakrawcz-rdx, line 196: _(note from author)_

We should probably discuss it further with @malexandru-rdx .
Maybe this never happens? (so, panic instead?)
Or maybe the status should be `AlreadyCommitted` instead?

(same in LTS endpoint)

core/src/test/java/com/radixdlt/api/core/NetworkSubmitTransactionTest.java
```diff
@@ -118,9 +118,93 @@ public void test_core_api_can_submit_and_commit_transaction() throws Exception {
...
+                              .network(networkLogicalName)
+                              .forceRecalculate(true)
+                              .notarizedTransactionHex(transaction.hexPayloadBytes())),
+              TransactionSubmitErrorResponse.class);
+
+      assertThat(errorResponse.getCode()).isEqualTo(400);
+      var details = (TransactionSubmitRejectedErrorDetails) errorResponse.getDetails();
+      assertThat(details.getIsIntentRejectionPermanent()).isTrue();
+      assertThat(details.getIsPayloadRejectionPermanent()).isTrue();
+      assertThat(details.getIsRejectedBecauseIntentAlreadyCommitted()).isTrue();
```
jakrawcz-rdx, line 154: _(note from author)_

I added these tests so that we precisely see how the 2 flavors differ (same intent in different payloads vs same payload) in terms of returned booleans.

E.g. my personal doubt is in this very line:
The test is named `rejected_when_same_payload`.
Logically the Node can reject it because same payload is committed (and this is a more "narrow" reason than "same intent committed").
So should we return `is_rejected_because_intent_already_committed` true or rather false?
(apparently I have problem with the word "because" :D)

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -4800,6 +4819,11 @@ components:
               description: |
                 Whether the cached rejection of this intent is due to the intent already having been committed.
                 If so, see the /transaction/receipt endpoint for further information.
+            intent_already_committed_as:
+              $ref: "#/components/schemas/CommittedIntentMetadata"
+              description: |
+                Details of an already-committed transaction that had the same intent.
+                This field is present when `is_rejected_because_intent_already_committed` is true.
```
jakrawcz-rdx, line 4826: _(note from author)_

There are a couple of ways to expose the "same intent-only vs same payload" distinction in the API.
The one picked here (`is_same_transaction: boolean`) is of course arbitrary, but that's what we discussed with David.

core/src/test/java/com/radixdlt/api/core/NetworkSubmitTransactionTest.java
malexandru-rdx: Technically, it is rejected because of `intent` (and not the `payload`) :P.

malexandru-rdx (approved):
LGTM!

core/src/test/java/com/radixdlt/api/core/NetworkSubmitTransactionTest.java
jakrawcz-rdx: hah, I see, so the official interpretation is:

"Yeah it would be totally legal for 2 committed transactions to have the same payload hash, but unfortunately it is not possible because that would mean the same intent hashes (which happens to be illegal)."

and then, indeed `IsRejectedBecauseIntentAlreadyCommitted == true` is valid for both cases 👍

core-rust/core-api-server/src/core_api/handlers/transaction_status.rs
jakrawcz-rdx: as discussed on Slack, it seems we can just keep constructing the (definitely known) error in this corner case.
In the latest commit I've migrated the error to its new representation and added some comments, PTAL.

PR #599, opened 2023-08-10
Fix SafetyState recovery; Add a check that loaded SafetyState belongs to the configured validator

LukasGasior1:


jakrawcz-rdx (commented):
I am very happy with the Guice-related changes, but I don't feel qualified to comment on the SafetyStore :(
I know I'd feel more reassured if there was a test showing how SafetyStore didn't work before this change (but ofc, as always - I will totally understand if it's too cumbersome to add at the moment).

LukasGasior1:
Adding a test is quite tricky as the issue was in our main module :man_shrugging:

jakrawcz-rdx (approved):
Please also get 🟢 from someone more knowledgeable in the SafetyState.

LukasGasior1:
> Please also get 🟢 from someone more knowledgeable in the SafetyState.

I'll note that this will need a retrospective review, but let's get it in for grape

dhedey (commented):
Really nice work!

PR #600, opened 2023-08-10
Distinction between owned and referenced entities in Core API schema

jakrawcz-rdx:
This addresses https://whimsical.com/babylon-work-tracker-Am2s7hSa7xHd5yCxgub67a@2bsEvpTYSt1Hj858bGqMKrKN669ZW1TRm1h

I went through the API yaml and inspected every `EntityReference`.
Some of them were populated from the Engine’s `Own` (most often `Vault(Own)`), and some were populated from raw `NodeId`. Namely: I think all `*SubstateValue`s used `Own`s, and component traverser + tx state changes used `NodeId`.
Anyways, for each `Own` case, I switched from `EntityReference` to a new `OwnedEntity`.


jakrawcz-rdx:
We discussed this change today briefly with GW's @krzlabrdx and the conclusions are:
- this PR provides some minor value to GW, albeit in some arbitrary way (i.e. via type system) while other are available (e.g. another `relation_type = OWNED | REFERENCED | ...` field).
- the specific impl in OpenApi's YAML (i.e. `allOf`) is not problematic (other than being an "artificial type", as mentioned above).
- this PR does not solve the GW's main pain point (regarding entity references): namely, a huge `instenceof` cascade starting [around here](https://github.com/radixdlt/babylon-gateway/blob/314b7521607087b8d256ccebdaadefe1715ed982/src/RadixDlt.NetworkGateway.PostgresIntegration/LedgerExtension/PostgresLedgerExtenderService.cs#L365), which still needs to manually scan all substates from receipt and build the map of relationships, etc (kinda similar to our parent-map, I believe)
- @krzlabrdx suggests that instead of this PR, we could design a way of providing the already-resolved information mentioned above.
  - One approach would be to migrate `new_global_entities` to just `new_entities`, and for each entity have a `{entity_reference, parent, root, [... potentially more relations captured here]}`
  - would this be related to work item https://whimsical.com/babylon-work-tracker-Am2s7hSa7xHd5yCxgub67a@2bsEvpTYSt1HjFSgBQM5AyCikHS15M6PRHX @dhedey ?

My takeaways for today:
- This PR can be closed without merging.
- The broader work described above will have to wait a bit...

dhedey:
Ahh fair.

In my mind, this PR still has merit by itself, as it more accurately represents the content in the engine.

But also happy to do something else in future to make Gateway's life easier.

PR #601, opened 2023-08-10
Expose Rust constants to Java & check config invariants

malexandru-rdx:


core/src/main/java/com/radixdlt/RadixNodeModule.java
```diff
@@ -207,6 +200,34 @@ protected void configure() {
...
+              + ").");
+    }
+    if (mempoolRelayerMaxRelayedSize < mempoolRelayerMaxMessagePayloadSize) {
+      throw new IllegalArgumentException(
+          "Invalid configuration: mempool.relayer.max_relayed_size ("
+              + mempoolRelayerMaxRelayedSize
+              + ") must be at least mempool.relayer.max_message_payload_size ("
+              + mempoolRelayerMaxMessagePayloadSize
+              + ").");
+    }
```
jakrawcz-rdx, line 230: `Preconditions.checkArgument()` may save a few lines here

core-rust/jni-export/src/constants.rs
```diff
@@ -0,0 +1,164 @@
...
+#[no_mangle]
+extern "system" fn Java_com_radixdlt_environment_NodeConstants_getMempoolTransactionOverheadFactorPercent(
+    env: JNIEnv,
+    _class: JClass,
+    request_payload: jbyteArray,
+) -> jbyteArray {
+    jni_sbor_coded_call(&env, request_payload, |_: ()| {
+        MEMPOOL_TRANSACTION_OVERHEAD_FACTOR_PERCENT
+    })
+}
```
jakrawcz-rdx, line 162: don't you prefer a single native method returning `struct AllTheseConsts{...}`?

(I know it will need an ugly `static {...}` initializer on java side)

malexandru-rdx: I thought about that, but decided to go for the current pattern of `ScryptoConstants`. What do you think of this refactor in a future more general `jni` refactor?

jakrawcz-rdx: If this is your plan in a refactoring that is really "coming soon", then I can live with the current shape for a moment 👍

core/src/main/java/com/radixdlt/RadixNodeModule.java
```diff
@@ -200,34 +201,30 @@ protected void configure() {
...
+            + ") must be at least the maximum transaction size ("
+            + NodeConstants.DEFAULT_MAX_TRANSACTION_SIZE
+            + ").");
+    Preconditions.checkArgument(
+        mempoolRelayerMaxRelayedSize < mempoolRelayerMaxMessagePayloadSize,
+        "Invalid configuration: mempool.relayer.max_relayed_size ("
+            + mempoolRelayerMaxRelayedSize
+            + ") must be at least mempool.relayer.max_message_payload_size ("
+            + mempoolRelayerMaxMessagePayloadSize
+            + ").");
```
jakrawcz-rdx, line 227: The `.checkArgument()` is convenient because it does formatting.
So you can do the multiline java string (`""" ...`) with placeholders (`%s`) - and it will also avoid rendering, as a bonus.

jakrawcz-rdx (approved):
Thank you for this clean-up 👍

PR #603, opened 2023-08-10
reorder steps in the Dockerfile to reduce the image size 

balda-rdx:
This reorder steps in the Dockerfile to reduce the image size. By unziping in the final stage we are adding an extra layer to the image that's not needed. it is better to unzip the artifacts in the build stage and then just copy, that way we prevent the extra layer

Also the user and groups creation were removed from the config script and moved to the Dockerfile. If there are any errors on those 2, we should be able to get them at build time and not at run time like now


docker/build_scripts/config_radixdlt.sh
```diff
@@ -1,20 +1,6 @@
 #!/bin/bash
-
 set -ex
 
-# Need this to run on Alpine (grsec) kernels without crashing
-find /usr -type f -name java -exec setfattr -n user.pax.flags -v em {} \;
-
-# Sets USER_ID to LOCAL_USER_ID if provided, else set it to 999
```
shambupujar, line 8: The lines below from down here should be kept intact. Tehy allow us to run the container as radixdlt user

shambupujar: Those steps needed at runtime as the vars are only avaliable on run time and they vary on individual node runners.

balda-rdx: @shambupujar so I only wanted to move the user creation to the Dockerfile, would you be ok with keeping the user and group creation there?
As of env vars, would you please let me know which ones are needed? I've tested the docker image created and it is starting just fine

balda-rdx: resolving as per the agreement on slack

PR #605, opened 2023-08-13
Metrics: a number of missed proposals (by this validator) during its N most recent rounds

jakrawcz-rdx:
This PR adds a `rn_ledger_recent_self_proposal_miss_count` metric (as promised in the last point of https://github.com/radixdlt/babylon-node/pull/582#issuecomment-1673254726). It will be needed for a computation of the "overall health" (see https://whimsical.com/node-status-and-health-metrics-QXjDREqeRDvGDu6xg3J8wx@2bsEvpTYSt1HjHxnUjRqMBYNQDsmaPNqZ5N).

In the impl I went for an actual ring-buffer. So this is not a heuristic, but an exact "Count of occurrences within N most recent opportunities".

In terms of manual verification: I have added a random `if (now % 100 < 17) { reject() }` to BFT's `processProposal()` method, and after a few seconds I captured the metrics (of a `Run Single Validator`):
```
# HELP rn_ledger_recent_self_proposal_miss_count A number of proposals missed by this validator during its 100 most recent rounds.
# TYPE rn_ledger_recent_self_proposal_miss_count gauge
rn_ledger_recent_self_proposal_miss_count 17
```

The readout is also aligned with:
```
# HELP rn_ledger_consensus_rounds_committed Count of rounds processed by consensus that reached the ledger commit.
# TYPE rn_ledger_consensus_rounds_committed counter
rn_ledger_consensus_rounds_committed{leader_component_address="833d72e50885b684fb3d59e496d02c824da5a7ca3cdc3cc3c2ee4fa2d701",round_resolution="MissedByFallback"} 566
rn_ledger_consensus_rounds_committed{leader_component_address="833d72e50885b684fb3d59e496d02c824da5a7ca3cdc3cc3c2ee4fa2d701",round_resolution="MissedByGap"} 0
rn_ledger_consensus_rounds_committed{leader_component_address="833d72e50885b684fb3d59e496d02c824da5a7ca3cdc3cc3c2ee4fa2d701",round_resolution="Successful"} 2805
```
_(`566 / (566 + 2805) ~= 0.168`)_


core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/CommitRequest.java
```diff
@@ -75,7 +76,9 @@
 public record CommitRequest(
     List<RawLedgerTransaction> transactions,
     LedgerProof proof,
-    Option<byte[]> postCommitVertexStoreBytes) {
+    Option<byte[]> postCommitVertexStoreBytes,
+    // for metrics calculation only
+    Option<ComponentAddress> selfValidatorAddress) {
```
jakrawcz-rdx, line 81: _(note from author)_

> Why is this a part of `CommitRequest` and not a part of `StateManager`'s initial configuration?

Because injecting `SelfValidatorInfo` to a `NodeRustEnvironment` introduces a circular dependency (the construction of `SelfValidatorInfo` requires `REv2LedgerInitializerToken`, which cannot be constructed without having `NodeRustEnvironment`).

I am ashamed of how we use Guice - still, if you see a way of improving this very line with a quick-win (i.e. without refactoring our entire DI), I am open to suggestions.

core-rust/state-manager/src/metrics.rs
```diff
@@ -387,3 +403,64 @@ impl MetricLabel for VertexPrepareStopReason {
         }
     }
 }
+
+/// A number of most recent [`RoundSlot`]s of a single validator to track for metrics purposes.
+const PROPOSAL_HISTORY_LEN: usize = 100;
```
jakrawcz-rdx, line 408: _(note from author)_

@LukasGasior1 this is the `N` you used - do you care to suggest some actual value :D ?
(mine was not engineered at all)

core-rust/state-manager/src/metrics.rs
```diff
@@ -394,3 +414,64 @@ impl MetricLabel for VertexPrepareStopReason {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum RoundSlot {
+    Success,
+    Missed,
+}
+
+/// A higher-level metric helper, tracking a number of recent proposal misses of a specific
+/// validator.
+pub struct SelfProposalMissTracker {
+    buffer: Mutex<RingBuffer<RoundSlot, PROPOSAL_HISTORY_LEN>>,
```
malexandru-rdx, line 431: Why is `buffer` behind a mutex?

malexandru-rdx (approved):
Looks good!

core-rust/state-manager/src/metrics.rs
jakrawcz-rdx: as discussed on Slack:

We want interior mutability here (as is the case with all other metric bumps), and we don't feel too comfortable with `unsafe`, so we prefer a hopefully-cheap mutex (which will never block in practice).

LukasGasior1 (approved):

PR #607, opened 2023-08-14
Storing (and returning from Core API) previous value of deleted substates

jakrawcz-rdx:
This addresses https://whimsical.com/babylon-work-tracker-Am2s7hSa7xHd5yCxgub67a@2bsEvpTYSt1Hj6ARxzroL4vo7cY4MiTG5Fd

### DIFF
--- a/core-rust/state-manager/src/staging/result.rs
+++ b/core-rust/state-manager/src/staging/result.rs
@@ -247,11 +247,13 @@ impl ProcessedCommitResult {
                                 new: value.clone(),
                                 previous,
                             }),
-                            Some(_) => None, /* Same value as before, ignore */
-                            None => Some(ChangeAction::Create(value.clone())),
+                            Some(_) => None, /* Same value as before (i.e. not updated), ignore */
+                            None => Some(ChangeAction::Create { new: value.clone() }),
                         }
                     }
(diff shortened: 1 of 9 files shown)


core-rust/state-manager/src/staging/result.rs
```diff
@@ -247,11 +247,13 @@ impl ProcessedCommitResult {
...
-                            Some(_) => None, /* Same value as before, ignore */
-                            None => Some(ChangeAction::Create(value.clone())),
+                            Some(_) => None, /* Same value as before (i.e. not updated), ignore */
+                            None => Some(ChangeAction::Create { new: value.clone() }),
                         }
                     }
-                    DatabaseUpdate::Delete => Some(ChangeAction::Delete),
+                    DatabaseUpdate::Delete => store
+                        .get_substate(&partition_key, &sort_key)
+                        .map(|previous| ChangeAction::Delete { previous }),
```
jakrawcz-rdx, line 256: _(note from author)_

This has the same behavior as the "previous value same" for the Update above:
If the value did not exist, then the Delete is ignored.

LukasGasior1 (approved):

malexandru-rdx (approved):

PR #610, opened 2023-08-15
revert the changes to how we handle users within the docker image

balda-rdx:


docker/build_scripts/config_radixdlt.sh
```diff
@@ -1,6 +1,16 @@
 #!/bin/bash
 set -ex
 
+# Sets USER_ID to LOCAL_USER_ID if provided, else set it to 999
+USER_ID=${LOCAL_USER_ID:-996}
```
marek-karwacki-rdx, line 5: The comment mentions `999` but the code has `996`

marek-karwacki-rdx (approved):

PR #612, opened 2023-08-16
Metrics: a rate of ledger proposer timestamp during N most recent ledger updates

jakrawcz-rdx:
⚠️ This builds on top of https://github.com/radixdlt/babylon-node/pull/605 (re-uses the ring-buffer impl)

This PR adds a `rn_ledger_recent_proposer_timestamp_progress_rate` metric (as promised in the https://github.com/radixdlt/babylon-node/pull/582#issuecomment-1673254726). It will be needed for a computation of the "overall health" (see https://whimsical.com/node-status-and-health-metrics-QXjDREqeRDvGDu6xg3J8wx@2bsEvpTYSt1Hir2FvjaStZLhMXjmH9Prk7p).

The metric:
```
# HELP rn_ledger_recent_proposer_timestamp_progress_rate A rate of the proposer timestamp progress (against wall-clock) averaged over 10 most recent ledger updates.
# TYPE rn_ledger_recent_proposer_timestamp_progress_rate gauge
rn_ledger_recent_proposer_timestamp_progress_rate 0.9982936655602106
```
_(nicely oscillates around ~1.0 on a healthy Node)_

### DIFF
--- a/core-rust/state-manager/src/metrics.rs
+++ b/core-rust/state-manager/src/metrics.rs
@@ -84,6 +84,7 @@ pub struct LedgerMetrics {
     pub last_update_epoch_second: Gauge,
     pub last_update_proposer_epoch_second: Gauge,
     pub recent_self_proposal_miss_count: SelfProposalMissTracker,
+    pub recent_proposer_timestamp_progress_rate: ProposerTimestampProgressRateTracker,
 }
 
 pub struct CommittedTransactionsMetrics {
@@ -108,6 +109,7 @@ impl LedgerMetrics {
         network: &NetworkDefinition,
         lock_factory: &LockFactory,
         registry: &Registry,
+        current_ledger_proposer_timestamp_ms: i64,
     ) -> Self {
         Self {
             address_encoder: AddressBech32Encoder::new(network),
@@ -145,6 +147,15 @@ impl LedgerMetrics {
                     &format!("A number of proposals missed by this validator during its {} most recent rounds.", PROPOSAL_HISTORY_LEN),
                 ),
                 &lock_factory.named("self_proposal_miss_tracker"),
+                registry,
+            ),
+            recent_proposer_timestamp_progress_rate: ProposerTimestampProgressRateTracker::new(
+                current_ledger_proposer_timestamp_ms,
+                opts(
+                    "ledger_recent_proposer_timestamp_progress_rate",
+                    &format!("A rate of the proposer timestamp progress (against wall-clock) averaged over {} most recent ledger updates.", PROGRESS_RATE_HISTORY_LEN),
+                ),
+                &lock_factory.named("progress_rate_tracker"),
                 registry,
             ),
         }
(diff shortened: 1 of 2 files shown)


core-rust/state-manager/src/metrics.rs
```diff
@@ -473,3 +486,73 @@ impl SelfProposalMissTracker {
         self.gauge.add(new_missed_count - outdated_missed_count);
     }
 }
+
+/// A number of most recent [`ProposerTimestampDatapoint`]s to track for metrics purposes.
+const PROGRESS_RATE_HISTORY_LEN: usize = 10;
```
malexandru-rdx, line 491: How was `10` calculated?

malexandru-rdx (approved):
LGTM!

core-rust/state-manager/src/metrics.rs
```diff
@@ -473,3 +486,73 @@ impl SelfProposalMissTracker {
...
+/// ledger.
+pub struct ProposerTimestampProgressRateTracker {
+    buffer: Mutex<RingBuffer<ProposerTimestampDatapoint, PROGRESS_RATE_HISTORY_LEN>>,
+    gauge: Gauge,
+}
+
+impl ProposerTimestampProgressRateTracker {
+    /// Creates a new tracker and registers its resulting [`Gauge`] (with the given options) at the
+    /// given registry.
+    /// Note: the [`LockFactory`] is required to ensure a thread-safe access to a ring-buffer used
```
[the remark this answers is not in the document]
jakrawcz-rdx: yeah it would increase the chances (A LITTLE BIT) of not staying like that forever... What would its impl be? (most importantly: would it still require a `LockFactory` dep, or are you thinking about something smarter)

On the other hand, the metrics lib we use has all its metric primitives thread-safe. Our new metric primitive happens to be a bit more complex and requires Mutex (well technically in Java it wouldn't require one...) - is it wrong to keep it thread-safe then? (I am afraid that callers may assume it - especially after you mentioned the future in your comment).

core-rust/state-manager/src/metrics.rs
jakrawcz-rdx: It is a wild guess :D
I am happy with any number that in practice translates to a "few seconds, up to 1 minute" order of magnitude.
(exposing an average-over-durations longer than Prometheus scraping interval [15-60 sec] does not make much sense)
I am open to more educated guesses.

malexandru-rdx: With current round rate (3/s - which should go up) 10 seems a little bit low. 20, 30?

LukasGasior1: > no server-side throttling there?

yup, there's no throttling at the moment

170 is a lot, but plausible

LukasGasior1: maybe 100 slots then? If we're syncing fast we don't care that the sample is for a shorter period, we're almost certainly healthy anyway (right?)

LukasGasior1 (approved):

core-rust/state-manager/src/metrics.rs
jakrawcz-rdx: Alright, I adjusted it to 100, lemme merge then 👍

PR #613, opened 2023-08-16
Adding an overall ledger health indicator metric (`rn_ledger_overall_health_factor`)

jakrawcz-rdx:
This PR finally implements https://whimsical.com/node-status-and-health-metrics-QXjDREqeRDvGDu6xg3J8wx@2Ux7TurymMmBywdkUdxU with minor tweaks:
- when the Node is synced, we don't consider its syncing rate (which is guaranteed to be ~1.0),
- the results are not boolean/enum, but a continuous 0.0-1.0 factor; and the operations are not logical AND, but multiplication (this is only a technical choice, to be better aligned with single-dimensional Prometheus signal)

### DIFF
--- a/core-rust/node-common/src/metrics.rs
+++ b/core-rust/node-common/src/metrics.rs
@@ -63,6 +63,7 @@
  */
 
 use prometheus::core::*;
+use prometheus::proto::{MetricFamily, MetricType};
 use prometheus::*;
 use std::mem;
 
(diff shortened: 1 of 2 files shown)


jakrawcz-rdx:
Regarding the manual verification:

I have added a `if random() < 0.01 { reject() }` to proposal processing, and then I waited to get that proposal miss:
```
# HELP rn_ledger_recent_self_proposal_miss_count A number of proposals missed by this validator during its 100 most recent rounds.
# TYPE rn_ledger_recent_self_proposal_miss_count gauge
rn_ledger_recent_self_proposal_miss_count 1
...
# HELP rn_ledger_overall_health_factor A proper fraction representing an overall local ledger health (with 0.0 = critical and 1.0 = healthy).
# TYPE rn_ledger_overall_health_factor gauge
rn_ledger_overall_health_factor 0.5
```
(the health factor correctly dropped from `1.0` 🟢  to `0.5` 🟡 )

When more proposals were missed, it got down to `0.0` 🔴 .

I was not able to easily test varying ledger-syncing speed. But the "am I synced" condition definitely worked in the above case (because the syncing speed was `~1.0`, which is `< 10` - yet it didn't affect healthiness, because of top-of-ledger timestamp was fresh enough).

LukasGasior1 (approved):
Looks good

PR #614, opened 2023-08-16
Align netty decoder limits with FrameCodec; add metrics for dropped too large messages

LukasGasior1:
- add metrics for messages that were dropped due to their excessive size

- fix LargeMessageTest (it's still a bit flaky and requires manual inspection, but at least now it actually works)

- slightly reorganize max message size limits / constants:

There's an application-level `@MaxMessageSize`, which is based on `proposalMaxUncommittedTransactionsPayloadSize`, `MAX_TXN_BYTES_FOR_A_SINGLE_RESPONSE` and `mempoolRelayerMaxMessagePayloadSize`. It must be at most `FrameCodec.MAX_FRAME_BODY_SIZE` (so that the message can always fit in an encrypted frame).

On the netty side, a maximum length for the length-prefixed frames is @MaxMessageSize + FrameCodec overhead


core/src/main/java/com/radixdlt/messaging/MaxMessageSize.java
```diff
@@ -0,0 +1,142 @@
...
+ * You are solely responsible for; testing, validating and evaluation of all operation
+ * logic, functionality, security and appropriateness of using the Work for any commercial
+ * or non-commercial purpose and for any reproduction or redistribution by You of the
+ * Work. You assume all risks associated with Your use of the Work and the exercise of
+ * permissions under this License.
+ */
+
+package com.radixdlt.messaging;
+
+/* Copyright 2021 Radix Publishing Ltd incorporated in Jersey (Channel Islands).
```
jakrawcz-rdx, line 67: unintentional? (2x copyright)

jakrawcz-rdx (approved):
Really nice 👍

core/src/main/java/com/radixdlt/messaging/MaxMessageSize.java
LukasGasior1: yup, removed

PR #619, opened 2023-08-18
DO 1663 / simplify docker-tagging

kofrdx:
Writes tag 'rcnet-v1' when creating a release for tag 'rcnet-v1'. Value equals tag of the release. This tag has highest priority.
Writes 'feature-123-54dfge56' for releases on branch 'feature-123' with commit sha '54dfge56'
Writes 'sha-54dfge56' with commit sha '54dfge56'

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -140,6 +140,9 @@ jobs:
       image_name: "babylon-node"
       tags: |
         type=ref,event=tag
+        type=semver,pattern=v{{version}}         # v1.0.0 if tag is v1.0.0
+        type=semver,pattern=v{{major}}.{{minor}} # v1.0 if tag is v1.0.0
+        type=semver,pattern=v{{major}}           # v1 if tag is v1.0.0
       flavor: |
         suffix=-amd64
       context: "."
@@ -162,6 +165,9 @@ jobs:
       image_name: "babylon-node"
       tags: |
         type=ref,event=tag
+        type=semver,pattern=v{{version}}         # v1.0.0 if tag is v1.0.0
+        type=semver,pattern=v{{major}}.{{minor}} # v1.0 if tag is v1.0.0
+        type=semver,pattern=v{{major}}           # v1 if tag is v1.0.0
       flavor: |
         suffix=-arm64
       context: "."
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -20,64 +20,9 @@ jobs:
         with:
           access_token: ${{ github.token }}
 
-  join_multiarch_image_dockerhub:
```
kofrdx, line 23: Moved down. Closer to the jobs it depends on.

.github/workflows/docker.yml
```diff
@@ -20,64 +20,9 @@ jobs:
...
-      - name: Join multiarch images
-        run: |
-          docker buildx imagetools create -t docker.io/radixdlt/babylon-node:${{ github.event.release.tag_name }} \
-            docker.io/radixdlt/babylon-node:${{ github.event.release.tag_name }}-arm64 \
-            docker.io/radixdlt/babylon-node:${{ github.event.release.tag_name }}-amd64
-
   build_deb:
     name: Build debian package
     runs-on: babylon-runner
-    outputs:
```
kofrdx, line 78: Obsolete. We do not need to create these tags when building the binary.

.github/workflows/docker.yml
```diff
@@ -86,22 +31,6 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - name: Docker meta
```
kofrdx, line 89: Obsolete and misleading. We do not use these tags for creating the image.

.github/workflows/docker.yml
```diff
@@ -160,84 +89,57 @@ jobs:
           name: deb4docker
           path: "${{ github.workspace }}/docker/*.deb"
 
-  setup_tags:
```
kofrdx, line 163: Same can be done with a single line of config: 
```
tags: |
        type=ref,event=tag
        type=raw,value={{branch}}-{{sha}}
        type=sha
```

Also this introduces errors if the hash is only numbers like '1456234'. However uncommon, this results in helm reformating this integer into its scientific representation like '1.456234+e5' or something similar. Not relevant here since the deployment happens without helm. But relevant for other repos with similar patterns.

PR #627, opened 2023-08-24
regenerate ts sdk for grape

dawidsowardx:


malexandru-rdx (approved):
Unfortunately we don't have ts-sdk tests run in CI and we have to do it manually for now.

I've fixed the tests.

PR #628, opened 2023-08-24
Update Scrypto dep in Node

jakrawcz-rdx:
This is a "regular" Scrypto version bump (to `develop-510c2872`).

It blocks a bump to `release/rcnet-v3`.

Most important changes:
- using proper "vault balance changes" (fixed and passed here from Engine) in order to compute "global balance changes"
  - this addresses https://whimsical.com/babylon-work-tracker-Am2s7hSa7xHd5yCxgub67a@2bsEvpTYSt1HjGQgXhRaLKQpx2PZ4dTiUsw
  - it had to happen in the "update scrypto" PR because the old, imperfect Engine "global balance changes" fields are GONE: thus, please pay attention to the new business logic added
- adjusting the Core API mapping logic to work with "versioned substates" (as described at https://rdxworks.slack.com/archives/C05FHR2L79N/p1692810180989879)

Please also see 3x marked `TODO(during review)` - any suggestions/solutions are welcome.


core-rust/state-manager/src/transaction/executable_logic.rs
```diff
@@ -76,14 +76,7 @@ impl ExecutionConfigurator {
...
                     ExecutionConfig {
-                        // Explanation: The first epoch change in Stokenet panicked with
-                        // `SystemModuleError(TransactionLimitsError(TooManyEntriesInTrack))` - so instead
-                        // we override these limits with the genesis limit until they can be fixed upstream.
-                        //
-                        // TODO(rcnet-v3) - Remove these down when we understand what the limit should be,
-                        // and have adjusted ExecutionConfig::for_system_transaction() in the engine.
-                        max_number_of_substates_in_track: 50_000,
-                        max_number_of_substates_in_heap: 50_000,
+                        // TODO(during review): were these "max number of substates" adjusted in the Engine?
```
malexandru-rdx, line 79: I see that now it's bounded by substate total bytes and it is 64MB. If genesis ingestion still works I guess it's fine. (and 64MB seems enough for 50000 substates).

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -5426,6 +5435,7 @@ components:
         status:
           $ref: "#/components/schemas/LtsCommittedTransactionStatus"
         fungible_entity_balance_changes:
+          # TODO(during review): is this description still true? we have improved both fees and balance changes recently...
```
malexandru-rdx, line 5438: The node logic seems right - unless something in RE makes it inaccurate (doubt it). It's most probably a leftover.

malexandru-rdx (approved):

jakrawcz-rdx:
Gods of Healthy Develop, please forgive me - I am merging this huge PR with just one approval, and with 1 more pending `TODO(during review)` item.
Another huge PR is due (the `release/rcnet-v3`), and it touches very similar aspects, so I am sure we will have opportunity to point out any remaining problems of this one.

PR #630, opened 2023-08-26
Release RCnet-v3 (bumping Scrypto dep to rcnet-v3-177d45b0)

jakrawcz-rdx:
This is basically a continuation of https://github.com/radixdlt/babylon-node/pull/628.

The most important changes:
- (continued) adjusting the Core API mapping logic to work with "versioned substates" (as described at https://rdxworks.slack.com/archives/C05FHR2L79N/p1692810180989879)
- (the only substantial Core API spec change) Adding a "field index" to ResourceManager's `mutable_fields`.


core/src/test/java/com/radixdlt/genesis/olympia/converter/REv2GenesisFromOlympiaTest.java
```diff
@@ -135,6 +135,10 @@ public void genesis_from_olympia_should_initialize_correct_xrd_balances() throws
     }
   }
 
+  /**
+   * Deserializes a file captured from Olympia Node's `StateIRSerializationTest` (namely,
+   * `it_should_successfully_construct_and_then_serialize_and_deserialize_the_intermediate_state`).
+   */
```
jakrawcz-rdx, line 141: _(note from author)_

⚠️ Important question!
I actually had to adjust the Olympia test a bit to get a valid validatorUrl in here.
**Will real Olympia validators all have URLs set?**
(if not, our Node will crash during Guice instantiation - which makes me think that a better solution is to have `if !valid { passDummyUrlToEngine() }` in our Node)

jakrawcz-rdx: Ok, so together with @LukasGasior1 we think this is a problem that has to be addressed (because otherwise it is too easy for any Olympia validator to crash Babylon Node during migration).

Facts:
1. In Olympia, the URL is de-facto optional (i.e. considered valid when empty).
2. In Olympia, when present, the URL is validated with [a regex](https://github.com/radixdlt/radixdlt/blob/main/radixdlt-engine/src/main/java/com/radixdlt/atom/REFieldSerialization.java#L82) different than Babylon Engine's [regex](https://github.com/radixdlt/radixdlt-scrypto/blob/015521043473efe58b20bf6ae254bfdf74ebdade/radix-engine-interface/src/api/node_modules/metadata/models/url.rs#L13) (e.g. Olympia allows `ftp://` URLs)

We have the following competing ideas on how to fix this:

**A)** Babylon Node could validate the Olympia's URLs using the Babylon Engine's regex - and when not valid, pass a hardcoded `https://www.radixdlt.com/` to the Engine.
   - This is a **minor Node change only**

**B)** Babylon Node could validate the Olympia's URLs using the Babylon Engine's regex - and when not valid, pass `None` to the Engine
   - This is a **minor Node and Engine change** (make validator URL optional)

**C)** Engine could relax the validation of this field: basically, allow any String, including empty, or a very long one (at least as long as allowed by Olympia)
   - This is a **minor Engine change only**

jakrawcz-rdx: After some consultation, I went with option **B)** (please see the latest commit).

core-rust/state-manager/src/jni/test_state_reader.rs
```diff
@@ -202,7 +200,7 @@ extern "system" fn Java_com_radixdlt_testutil_TestStateReader_componentXrdAmount
             let read_store = database.read();
 
             // a quick fix for handling virtual accounts
-            // TODO: fix upstream
+            // TODO(during review): has this been "fixed upstream"?
```
dhedey, line 203: No, it hasn't - but I suggest removing this TODO. I believe this is only used in tests anyway.

core-rust/core-api-server/src/core_api/conversions/substates/resource.rs
```diff
@@ -148,32 +153,39 @@ pub fn to_api_non_fungible_resource_manager_total_supply_substate(
...
         },
         Value {
-            // TODO(RCnet-v3) - We didn't want to change this for RCnet-v2 phase 2
-            // So for RCnet-v3, fix this to map properly to include the field indices
-            mutable_fields: mutable_field_index.keys().cloned().collect(),
+            // TODO(during review): Please thoroughly review the feature change (i.e. my
+            // interpretation of "include the field indices").
+            mutable_fields: mutable_field_index
+                .iter()
+                .map(|(name, index)| to_api_mutable_field(context, name.as_str(), *index))
```
dhedey, line 170: Looks good to me

core-rust/core-api-server/src/core_api/conversions/substates/royalty_module.rs
```diff
@@ -7,17 +7,14 @@ use radix_engine_queries::typed_substate_layout::*;
...
         substate,
         RoyaltyModuleFieldState,
-        ComponentRoyaltySubstate {
-            enabled,
-            royalty_vault,
-        },
+        ComponentRoyaltySubstate { royalty_vault },
         Value {
-            is_enabled: *enabled,
+            is_enabled: true, // TODO(during review): drop it instead?
```
dhedey, line 17: I don't mind either way. The substate won't exist any more if royalties aren't enabled

core/src/test/java/com/radixdlt/genesis/olympia/converter/REv2GenesisFromOlympiaTest.java
dhedey: Yep, (B) sounds right to me.

dhedey (approved):

dhedey:
I'm not sure what this TODO is referring to?
```
// TODO(during review): there is some convoluted logic there; does it need to change now?
```

core-rust/core-api-server/src/core_api/conversions/substates/consensus_manager.rs
```diff
@@ -9,31 +11,32 @@ use radix_engine_queries::typed_substate_layout::*;
...
+        TypedSubstateKey::MainModule(TypedMainModuleSubstateKey::ConsensusManager(
+            ConsensusManagerTypedSubstateKey::RegisteredValidatorByStakeSortedIndexEntry(
+                sorted_validator
             )
-        )
+        ))
     );
-    let validator = substate;
-    Ok(index_substate!(
+    let (divided_stake, validator_address) = sorted_validator.into_sort_key_and_content();
```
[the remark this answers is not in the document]
jakrawcz-rdx: perfect, lemme pull the latest scrypto then 👍

core/src/main/java/com/radixdlt/genesis/olympia/converter/EngineUrlPredicate.java
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * A URL validator enforcing the rules used by the Engine, as captured from <a
+ * href="https://github.com/radixdlt/radixdlt-scrypto/blob/rcnet-v3-a74271b8/radix-engine-interface/src/api/node_modules/metadata/models/url.rs">
+ * rcnet-v3-a74271b8</a>.
+ *
+ * <p>We can only propagate URL metadata items which satisfy the Engine's rules, since an invalid
+ * URL would cause a transaction to fail.
+ */
+public final class EngineUrlPredicate implements Predicate<String> {
```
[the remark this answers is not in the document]
jakrawcz-rdx: > for now.

AFAIU, this is needed only for Olympia migration, so I imagine it will vanish completely quite soon (and no point having a future-proof solution).

core/src/main/java/com/radixdlt/genesis/olympia/converter/OlympiaValidatorsConverter.java
```diff
@@ -78,8 +78,64 @@
...
+
+  /**
+   * A URL string validator based on {@code URL_REGEX} used by the Engine, captured at <a
+   * href="https://github.com/radixdlt/radixdlt-scrypto/blob/rcnet-v3-a74271b8/radix-engine-interface/src/api/node_modules/metadata/models/url.rs">
+   * rcnet-v3-a74271b8</a>.
+   *
+   * <p>We will only propagate URL metadata items which satisfy this regex, since an invalid URL
+   * would cause transaction to fail.
+   */
+  private static final Predicate<String> VALID_ENGINE_URL =
```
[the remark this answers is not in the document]
jakrawcz-rdx: ahhh, it's a unused leftover after extracting the `EngineUrlPredicate`, thanks for spotting!

core-rust/core-api-server/src/core_api/conversions/substates/transaction_tracker.rs
```diff
@@ -12,12 +13,14 @@ pub fn to_api_transaction_tracker_substate(
     Ok(field_substate!(
```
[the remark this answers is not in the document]
jakrawcz-rdx: because of:
```
error[E0599]: no method named `as_latest_ref` found for reference `&radix_engine_queries::typed_substate_layout::TransactionTrackerSubstate` in the current scope
   --> core-api-server/src/core_api/conversions/substates/mod.rs:101:64
    |
101 |                       let $value_unpacking = $substate.payload().as_latest_ref()
    |                                                                  ^^^^^^^^^^^^^ method not found in `&TransactionTrackerSubstate`
    |
...
```

The versioned substate helpers for transaction tracker are apparently hand-crafted (i.e. not using `declare_native_blueprint_state!`). I assumed there is a reason for it (some custom handling needs, etc) - are you saying that it is simply an omission on Engine side?

jakrawcz-rdx:
> I'm not sure what this TODO is referring to?
> `// TODO(during review): there is some convoluted logic there; does it need to change now?`

In that line, we previously passed the imperfect `state_update_summary.balance_changes` provided by the Engine. I changed it to `&local_execution.global_balance_changes` (computed from the Engine's `state_update_summary.vault_balance_changes` by Node).
Hence, I wonder if this affects any heuristics used by `to_api_lts_fungible_balance_changes()`...? Does `FeePaymentComputer` need any changes?

jakrawcz-rdx:
I see that both runs of `normal_run_with_transactions_should_not_cause_unexpected_errors` fail (both on CI and on my machine) due to some BFT reasons:

```
com.radixdlt.integration.steady_state.deterministic.rev2.SanityTest normal_run_with_transactions_should_not_cause_unexpected_errors[0] FAILED (36.7s)

  com.radixdlt.harness.deterministic.DeterministicNodes$EventHandleException: Exception: com.radixdlt.harness.deterministic.invariants.DeterministicMonitors$ConsensusLivenessException: Liveness Increasing QC invariant of 3000 ms broken. LastSeen: EpochRound{epoch=2 round=2} 225 CurrentTime: 3255
  On message: 19->7(3255) SyncRequest{header=DtoLedgerProof{header=LedgerHeader{version=32 hashes=LedgerHashes{stateRoot=a2ea61daada9907458d5a1dc151c9254bf7ab9400bfeec42a8c2e1be4a69c501, transactionRoot=d7448ea878c0e295c5673e490c32f5be273391cf3aa10bb18ac02aaaadc3e79e, receiptRoot=97709ae86f40baa53a960566f9d3727c2bffcac129e5f3e8b95294aa2afff9c9} consensus_parent_round_timestamp=2465 proposer_timestamp=2525 epoch=2 round=23 next_epoch=null}}}
```

I think this is because I pulled https://github.com/radixdlt/babylon-node/pull/637 from `develop` into here.
@LukasGasior1 does this indicate some actual problem, or do we simply have to adjust the test assert?

core-rust/core-api-server/src/core_api/conversions/substates/transaction_tracker.rs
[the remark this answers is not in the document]
dhedey: Ahh I see, weird. But okay, no worries :) - if/when it's fixed engine side we can fix it here too.

dhedey:
> In that line, we previously passed the imperfect state_update_summary.balance_changes provided by the Engine. I changed it to &local_execution.global_balance_changes (computed from the Engine's state_update_summary.vault_balance_changes by Node).
Hence, I wonder if this affects any heuristics used by to_api_lts_fungible_balance_changes()...? Does FeePaymentComputer need any changes?

I don't believe it should - the balance changes should be the same, except also include recalls, and this logic should all handle that. 👍

PR #635, opened 2023-08-29
Returning AccessController substate as typed struct (instead of raw bytes) in Core API

jakrawcz-rdx:
Addresses https://whimsical.com/babylon-work-tracker-Am2s7hSa7xHd5yCxgub67a@2bsEvpTYSt1HjDQywpU1gxwWn4sHpjXp5jq.


core-rust/core-api-server/src/core_api/conversions/substates/access_controller.rs
```diff
@@ -12,11 +13,95 @@ pub fn to_api_access_controller_substate(
...
+                recovery_role_badge_withdraw_attempt_state,
+                RecoveryRoleBadgeWithdrawAttemptState::BadgeWithdrawAttempt
+            ),
         }
     ))
 }
+
+pub fn to_api_recovery_proposal(
+    context: &MappingContext,
+    substate: &RecoveryProposal,
```
dhedey, line 87: Very minor - but this shouldn't be called substate.

And we generally should be using an explicit destructuring so that if new fields are added, we have a compile error and are prompted to add them.

core-rust/core-api-server/src/core_api/conversions/substates/access_controller.rs
```diff
@@ -12,11 +13,95 @@ pub fn to_api_access_controller_substate(
...
+                primary_role_locking_state,
+                primary_role_recovery_attempt_state,
+                primary_role_badge_withdraw_attempt_state,
+                recovery_role_recovery_attempt_state,
+                recovery_role_badge_withdraw_attempt_state,
+            )
         },
         Value {
-            data_struct: Box::new(to_api_data_struct_from_bytes(context, &data)?),
+            controlled_vault_entity: Box::new(to_api_entity_reference(
```
dhedey, line 29: This should be an owned entity, rather than a reference

core-rust/core-api-server/src/core_api/conversions/substates/access_controller.rs
```diff
@@ -12,11 +13,95 @@ pub fn to_api_access_controller_substate(
...
+                context,
+                controlled_asset.as_node_id()
+            )?),
+            timed_recovery_delay_minutes: timed_recovery_delay_in_minutes
+                .as_ref()
+                .map(|minutes| to_api_u32_as_i64(*minutes)),
+            recovery_badge_resource_address: to_api_resource_address(context, recovery_badge)?,
+            is_primary_role_locked: matches!(
```
dhedey, line 37: Minor, but would slightly rather do an explicit match to ensure we get a compile error if new variants get added.

core-rust/core-api-server/src/core_api/conversions/substates/access_controller.rs
```diff
@@ -12,11 +13,95 @@ pub fn to_api_access_controller_substate(
...
+                PrimaryRoleRecoveryAttemptState::NoRecoveryAttempt => None,
+                PrimaryRoleRecoveryAttemptState::RecoveryAttempt(recovery_proposal) =>
+                    Some(Box::new(models::PrimaryRoleRecoveryAttempt {
+                        recovery_proposal: Box::new(to_api_recovery_proposal(
+                            context,
+                            recovery_proposal
+                        )?)
+                    })),
+            },
+            has_primary_role_badge_withdraw_attempt: matches!(
```
dhedey, line 51: Ditto, and the match below

dhedey (approved):
Looks good. Some minor code tweak suggestions.

core-rust/core-api-server/src/core_api/conversions/substates/access_controller.rs
jakrawcz-rdx: We don't actually have a way (in the Core API) to indicate this difference.
There was some idea to do it, but a PR was abandoned, since it did not solve any client problem - see https://github.com/radixdlt/babylon-node/pull/600#issuecomment-1687816432

But let me apply the weak naming convention that we have for "owned vaults" (see `rewards_vault`, `stake_xrd_vault` and more): drop the `_entity` 👍

core-rust/core-api-server/src/core_api/conversions/substates/access_controller.rs
jakrawcz-rdx: true; I forget how useful this is 👍

PR #636, opened 2023-08-29
Core api: lts account resultant

malexandru-rdx:
Implement LTS account resultant balances by processing receipt's substate changes and reading vault resource address (which is assumed to never change).


core-rust/core-api-server/src/core_api/conversions/lts.rs
```diff
@@ -365,13 +371,99 @@ pub fn to_api_lts_fungible_resource_balance_change(
...
+                        &RESOURCE_PACKAGE,
+                        FUNGIBLE_VAULT_BLUEPRINT,
+                    )),
+                "Vault's TypeInfo wrongly says this is not a fungible vault."
+            );
+
+            let resource_address =
+                ResourceAddress::new_or_panic(vault_type_info.get_outer_object().into());
+
+            let fungible_vault_balance_substate: FieldSubstate<VersionedFungibleVaultBalance> =
```
jakrawcz-rdx, line 436: This line works probably only thanks to incidental `sbor(transparent)`.
I recommend deserializing `FungibleVaultBalanceFieldSubstate` (which is `FieldSubstate<FungibleVaultBalanceFieldPayload>`) and then doing `.into_payload().into_latest()`

core-rust/core-api-server/src/core_api/conversions/lts.rs
```diff
@@ -365,13 +371,99 @@ pub fn to_api_lts_fungible_resource_balance_change(
...
+                return None;
+            }
+
+            if substate_reference.2 != FungibleVaultField::Balance.into() {
+                return None;
+            }
+
+            let account_address = node_id_to_ancestor.get(vault_id).unwrap().parent.0;
+            let Ok(account_address) = GlobalAddress::try_from(account_address) else {
+                return None;
```
jakrawcz-rdx, line 407: Are we sure that we will always be interested only in the global accounts? (the field name or doc does not mention "global")

E.g. I wonder how `InternalAccount`s play with this.

core-rust/core-api-server/src/core_api/conversions/lts.rs
```diff
@@ -365,13 +371,99 @@ pub fn to_api_lts_fungible_resource_balance_change(
...
+            };
+
+            Some((account_address, resource_address, resultant_balance))
+        })
+        .for_each(|(account_address, resource_address, resultant_balance)| {
+            resultant_fungible_balances_by_account
+                .entry(account_address)
+                .or_insert(Vec::new())
+                .push(models::LtsResultantFungibleBalance {
+                    resource_address: to_api_resource_address(context, &resource_address).unwrap(),
```
jakrawcz-rdx, line 456:
```suggestion
                    resource_address: to_api_resource_address(context, &resource_address)?,
```

core-rust/core-api-server/src/core_api/conversions/lts.rs
```diff
@@ -365,13 +371,99 @@ pub fn to_api_lts_fungible_resource_balance_change(
...
+                    resultant_balance: to_api_decimal(&resultant_balance),
+                })
+        });
+    resultant_fungible_balances_by_account
+        .into_iter()
+        .map(|entry| models::LtsResultantAccountFungibleBalances {
+            account_address: to_api_global_address(context, &entry.0).unwrap(),
```
jakrawcz-rdx, line 463:
```suggestion
            account_address: to_api_global_address(context, &entry.0)?,
```

core-rust/core-api-server/src/core_api/conversions/lts.rs
```diff
@@ -365,13 +371,99 @@ pub fn to_api_lts_fungible_resource_balance_change(
...
+    let ancestries = database
+        .batch_get_ancestry(fungible_vaults.clone())
+        .into_iter()
+        .map(|record| record.expect("InternalFungibleVault does not have a parent"));
+    let node_id_to_ancestor: HashMap<_, _> = fungible_vaults.into_iter().zip(ancestries).collect();
+    let system_db_reader = SystemDatabaseReader::new(database);
+    let mut resultant_fungible_balances_by_account = HashMap::new();
+    substate_changes
+        .iter()
+        .filter_map(|(substate_reference, change_action)| {
```
jakrawcz-rdx, line 391: can we elegantly extract at least some large part from this closure? (into some `calculate_resultant_balance()`)

core-rust/core-api-server/src/core_api/conversions/lts.rs
```diff
@@ -365,13 +371,99 @@ pub fn to_api_lts_fungible_resource_balance_change(
...
+
+            if substate_reference.1 != MAIN_BASE_PARTITION {
+                return None;
+            }
+
+            if substate_reference.2 != FungibleVaultField::Balance.into() {
+                return None;
+            }
+
+            let account_address = node_id_to_ancestor.get(vault_id).unwrap().parent.0;
```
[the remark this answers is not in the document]
jakrawcz-rdx: hmm, I see... so in order to catch the broken assumption quicker (which will happen at some future Scrypto version _with certainty_, given how wildly things change there), how about we have `if parent != root { panic(super-detailed message) }` somewhere here?

core-rust/core-api-server/src/core_api/conversions/lts.rs
malexandru-rdx: As discussed on slack, `InternalAccount`s are not used in any native component and the aim of LTS is to simplify for accounts specifically.

core-rust/core-api-server/src/core_api/conversions/lts.rs
malexandru-rdx: I can change it into `.expect("some more clearer message")` but I would not pass that up as an "recoverable" error. I assume mapping error means that you (the client) passed up a wrong address, not that some random internal logic failed because invariants were broke. This clearly indicates something is wrong with the state, I think a panic is more suited.

core-rust/core-api-server/src/core_api/conversions/lts.rs
malexandru-rdx: Nice catch, thanks!

core-rust/core-api-server/src/core_api/conversions/lts.rs
malexandru-rdx: I'm not sure I can extract anything (reusable/useful), however I've refactored it to use multiple `filter`, `filter_map` and `map`s which are clearly showing the progression from `(substate_reference, change_action)` into `(account_address, resource_address, resultant_balance)`. 

I hope this is an improvement.

dhedey (approved):
I think it's okay for merging as-is - although in retrospect, I do wonder if we should instead be calculating these at balance change resolution time (see one of my comments). Would be interested to hear thoughts on that.

core-rust/core-api-server/src/core_api/conversions/lts.rs
```diff
@@ -365,13 +372,111 @@ pub fn to_api_lts_fungible_resource_balance_change(
...
+        .iter()
+        .filter(|(substate_reference, _)| substate_reference.0.is_internal_fungible_vault())
+        .filter(|(substate_reference, _)| substate_reference.1 == MAIN_BASE_PARTITION)
+        .filter(|(substate_reference, _)| {
+            substate_reference.2 == FungibleVaultField::Balance.into()
+        })
+        .filter_map(|(substate_reference, change_action)| {
+            let vault_id = &substate_reference.0;
+
+            let ancestor_record = node_id_to_ancestor
```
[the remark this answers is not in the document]
malexandru-rdx: This is just a map from already retreived (in bulk) records at line 383. It also filters fungible vault node_ids and deduplicates them to be slightly more performant.

core-rust/core-api-server/src/core_api/conversions/lts.rs
```diff
@@ -365,13 +372,111 @@ pub fn to_api_lts_fungible_resource_balance_change(
...
+            Some((account_address, substate_reference, change_action))
+        })
+        .map(|(account_address, substate_reference, change_action)| {
+            let vault_id = &substate_reference.0;
+
+            let vault_type_info = system_db_reader
+                .get_type_info(vault_id)
+                .expect("Vault missing TypeInfo substate");
+
+            let TypeInfoSubstate::Object(vault_type_info) = vault_type_info else {
```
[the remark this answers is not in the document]
jakrawcz-rdx: > now that we compute the balance changes in the node

The Engine still computes "vault balance changes" ([here](https://github.com/radixdlt/radixdlt-scrypto/blob/94cb8f863b4deb8c452f82bb1431e2d631571dea/radix-engine/src/transaction/state_update_summary.rs#L53)) and the Node only aggregates them "by global address" ([here](https://github.com/radixdlt/babylon-node/blob/c987b192f455f1b4792a45c62d1b83f413c65daf/core-rust/state-manager/src/staging/result.rs#L283)).

So actually we'd need a change in the Engine first.

However, with that change, we could reduce this >100 lines `to_api_lts_resultant_account_fungible_balances()` to <25 lines:
image

So, this seems beneficial (assuming it is conceptually ok to return "a balance change accompanied by resultant balance but only for fungibles" from the Engine). Technically, we could also resolve the "resultant balance" bit from `substate_changes` inside `compute_global_balance_changes_update()` and keep this Node-only...

jakrawcz-rdx (commented):
This also needs an apidoc update: https://github.com/radixdlt/babylon-node/blob/0d1e5b03383591193d869009be157ae8486c1473/core-rust/core-api-server/core-api-schema.yaml#L5537

PR #639, opened 2023-08-29
Add enable_unbounded_endpoints core-api flag

malexandru-rdx:
Add `api.core.flags.enable_unbounded_endpoints` config entry (default is `true`, docker env var `RADIXDLT_CORE_API_FLAGS_ENABLE_UNBOUNDED_ENDPOINTS`) which disables all unbounded endpoints (mostly ones that use `dump_component_state` + `/mempool/list`).

### DIFF
--- a/core-rust/core-api-server/src/core_api/errors.rs
+++ b/core-rust/core-api-server/src/core_api/errors.rs
@@ -133,7 +133,7 @@ pub(crate) fn assert_unbounded_endpoints_flag_enabled<E: ErrorDetails>(
 ) -> Result<(), ResponseError<E>> {
     if !state.flags.enable_unbounded_endpoints {
         return Err(client_error(
-            "This endpoint is unbounded and `enable_unbounded_endpoints` flag is disabled.",
+            "This endpoint is disabled as the response is potentially unbounded, and this node is configured with `enable_unbounded_endpoints` false.",
         ));
     }
     Ok(())
--- a/core-rust/core-api-server/src/core_api/errors.rs
+++ b/core-rust/core-api-server/src/core_api/errors.rs
@@ -9,7 +9,7 @@ use hyper::StatusCode;
 use radix_engine_interface::network::NetworkDefinition;
 use tower_http::catch_panic::ResponseForPanic;
 
-use super::models;
+use super::{models, CoreApiState};
 use models::{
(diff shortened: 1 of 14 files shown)


docker/config/default.config.envsubst
```diff
@@ -26,9 +26,11 @@ mempool.reevaluation.max_count=${RADIXDLT_MEMPOOL_REEVALUATION_MAX_COUNT}
 protocol.vertex.max_transaction_count=${RADIXDLT_PROTOCOL_VERTEX_MAX_TRANSACTION_COUNT}
 protocol.vertex.max_total_transactions_size=${RADIXDLT_PROTOCOL_VERTEX_MAX_TOTAL_TRANSACTIONS_SIZE}
 protocol.vertex.max_total_execution_cost_units_consumed=${RADIXDLT_PROTOCOL_VERTEX_MAX_TOTAL_EXECUTION_COST_UNITS_CONSUMED}
+protocol.vertex.max_total_finalization_cost_units_consumed=${RADIXDLT_PROTOCOL_VERTEX_MAX_TOTAL_FINALIZATION_COST_UNITS_CONSUMED}
```
malexandru-rdx, line 29: (Note from author)

This is a drive-by fix probably useful for `rcnet-v3`.

malexandru-rdx (commented):
After looking at the substates structure for the traversed components, it seems that `/state/validator` is indeed bounded (a big state, but bounded).

core-rust/core-api-server/src/core_api/errors.rs
```diff
@@ -128,6 +128,17 @@ pub(crate) fn assert_matching_network<E: ErrorDetails>(
     Ok(())
 }
 
+pub(crate) fn assert_unbounded_endpoints_flag_enabled<E: ErrorDetails>(
+    state: &CoreApiState,
+) -> Result<(), ResponseError<E>> {
+    if !state.flags.enable_unbounded_endpoints {
+        return Err(client_error(
+            "This endpoint is unbounded and `enable_unbounded_endpoints` flag is disabled.",
```
dhedey, line 136: Looks nice, although technically it's the response, not the endpoint itself which is unbounded - so I would slightly rather: "This endpoint is disabled as the response from this endpoint is potentially unbounded, and this node is configured with `enable_unbounded_endpoints` false"

dhedey (approved):

PR #643, opened 2023-08-30
Refactor: precomputing resultant fungible account balances together with global balance changes

jakrawcz-rdx:
As promised in https://github.com/radixdlt/radixdlt-scrypto/pull/1415#issuecomment-1699649478.
The goal is to simplify LTS endpoint's, as noticed in https://github.com/radixdlt/babylon-node/pull/636#discussion_r1310031671.

The PR is "more lines deleted than added", which weakly proves simplification.
The `LtsTransactionOutcomesTest.test_resultant_account_balances()` keeps passing verbatim, which weakly proves correctness.

### DIFF
--- a/core-rust/state-manager/src/staging/result.rs
+++ b/core-rust/state-manager/src/staging/result.rs
@@ -73,6 +73,7 @@ use crate::{
     StateHash, StateVersion, TransactionTreeHash,
 };
 use radix_engine::blueprints::consensus_manager::EpochChangeEvent;
+use radix_engine::blueprints::resource::{FungibleVaultBalanceFieldSubstate, FungibleVaultField};
 use radix_engine::transaction::{
     AbortResult, BalanceChange, CommitResult, CostingParameters, RejectResult,
     TransactionFeeSummary, TransactionReceipt, TransactionResult,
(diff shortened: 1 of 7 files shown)


core-rust/state-manager/src/staging/result.rs
```diff
@@ -373,29 +376,98 @@ impl ProcessedCommitResult {
...
             global_balance_changes
-                .entry(global_ancestor_address)
+                .entry(root_address)
                 .or_insert_with(index_map_new::<ResourceAddress, BalanceChange>)
                 .entry(*resource_address)
                 .and_modify(|existing| *existing = existing.clone() + balance_change.clone())
                 .or_insert_with(|| balance_change.clone());
+
+            // Collect (i.e. not sum) resultant balances for fungible resources of global accounts.
+            if vault_id.is_internal_fungible_vault() && root_address.is_account() {
```
dhedey, line 431: Can we also check the ancestry that the vault has a parent substate in the `AccountPartitionOffset::ResourceVaultKeyValue` partition?

Ie that `ancestry.unwrap().parent.1 == AccountPartitionOffset::ResourceVaultKeyValue.as_main_partition()`?

This will protect us against the possibility of Account getting other vaults in future, or EG the royalty vault coming back(!)

dhedey (approved):
Nice, looks really good to me, and keeps this endpoint being a lot more efficient.

core-rust/state-manager/src/staging/result.rs
dhedey: I'll make these changes and merge

PR #647, opened 2023-09-01
Consensus-messaging-related performance tweaks

LukasGasior1:
This includes a few tweaks:
- `UInt192` json/dson serialziation using `byte[]`, instead of `String`
- Use `LruCache` for `verifiedCertificatesCache`
- Don't dson-encode (and hash) VoteData separately for each signature in a QC
- Additional metrics for outbound messages
- increase max pacemaker timeout to `(1.2^13)×3 ~= 32s`

### DIFF
--- a/core/src/main/java/com/radixdlt/messaging/core/MessageCentralImpl.java
+++ b/core/src/main/java/com/radixdlt/messaging/core/MessageCentralImpl.java
@@ -70,6 +70,7 @@ import com.google.inject.Inject;
 import com.google.inject.Provider;
 import com.radixdlt.addressing.Addressing;
 import com.radixdlt.messaging.MaxMessageSize;
+import com.radixdlt.messaging.consensus.ConsensusEventMessage;
 import com.radixdlt.monitoring.Metrics;
 import com.radixdlt.p2p.NodeId;
 import com.radixdlt.p2p.PeerControl;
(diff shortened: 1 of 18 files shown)


core/src/main/java/com/radixdlt/messaging/core/MessageCentralImpl.java
```diff
@@ -222,13 +224,36 @@ public void close() {
   @Override
   public void send(NodeId receiver, Message message) {
     final var event = new OutboundMessageEvent(receiver, message, System.nanoTime() - timeBase);
-    if (!outboundQueue.offer(event) && outboundLogRateLimiter.tryAcquire()) {
-      log.error("Outbound message to {} dropped", receiver);
+    if (outboundQueue.offer(event)) {
+      final String typeForMetrics;
```
jakrawcz-rdx, line 228: (readability only)

in metrics:
`public enum OutboundMessageType {`

here:
`OutboundMessageType::from(message)`

jakrawcz-rdx (approved):
This is far outside my area of expertise, but I give 🟢 to whatever I understood.

core/src/main/java/com/radixdlt/messaging/core/MessageCentralImpl.java
LukasGasior1: hmh there are quite a few of these messages, not sure if it's worth this huge match statement?

LukasGasior1: _(happy to add it in a follow up PR if you think it's worth it_)

PR #649, opened 2023-09-04
Core api: add previous_state_identifiers for /stream/transaction

malexandru-rdx:


core-rust/core-api-server/src/core_api/handlers/stream_transactions.rs
```diff
@@ -51,10 +51,27 @@ pub(crate) async fn handle_stream_transactions(
             Please note the resync will take a while.",
         ));
     }
+    let previous_state_identifiers = match from_state_version.previous() {
```
dhedey, line 54: Might be nice to have a mini integration test for this - eg stream 5 transactions, then another 5, and check the identifiers agree between both

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -6778,6 +6778,11 @@ components:
         - max_ledger_state_version
         - transactions
       properties:
+        previous_state_identifiers:
+          $ref: "#/components/schemas/CommittedStateIdentifier"
+          description: |
+            Identifiers for the state on top of which transactions are executed.
```
dhedey, line 6784: May I suggest the following?

```
Identifiers for the state on top of which the returned transactions were executed (ie `from_state_version - 1`).
This should be used for sanity-checking that you're reading from the ledger history you're expecting.
If this is field is missing, the previous state does not exists (`from_state_version` is 0).
```

dhedey (approved):
Nice, thanks

PR #650, opened 2023-09-05
Warn if any protocol.** property is set

LukasGasior1:


core/src/main/java/com/radixdlt/RadixNodeModule.java
```diff
@@ -383,4 +406,12 @@ protected void configure() {
             .orElse(LedgerSyncCapability.Builder.asDefault());
     install(new CapabilitiesModule(builder.build()));
   }
+
+  private void warnProtocolPropertySet(String prop) {
+    log.warn(
+        "WARNING: A {} property was set. It is highly NOT recommended to modify any protocol.**"
+            + " properties as this may cause your node to disagree with the rest of the network"
+            + " and, as a result, miss proposals!",
```
malexandru-rdx, line 414: It is even more dangerous than just missing (some) proposals. The store itself can easily diverge: proposed vertex contains transactions that are not rejected because of increased limits - right now this is fixable only by resyncing the whole node.

LukasGasior1: hmh, I don't think that's the case? diverging vertex will ultimately be replaced by a new root vertex as the ledger is being synced

dhedey (approved):

PR #654, opened 2023-09-06
Allow to prolong the round timeout if scheduledLocalTimeout event was received

LukasGasior1:


core/src/main/java/com/radixdlt/consensus/liveness/Pacemaker.java
```diff
@@ -411,8 +411,7 @@ private boolean canRoundTimeoutBeProlonged(ScheduledLocalTimeout originalSchedul
         this.highestReceivedProposalRound.gte(currentRound());
 
     return (receivedAnyQcForThisOrHigherRound || receivedProposalForThisOrFutureRound)
-        && timeoutCalculator.additionalRoundTimeIfProposalReceivedMs() > 0
-        && !scheduledRoundTimeoutHasOccurred;
```
LukasGasior1, line 415: Round prolongation (if possible) is triggered by the occurrence of `ScheduledRoundTimeout`, and so is this method (to check if the round can be prolonged). So this condition was always false (because `scheduledRoundTimeoutHasOccurred` was always set to true before calling this method).

Other conditions, earlier in this method, already verify what I believe was meant by this buggy condition, that is:
- round can only be prolonged once (`originalScheduledLocalTimeout.hasBeenProlonged()` must be false)
- round can only be prolonged if an actual timeout hasn't yet been processed (`roundStatus` must be `RoundStatus.UNDISTURBED`)

dhedey (approved):

PR #659, opened 2023-09-07
add proposer_timestamp_ms in CommittedTransactionIdentifiers (RocksDB…

malexandru-rdx:
…) and CommittedTransaction (Core API)

Addresses [item](https://whimsical.com/babylon-work-tracker-Am2s7hSa7xHd5yCxgub67a@2bsEvpTYSt1HizRuCs9TzPXZVS4i4a37NdJ)

### DIFF
--- a/core-rust/state-manager/src/state_computer.rs
+++ b/core-rust/state-manager/src/state_computer.rs
@@ -986,6 +986,14 @@ where
         // Step 1.: Parse the transactions (and collect specific metrics from them, as a drive-by)
         let mut prepared_transactions = Vec::new();
         let mut leader_round_counters_builder = LeaderRoundCountersBuilder::default();
+        let mut proposer_timestamps = Vec::new();
+        let mut proposer_timestamp_ms = self
+            .store
+            .read()
+            .get_last_proof()
+            .unwrap()
+            .ledger_header
+            .proposer_timestamp_ms;
         for (index, raw_transaction) in commit_request.transactions.iter().enumerate() {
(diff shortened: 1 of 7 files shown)


core-rust/state-manager/src/state_computer.rs
```diff
@@ -1006,10 +1014,12 @@ where
                     .expect("the same transaction was parsed fine above");
                 if let LedgerTransaction::RoundUpdateV1(round_update) = round_update {
                     leader_round_counters_builder.update(&round_update.leader_proposal_history);
+                    proposer_timestamp_ms = round_update.proposer_timestamp_ms;
```
LukasGasior1, line 1017: this is maybe a tiny bit of an inconsistency:
- for all other transactions `proposer_timestamp_ms` is "a timestamp _before_ (and during) the execution of this tx"
- but for RoundUpdates it is: "a timestamp _after_ the execution of this tx"

not a big deal, but it just caught my eye :smile:

LukasGasior1 (approved):
Looks good!

core-rust/state-manager/src/state_computer.rs
malexandru-rdx: I'm assuming `proposer_timestamp_ms`'s Substate is updated only at `RoundUpdateV1` transactions. In which case `proposer_timestamp_ms` can be seen as "proposer timestamp ms at state version X (after we executed & applied the transaction)", is this right?

PR #665, opened 2023-09-08
Replacing a single OrdMap with a 2x HashMap in staging cache

jakrawcz-rdx:
This is a 100% pure refactoring (hoping no behavior change).

This PR follows a "tiered storage" approach. The goal is to allow for easy "partition delete" impl in the staging cache.

In short, we replace:
`substate_updates: ImmutableOrdMap<(DbPartitionKey, DbSortKey), DatabaseUpdate>`
with
`substate_updates: ImmutableHashMap<DbPartitionKey, ImmutableHashMap<DbSortKey, DatabaseUpdate>>`

### DIFF
--- a/core-rust/state-manager/src/staging/cache.rs
+++ b/core-rust/state-manager/src/staging/cache.rs
@@ -64,7 +64,7 @@
 
 use super::stage_tree::{Accumulator, Delta, DerivedStageKey, StageKey, StageTree};
 use super::ReadableStore;
-use std::collections::Bound::{Included, Unbounded};
+use std::iter;
 
 use crate::accumulator_tree::storage::{ReadableAccuTreeStore, TreeSlice};
 use crate::staging::{
@@ -73,8 +73,8 @@ use crate::staging::{
 };
 use crate::{EpochTransactionIdentifiers, ReceiptTreeHash, StateVersion, TransactionTreeHash};
 use im::hashmap::HashMap as ImmutableHashMap;
+use itertools::Itertools;
 
-use im::ordmap::OrdMap as ImmutableOrdMap;
 use radix_engine_common::prelude::NodeId;
 
 use radix_engine_store_interface::db_key_mapper::SpreadPrefixKeyMapper;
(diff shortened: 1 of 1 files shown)


jakrawcz-rdx:
For rationale behind this, here is a list of alternatives that I consulted with @malexandru-rdx :

**A) [emulation]** `substate_updates.range(partition_key, partition_key + 1) -> for each key { substate_updates.insert(key, Delete) }`
   - bulletproof (literally translates the batch op to a series of individual ops)
   - O(n)
   - super-lame

**B) [no rewrite]** add another `deleted_partitions: ImmutableHashSet<PartitionKey>`. On delete, add stuff there. When getting substate/ listing partition, first check it is not deleted. When putting substate, panic if partition was deleted.
   - this is allowed, since our current use-case (i.e. transaction hash tracker inside Engine) never writes again to the same partition (or "it may re-use a partition, but after long time, enough that the delete is rotated away from the Node's staging tree", as I've heard)
   - O(1) delete + O(1) time added to every get/put/list

**C) [tiered]** migrate `substate_updates: ImmutableOrdMap<DbSubstateKey, DatabaseUpdate>` to `ImmutableHashMap<DbPartitionKey, ImmutableHashMap<DbSortKey, DatabaseUpdate>>` . Listing is easy, and delete partition is obvious.
   - O(1) delete
   - put/get actually goes down from O(logn) to 2 * O(1)
   - listing goes up from O(logn) + O(n) to O(nlogn)
     - _(alternatively, the inner map may be `ImmutableOrdMap`, but I don't think it's a good trade-off)_

**D) [versioned]** do the same as **B)** but use `substate_updates: ImmutableOrdMap<DbSubstateKey, VersionedDatabaseUpdate>` and `deleted_partitions: ImmutableHashSet<VersionedPartitionKey>`. Basically, for every substate update and deleted partition, keep a version number at which it happened. Then, inserts to deleted partitions is allowed. When getting/listing, we only return a substate if its version is higher than version of the partition delete operation.
   - characteristics same as **B)**
   - however, worst-case of listing is funny :)

---

And this PR prepares us for **C)** that we picked.

malexandru-rdx (approved):
As discussed on slack, we might need to change to `ImmutableHashMap<DbPartitionKey, ImmutableOrdMap<DbSortKey, DatabaseUpdate>>` so that the iterator construction goes from `O(Ps log Ps)` complexity, where `Ps = partitions in staging` back to `O(log Ps)` (along side with a benchmark). This is going to be exacerbated once we will keep long/er histories of the staged store.

PR #667, opened 2023-09-11
Implementing partition delete for JMT store, substate store and staging cache

jakrawcz-rdx:
This also bumps the scrypto dep to tag `develop-f00e2d48`.

Regarding the handling of partition delete:
- the most involved part is in the staging cache: `ImmutableHashMap<DbPartitionKey, ImmutableHashMap<DbSortKey, DatabaseUpdate>>` became `ImmutableHashMap<DbPartitionKey, ImmutablePartitionUpdates>` (and the `ImmutablePartitionUpdates` is now the place which can capture a "partition reset" operation)
- in general, we now have a couple of places in Node's code (+ in Engine's) which look very similar (i.e. `match PartitionUpdates, apply Reset on top of Delta, or Delta on top of Reset, or Delta on top of Delta, ...`) but I cannot find an easy way to extract the common logic out (they differ because e.g. staging cache needs to use `im:*` structures).


jakrawcz-rdx:
This is now ready for review.

HOWEVER, this related aspects are out of scope of this PR (will come soon):
- including the partition deletes in a merkle hash of transaction's state changes
- returning the partition deletes from Core API

malexandru-rdx (approved):
LGTM!

core-rust/state-manager/src/staging/cache.rs
```diff
@@ -420,17 +431,108 @@ impl<K, N> AccuTreeDiff<K, N> {
...
+                } => {
+                    for (db_sort_key, database_update) in db_substate_updates {
+                        match database_update {
+                            DatabaseUpdate::Set(value) => {
+                                new_substate_values.insert(db_sort_key.clone(), value.clone());
+                            }
+                            DatabaseUpdate::Delete => {
+                                let existed = new_substate_values.remove(db_sort_key).is_some();
+                                if !existed {
+                                    panic!("broken invariant: deleting non-existent substate");
```
LukasGasior1, line 513: _(discussed on a call, but for the record: we should double check that the same invariant always holds in the engine)_

LukasGasior1 (approved):
Looks good!

PR #670, opened 2023-09-11
Add more validator address / seed nodes related checks and logs

LukasGasior1:


core/src/main/java/com/radixdlt/p2p/discovery/PeerDiscovery.java
```diff
@@ -136,6 +136,12 @@ public EventProcessor<DiscoverPeers> discoverPeersEventProcessor() {
       final var seedNodes = seedNodesConfigParser.getResolvedSeedNodes();
       this.addressBook.addUncheckedPeers(seedNodes);
 
+      if (seedNodes.isEmpty() && this.addressBook.knownPeers().isEmpty()) {
+        log.warn(
+            "Warning! No valid seed nodes have been configured for this node."
```
jakrawcz-rdx, line 141: "... and no peers in address book" ?
(or the `if` should only check `seedNodes`?)

jakrawcz-rdx (approved):
As always, I can only recommend `"""` for those long literals (but not strictly required).
AFAIS, it is mostly moving code around and adding a few `if ... then log(...)`, looks safe 👍

core/src/main/java/com/radixdlt/p2p/discovery/PeerDiscovery.java
LukasGasior1: I did that on purpose :)
- if there are no seed nodes, but some peers have been added to addr book (e.g. other peers connected to us and address-book-synced), then it's fine - no need to log
- if there are no peers in the address book (yet) BUT there are some valid seed nodes - it's fine too, no need to log 
- if there are no peers in the address book AND none were configured - it's a warning, but the root cause is missing seed nodes (so no need to include "... and no peers in address book")

LukasGasior1:
oh god, how have I missed it? :cold_sweat: thanks @jakrawcz-rdx

jakrawcz-rdx:
Fixed by https://github.com/radixdlt/babylon-node/pull/672

PR #671, opened 2023-09-13
Including partition deletes in tx state changes merkle hash + Returning them from Core API

jakrawcz-rdx:
As mentioned at https://github.com/radixdlt/babylon-node/pull/667#issuecomment-1716044067.


malexandru-rdx (approved):
Looks good.
I'm wondering what data layer is this supposed to be? It was a bit confusing/had to look more how this translates to `NodeId`s (or rather how those translate into this).

jakrawcz-rdx:
> I'm wondering what data layer is this supposed to be? 

As discussed on Slack - most confusion comes from different (logical) representations of our substate "hierarchy" - sometimes we act as if it is a 3-Tier `(NodeId, PartitionNum, SortKey)` and sometimes we as 2-Tier `(PartitionKey, SortKey)`. In this PR I chose to represent the CoreAPI's `PartitionId` simply as a subset (prefix) of fields present in `SubstateId` (which is kinda rich: `entity_type, entity_address, entity_module, partition_kind, partition_number`, but should be "least surprise" for the API consumer)

jakrawcz-rdx:
I am going to merge this PR just for the sake of "not keeping PRs opened for too long".
@dhedey you may still want to look at how the `[deleted_partitions, changed_substates].concat()` are fed to the transient "state changes" merkle tree, since that is the most arbitrary choice I had to make here.

PR #673, opened 2023-09-14
Add more vote/proposal metrics

LukasGasior1:


core/src/main/java/com/radixdlt/consensus/liveness/Pacemaker.java
```diff
@@ -308,6 +308,24 @@ private void dispatchVote(Vote vote) {
       this.voteDispatcher.dispatch(this.latestRoundUpdate.getNextLeader(), vote);
     }
 
+    // This should always be Some(), we've just sent a vote for this vertex
+    // so it must have been successfully executed!
+    // But better not to unwrap() just to bump the metrics, even if it's safe :)
+    this.vertexStore
+        .getExecutedVertex(vote.getVoteData().getProposed().getVertexId())
```
jakrawcz-rdx, line 315: do we have to reach to `vertexStore` in order to bump a metric?
Could the callers pass the `isFallback` information? (I see that one of them knows the entire `executedVertex` and the other one only always does fallbacks)

_[I know that passing another param "for metrics purposes only" is not perfect either, but at least we won't ever have to wonder "hmm maybe the vertexStore has some bug and that's why we don't see this metric bumped"]_

LukasGasior1: unfortunately we don't have the vertex at hand when resending the previous vote (`resendPreviousVoteWithTimeoutOrVoteForFallbackVertex`) - this can be either:
- resending the vote for a regular vertex with a timeout flag
- resending the vote for a fallback vertex with a timeout flag
- creating and voting for a new fallback vertex, also with a timeout flag (the "orElse" part)

we only know it's a fallback in the last case.

I assume your concern is just that it doesn't look too good? :smile: performance-wise this is just a hashmap lookup

jakrawcz-rdx: > I assume your concern is just that it doesn't look too good?

Yeah, and the comment `// But better not to unwrap() just to bump the metrics, even if it's safe :)` also suggests that some hack takes place.
But you are right, it doesn't hurt that much :p

jakrawcz-rdx (approved):

PR #674, opened 2023-09-14
Disable delayed TC resolution

LukasGasior1:
Disable delayed TC resolution - it doesn't really help in most cases.


malexandru-rdx (approved):
LGTM!

PR #677, opened 2023-09-15
Adding versioning wrappers to applicable rocksdb tables

jakrawcz-rdx:
Addresses https://whimsical.com/babylon-work-tracker-Am2s7hSa7xHd5yCxgub67a@2bsEvpTYSt1Hj7bgjUoDDz1bUNSz1Kr3irs .

On the way, I tried to clean-up a few naming inconsistencies. Most notably:
- I introduced a convention for our table names: a table keyed by Foo and containing Bars is called `FooToBar` (it used to be ~mostly `BarsByFoo`, which confused me a bit during my work here)
- I fixed one bug (struct X persisted, struct Y read from that same table)


core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -97,114 +98,128 @@ use crate::transaction::{
...
-            TransactionAccuTreeSliceByStateVersion => {
-                "transaction_accu_tree_slice_by_state_version"
-            }
-            ReceiptAccuTreeSliceByStateVersion => "receipt_accu_tree_slice_by_state_version",
-            ExtensionsData => "extensions_data",
-            AccountChangeStateVersions => "account_change_state_versions",
-            ExecutedGenesisScenarios => "executed_genesis_scenarios",
-        };
-        write!(f, "{str}")
+        f.write_str(self.into())
```
jakrawcz-rdx, line 211: _(note from author)_

If there is some legitimate reason why it was done the manual way before - please shout!

jakrawcz-rdx: I think I understand... any change here effectively requires a ledger-wipe :(

malexandru-rdx (approved):

core-rust/state-manager/src/store/rocks_db.rs
jakrawcz-rdx: As discussed with @malexandru-rdx on Slack:
It's important to not make it refactoring-sensitive. Especially with people on the team that love refactorings. Hence I bring back the manual mappings here.

jakrawcz-rdx: (I also changed some names to be more like "naming a concept" and not "describing the schema", since these should be as future proof as possible)

jakrawcz-rdx:
since we resolved the only doubt (at https://github.com/radixdlt/babylon-node/pull/677#discussion_r1330328851), and the rest of the PR is a mechanical `VersionedX::new_from_latest(x)`, and the CI is as 🟢  as it gets - let me merge

PR #678, opened 2023-09-18
tweak: Remove SBOR JSON by default from the transaction stream

dhedey:
# Summary

* After consulting Builders and Gateway teams, this removes the JSON encoding of SborData in the Core API - just for the stream transactions endpoint. This results in additional work and lots of additional bandwidth, so cutting this out should make the API calls more efficient.
* This PR also turns on typed models for the transaction message by default.

### DIFF
--- a/core-rust/core-api-server/src/core_api/generated/models/sbor_format_options.rs
+++ b/core-rust/core-api-server/src/core_api/generated/models/sbor_format_options.rs
@@ -17,7 +17,7 @@ pub struct SborFormatOptions {
     /// Whether to return the raw hex-encoded bytes (default true)
     #[serde(rename = "raw", skip_serializing_if = "Option::is_none")]
     pub raw: Option<bool>,
-    /// Whether to return the programmatic json format (default true)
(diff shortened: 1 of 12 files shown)


core-rust/core-api-server/src/core_api/generated/models/sbor_format_options.rs
```diff
@@ -17,7 +17,7 @@ pub struct SborFormatOptions {
     /// Whether to return the raw hex-encoded bytes (default true)
     #[serde(rename = "raw", skip_serializing_if = "Option::is_none")]
     pub raw: Option<bool>,
-    /// Whether to return the programmatic json format (default true)
+    /// Whether to return the programmatic json format (normally default true, defaults false for streamed transactions)
```
jakrawcz-rdx, line 20: The only improvement (over this "default is X except when it's not") that I can think of is to put some extra note in the /transactions/stream endpoint's docs (not sure where), saying "unlike others, this endpoint does not output jsons by default".

jakrawcz-rdx (approved):

malexandru-rdx (approved):

PR #681, opened 2023-09-18
Mempool reevaluation: add executed at state version index and update from result cache

malexandru-rdx:
Addresses [item](https://whimsical.com/babylon-work-tracker-Am2s7hSa7xHd5yCxgub67a@2bsEvpTYSt1Hj5k4wdPT52kVsqwe78w3bYn):
- Add mempool executed at state version index.
- Result cache now sends updates to mempool
- Mempool reevaluation picks transactions by state version (earliest first)


core-rust/state-manager/src/mempool/mempool_manager.rs
```diff
@@ -66,25 +66,36 @@ use super::metrics::MempoolManagerMetrics;
...
+}
+
+pub struct StateVersionUpdate {
+    pub payload_hash: NotarizedTransactionHash,
+    pub state_version: StateVersion,
+}
+
 /// A high-level API giving a thread-safe access to the `PriorityMempool`.
 pub struct MempoolManager {
+    deferred_updates_rx: Mutex<Receiver<MempoolUpdate>>,
```
malexandru-rdx, line 98: _(note from author)_

The `Mutex<>` is needed solely to circumvent `jni`s `Sync` requirement. The overhead is just the extra memory access (+ compare) since it will never actually wait.
In a pure Rust environment this is not needed and should be removed in future.

core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
```diff
@@ -431,14 +454,20 @@ const NON_REJECTION_RECALCULATION_DELAY: Duration = Duration::from_secs(120);
 const MAX_RECALCULATION_DELAY: Duration = Duration::from_secs(1000);
 
 pub struct PendingTransactionResultCache {
+    mempool_deferred_updates_tx: Mutex<Sender<MempoolUpdate>>,
```
malexandru-rdx, line 457: _(note from author)_

Same note about `Mutex` as with `deferred_updates_rx` from `PriorityMempool`.

core-rust/state-manager/src/mempool/mempool_manager.rs
```diff
@@ -166,36 +181,50 @@ impl MempoolManager {
...
-            max_reevaluated_count,
-            MIN_TRANSACTIONS_TO_CHECK_FOR_REEVALUATION,
-        )
-            as usize);
-
         let mut transactions_to_remove = Vec::new();
-        let mut reevaluated_count = 0;
-        for candidate_transaction in candidate_transactions {
-            let (record, was_cached) = self
+        for candidate_transaction in self
```
malexandru-rdx, line 213: _(note from author)_

Now we can directly start to reexecute transactions since all rejected transaction (hitting `PendingTransactionResultCache`) are removed via deferred updates in `execute_deferred_updates` called exactly before this.

core-rust/node-common/src/config/limits.rs
```diff
@@ -64,7 +64,6 @@
 
 use radix_engine_common::prelude::*;
 
-// TODO: revisit & tune before Babylon
```
malexandru-rdx, line 67: _(note from author)_

This has been previously addressed.

core-rust/state-manager/src/mempool/mempool_manager.rs
jakrawcz-rdx: plz paste this note's contents into `// TODO(after 100% Rust migration): ...`

core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
jakrawcz-rdx: Same TODO request as in https://github.com/radixdlt/babylon-node/pull/681#discussion_r1330620776 :p

core-rust/state-manager/src/jni/mempool.rs
```diff
@@ -176,6 +176,8 @@ extern "system" fn Java_com_radixdlt_mempool_RustMempool_reevaluateTransactionCo
 ) -> jbyteArray {
     jni_sbor_coded_call(&env, request_payload, |max_reevaluated_count: u32| {
         let mempool_manager = JNINodeRustEnvironment::get_mempool_manager(&env, j_node_rust_env);
+        // Once we have an dedicated actor thread for Mempool, move this call there.
```
jakrawcz-rdx, line 179:
```suggestion
        // TODO(after XXXXX): Do YYYYYYY.
```

core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
```diff
@@ -259,6 +262,23 @@ pub enum AtState {
     },
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct AtStateIntoStateVersionError(pub ());
```
jakrawcz-rdx, line 266: what does `(pub ())` do? how does it differ from `{ }`?

core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
```diff
@@ -338,7 +358,7 @@ impl PendingTransactionRecord {
     pub fn should_accept_into_mempool(
         self,
         check: CheckMetadata,
-    ) -> Result<Box<ValidatedNotarizedTransactionV1>, MempoolAddRejection> {
+    ) -> Result<(Box<ValidatedNotarizedTransactionV1>, StateVersion), MempoolAddRejection> {
```
jakrawcz-rdx, line 361:
```suggestion
    ) -> Result<SomeWellNamedType, MempoolAddRejection> {
```

I only suggest this since `SomeWellNamedType` would be a nice place to explain what is the returned state version.

core-rust/state-manager/src/mempool/mempool_manager.rs
```diff
@@ -166,36 +181,50 @@ impl MempoolManager {
...
+            .mempool
+            .read()
+            .iter_by_state_version()
```
jakrawcz-rdx, line 216: wait, doesn't this hold the mempool's read lock for the entire duration of the loop? (which may be quite long, if it needs to execute maaany transactions?)
shouldn't we, instead, call `.get_k_earliest_state_version_transactions(k) -> Vec<_>` ?

core-rust/state-manager/src/mempool/mempool_manager.rs
jakrawcz-rdx: anyways - is there a conceptual reason for "deferring" updates?
AFAI-vaguely-R from our Slack conversations, the Tx/Rx exists only to break some cyclic dep between the mempool and PTRCache...? If this is the case, then at the very least we need a fat comment about this in the code. Also, please give me a moment to play with it (and possibly invent some other hack for the cyclic dep, _once you confirm you don't need deferring for other reasons_)

core-rust/state-manager/src/mempool/mempool_manager.rs
malexandru-rdx: Damn, right. Nice catch!

core-rust/state-manager/src/mempool/mempool_manager.rs
malexandru-rdx: >AFAI-vaguely-R from our Slack conversations, the Tx/Rx exists only to break some cyclic dep between the mempool and PTRCache...? If this is the case, then at the very least we need a fat comment about this in the code. Also, please give me a moment to play with it (and possibly invent some other hack for the cyclic dep, once you confirm you don't need deferring for other reasons)

This is the main reason for the current goal. Either we fully tie result cache with mempool (which would add more contention) or we do this decoupling (otherwise it's impossible since we get circular dependencies). I am pretty sure we will end up having an actor based model where `MempoolManager` will have it's own thread, in which case this is no longer "deferred" work, but the channel actually used as a proper channel.

malexandru-rdx: There is also the option of redesigning the mempool and result cache. I think it makes more sense to have the results of transactions in mempool to live in the mempool and leave result cache for rejected transactions and committable transactions that don't make it into mempool (full mempool).

core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
malexandru-rdx: It does the same thing. This is the same pattern used in `std` (i.e. `pub struct TryFromIntError(pub(crate) ());`).

Here is a comprehensive stackoverflow [thread](https://stackoverflow.com/questions/50162597/what-are-the-differences-between-the-multiple-ways-to-create-zero-sized-structs). Unit-valued tuple struct (current one) is the only way to make it impossible to construct the struct outside it's crate (obviously, this is not the case for this PR as it does have a `pub`). The `{ }` is might mess with macros and generated code (see https://rust-lang.github.io/rfcs/0218-empty-struct-with-braces.html).

core-rust/state-manager/src/mempool/mempool_manager.rs
jakrawcz-rdx: @malexandru-rdx I think we can now un-pause this PR.

I gave it a try to model the dependencies here in some other way, and my suggestion is at https://github.com/radixdlt/babylon-node/pull/696 - please have a look there and feel free to merge it into this PR if you find it beneficial.

core-rust/state-manager/src/mempool/mempool_manager.rs
```diff
@@ -166,36 +163,28 @@ impl MempoolManager {
...
-        let mut reevaluated_count = 0;
         for candidate_transaction in candidate_transactions {
-            let (record, was_cached) = self
+            let (record, _was_cached) = self
                 .cached_committability_validator
                 .check_for_rejection_cached_prevalidated(
-                    &candidate_transaction.validated,
-                    ForceRecalculation::No,
+                    &candidate_transaction.transaction.validated,
+                    ForceRecalculation::Yes,
```
jakrawcz-rdx, line 183: I remember that 2 weeks ago I (thought I) knew why this change happened, but now I don't :D can you please confirm it is intentional?

malexandru-rdx: Yes, it is intentional. Before, the result cache didn't remove anything from here, hence why we preferred to only recalculate with the cache decided so. Now, before getting here, we remove from the mempool everything that is invalidated and everything left will be valid (for that state version) if we do not reexecute.

core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
```diff
@@ -282,6 +282,24 @@ pub enum RetryFrom {
     Whenever,
 }
 
+#[derive(Debug, Clone)]
+pub struct DynamicValidatedTransaction {
```
malexandru-rdx, line 286: _(note from author)_

There is also the naming option of `PendingExecutedTransaction`.

jakrawcz-rdx (approved):

core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
```diff
@@ -366,7 +399,15 @@ impl PendingTransactionRecord {
             CheckMetadata::Cached => {
                 panic!("Precondition was not met - the result was cached, but the latest attempt was not a rejection")
             }
-            CheckMetadata::Fresh(StaticValidation::Valid(transaction)) => Ok(transaction),
+            CheckMetadata::Fresh(StaticValidation::Valid(transaction)) => {
+                Ok(PendingExecutedTransaction::new(
+                    transaction,
+                    self.latest_attempt
+                        .against_state
+                        .specific_version()
```
dhedey, line 407: Storing this isn't super accurate, because we lose a bit of specificity.

I think I'd prefer to store something like `ExecutedAtState::Committed { state_version }` or `ExecutedAtState::PendingPreparingVertices { base_committed_state_version }`, even if we use the "specific version" for the inverse index.

But it's not a big deal :D

core-rust/state-manager/src/mempool/priority_mempool.rs
```diff
@@ -97,12 +98,26 @@ pub struct MempoolData {
     pub source: MempoolAddSource,
 }
 
+// There is only 1 [`MempoolData`] per unique [`MempoolTransaction`]
+impl Hash for MempoolData {
```
dhedey, line 102: This feels weird / scary to me that we have an `Eq` and `Hash` which disagree on equality checks.

I think the issue is that it seems wrong to me that `MempoolData` is a key in a hashmap.

core-rust/state-manager/src/mempool/priority_mempool.rs
```diff
@@ -97,12 +98,26 @@ pub struct MempoolData {
...
+}
+
 #[derive(Debug, Clone, Eq)]
 pub struct MempoolTransaction {
     pub validated: Box<ValidatedNotarizedTransactionV1>,
     pub raw: RawNotarizedTransaction,
 }
 
+// We can uniquely identify the whole [`MempoolTransaction`] just by the [`RawNotarizedTransaction`] hash.
+impl Hash for MempoolTransaction {
```
dhedey, line 115: Ditto - this feels weird/scary to me, and relatively expensive needing to hash a payload.

We should use a key like `NotarizedTransactionHash` instead for hashmaps.

dhedey (commented):
On reflection, I think there are one or two things that might be worth discussing before we merge this, let's chat :)

core-rust/state-manager/src/state_manager.rs
```diff
@@ -208,11 +208,23 @@ impl StateManager {
             ),
         );
 
-        let pending_transaction_result_cache = Arc::new(
+        let mut costing_parameters = CostingParameters::default();
```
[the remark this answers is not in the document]
jakrawcz-rdx: ahhh an omitted "update from develop" artifact; somehow not reported as "unused variable"...
Now removed - thanks for catching!

dhedey (approved):
Approving, we can do markups / refactors in a following PR.

core-rust/state-manager/src/transaction/validation.rs
```diff
@@ -461,38 +449,26 @@ impl CachedCommittabilityValidator {
...
     ///
     /// The result is stored in the cache.
     /// If the transaction is freshly rejected, the caller should perform additional cleanup,
     /// e.g. removing the transaction from the mempool.
     ///
-    /// Its pending transaction record is returned, along with a boolean about whether the last
-    /// attempt was cached.
-    pub fn check_for_rejection_cached_prevalidated(
+    /// Returns the transaction's new pending transaction record.
+    pub fn check_for_rejection_validated(
```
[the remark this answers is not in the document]
jakrawcz-rdx: On the surface: Yes (just a bit of wasted work - `TransactionMetadata` construction, `committability_validator.read()` lock)
There is one [blocker] gotcha though: re-captured `SystemTime::now()` (which may differ!). So we can at most use an extracted helper, which would have to accept the metadata, validator, transaction and timestamp... Not worth it, to me.

While playing here, I noticed some other ugliness - a `CommittabilityValidator.user_transaction_validator` dependency which is only used for pure delegation of certain calls (`prepare_from_raw(), validate()`). But AFAIS the "apply protocol update" logic kinda relies on that :( it does look a bit messy and brittle :( I'll try to compose my proto-update-apply concerns into a separate ticket.

core-rust/state-manager/src/mempool/mempool_manager.rs
```diff
@@ -164,36 +162,24 @@ impl MempoolManager {
...
             }
-            if was_cached == PrevalidatedCheckMetadata::Fresh {
-                reevaluated_count += 1;
-                if reevaluated_count >= max_reevaluated_count {
-                    break;
-                }
-            }
         }
 
         if !transactions_to_remove.is_empty() {
```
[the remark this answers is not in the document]
jakrawcz-rdx: wow, indeed, all these lines are no-op right now 👍 

I went through the other `mempool`-callers (of `MempoolManager`) and there was one more suspect - the `cached_committability_validator.check_for_rejection_cached()` followed by `mempool.add_transaction_if_not_present()` (inside `add_if_committable_internal()`). However, `track_transaction_result()` cannot really add transaction (only update its metadata if it already exists in the mempool). So the other methods look fine.

core-rust/state-manager/src/mempool/priority_mempool.rs
```diff
@@ -229,6 +265,10 @@ pub struct PriorityMempool {
     proposal_priority_index: BTreeSet<MempoolDataProposalPriorityOrdering>,
     /// Keeps ordering of the transactions by end epoch.
     end_epoch_exclusive_index: BTreeSet<MempoolDataEndEpochExclusiveOrdering>,
+    /// Keeps ordering of the transactions by latest successful execution
+    executed_state_version_index: BTreeSet<MempoolDataStateVersion>,
```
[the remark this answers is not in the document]
jakrawcz-rdx: yeah, it's time we discuss the complicated mempool vs ptr-cache relationship :p

PR #686, opened 2023-09-21
Implementing `/lts/state/account-deposit-behaviour` endpoint

jakrawcz-rdx:
Addresses https://whimsical.com/babylon-work-tracker-Am2s7hSa7xHd5yCxgub67a@2bsEvpTYSt1Hj3SDQYhbvr4oNy2rQseEsc6

More detailed spec agreed on Slack:
> New “lts/state/account-deposit-behaviour” takes account address, list of resource addresses (0 ≤ x ≤ 20), and optional badge for the deposit.
> Returns:
> - General account resource preference
> - Whether the badge is approved for deposits
> - For each requested resource address:
>    - What its overriden behaviour is (if at all)
>    - Whether the vault exists already for this resource
>    - [Whether the resource is XRD]
>    - The resolved ability for whether a deposit of that resource to that account with that badge can succeed.
> - ~~Whether all requested deposits resolve successfully - ie whether the deposit instruction will succeed (or abort/refund which it does atomically)~~

> Background reading: https://docs-babylon.radixdlt.com/main/scrypto/native-blueprints/accounts.html#_configuring_account_deposit_modes_and_resource_preference_map (edited)


core-rust/core-api-server/src/core_api/handlers/lts/state_account_deposit_behaviour.rs
```diff
@@ -0,0 +1,245 @@
...
+        badge_of_authorized_depositor,
+        resource_specific_behaviours,
+    })
+    .map(Json)
+}
+
+/// Resolves whether the deposit is allowed, based on raw inputs.
+/// See https://docs-babylon.radixdlt.com/main/scrypto/native-blueprints/accounts.html for a
+/// flow chart.
+fn allows_deposit(
```
jakrawcz-rdx, line 221: _(note from author)_

## Dear reviewers,

Please pay special attention to this method.

LukasGasior1 (approved):
Looks good as far as I can tell, but maybe someone from the builders team could double check the `allows_deposit` logic (@talekhinezh , @iamyulong ?)

core-rust/core-api-server/src/core_api/handlers/lts/state_account_deposit_behaviour.rs
iamyulong: Maybe rename to `allows_try_deposit` to be consistent with the blueprint?

core-rust/core-api-server/src/core_api/handlers/lts/state_account_deposit_behaviour.rs
```diff
@@ -0,0 +1,245 @@
...
+        Some(AccountResourcePreference::Disallowed) => badge_of_authorized_depositor == &Some(true),
+        // No explicit resource preference - try the same logic with the default deposit rule:
+        None => {
+            match default_deposit_rule {
+                // The default rule allows it:
+                DefaultDepositRule::Accept => true,
+                // The default rule disallows it, but a present authorized badge can override this:
+                DefaultDepositRule::Reject => badge_of_authorized_depositor == &Some(true),
+                // The extra case of the default rule, allowing existing vaults *or* XRD:
+                DefaultDepositRule::AllowExisting => vault_exists || is_xrd,
```
iamyulong, line 241: I think this is incorrect. 

Should be `vault_exists || is_xrd || badge_of_authorized_depositor == &Some(true)`?

Maybe @0xOmarA can confirm :-D

jakrawcz-rdx: OMG you are right, and even the flow chart in the official docs agrees 😱 

Basically it looks like presenting a badge of authorized depositor overrides everything, so I now just moved this condition first, PTAL.

core-rust/core-api-server/src/core_api/handlers/lts/state_account_deposit_behaviour.rs
iamyulong: It's not a big deal. Pick whichever makes more sense within your context. :)

iamyulong:
The updated `allows_deposit` logic looks good to me.

iamyulong (approved):
I only reviewed the `allows_deposit` logic. :-D

core/src/test/java/com/radixdlt/api/core/LtsAccountDepositBehaviourTest.java
```diff
@@ -0,0 +1,154 @@
...
+      assertThat(differentResult.getResourceSpecificBehaviours())
+          .isEqualTo(
+              Map.of(
+                  addressing.encode(ScryptoConstants.XRD_RESOURCE_ADDRESS),
+                  new ResourceSpecificDepositBehaviour()
+                      .resourcePreference(null)
+                      .vaultExists(true)
+                      .isXrd(true)
+                      .depositAllowed(true)));
+    }
```
dhedey, line 152: I'm slightly worried we haven't tested that they keys and payloads are serialized correctly from the stores.

I suggest the following two test cases:
* Changes default behaviour to DenyAll & a test which checks if we can submit XRD - this is most important test for the exchange use case.
* Sets default behaviour to Existing, set an authorized depositor badge and a resource preference on the account, and check that the response is correct 

We'll need to add some manifests to `Manifest.java` for these.

And then can submit those manifests with eg:
```java
      var accountKeyPair = ECKeyPair.generateNew();
      var accountAddress = Address.virtualAccountAddress(accountKeyPair.getPublicKey());
      submitAndWaitForSuccess(test, Manifest.zYZ(accountAddress), List.of(accountKeyPair));
```

dhedey (approved):
Really nice. A few minor renames / additional tests suggestions. Please fix, add tests and merge into develop when they pass.

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -5544,6 +5603,82 @@ components:
...
+      properties:
+        state_version:
+          $ref: "#/components/schemas/StateVersion"
+          description: The state version at which the query was performed.
+        ledger_header_summary:
+          $ref: "#/components/schemas/LedgerHeaderSummary"
+          description: The excerpt from the ledger header committed at the `state_version`.
+        default_deposit_rule:
+          $ref: '#/components/schemas/DefaultDepositRule'
+        badge_of_authorized_depositor:
```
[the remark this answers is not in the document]
dhedey: (Just a general principle that I prefer booleans to have verbs)

core/src/test/java/com/radixdlt/api/core/LtsAccountDepositBehaviourTest.java
jakrawcz-rdx: Added 👍 
(almost exactly the 2 scenarios you describe; anyway I am sure that they verify that "the right resource preference is read from the KV" and "the right depositor badge is read from the KV")

jakrawcz-rdx:
I only see 1 failure (the known `MultiNodeRebootTest restart_all_nodes_intermittently[2]` instability), hence I consider the CI 🟢 ; merging!

PR #687, opened 2023-09-22
Exposing the recent self proposal miss count via health endpoint

jakrawcz-rdx:
Minor last-minute improvement.

image


jakrawcz-rdx:
The newest adjustment:
image

dhedey (approved):

PR #694, opened 2023-09-25
Bump BouncyCastle to 1.70; Add more RadixKeyStore tests; Remove default KeyStore password

LukasGasior1:


common/src/main/java/com/radixdlt/crypto/RadixKeyStore.java
```diff
@@ -164,11 +155,34 @@ public final class RadixKeyStore implements Closeable {
   public static RadixKeyStore fromFile(File file, char[] storePassword, boolean create)
       throws IOException, KeyStoreException {
     try {
-      var ks = KeyStore.getInstance("pkcs12", BouncyCastleProviderInstance.get());
-      var usedStorePassword =
-          (storePassword == null || storePassword.length == 0) ? defaultKey : storePassword;
```
LukasGasior1, line 169: I really hope no one has their key password set to "radix" :smile:  and I haven't found anywhere where we would default to it at generation time or in docs. But if someone was relying on it, they'll need to update their env variable, which I think is fine?
Also, this change allows us to test an empty password.

jakrawcz-rdx (approved):

PR #696, opened 2023-09-25
Migrating from tx/rx-based deferred updates to direct calls using a shared Mempool dep

jakrawcz-rdx:
⚠️ This wants to go into https://github.com/radixdlt/babylon-node/pull/681 (it is a slight refactor which should make that behaviour change simpler)

The `Arc<RwLock<PriorityMempool>>` was already there, extracted for other purposes some time ago, so I used it.

### DIFF
--- a/core-rust/node-common/src/locks.rs
+++ b/core-rust/node-common/src/locks.rs
@@ -83,12 +83,16 @@ use tracing::{error, info};
 // at least "poisoning" the lock).
 //==================================================================================================
 
+/// A factory of panic-safe lock primitives.
 pub struct LockFactory {
     stopper: Arc<TakeOnce<Box<dyn FnOnce() + Send>>>,
     name: String,
 }
 
 impl LockFactory {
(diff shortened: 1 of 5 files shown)


malexandru-rdx:
This is nice because it removes the queue (and it's limits/size problems associated with it), making it easy to merge ASAP. 

However we will most probably revert it at some point since:
1. Channels (tx/rx) is still the best sync mechanism for actor threads.
2. The life/path of a submitted transaction should be more like in [this comment](https://github.com/radixdlt/babylon-node/pull/681#discussion_r1330706419).
3. Since this writes to mempool at every submit means more uncontrolled contention overlap.
4. This is mixing/changing abstraction and IMO requires a further refactor:
   - It degrades the role of `MempoolManager` which was responsible for writing/keeping mempool up to date
   - `PriorityMempool` is no longer a contained "leaf"
   - Drifts from a clearer responsability split into something that could easily be argued towards something monolithic (which is not the approach we want).

malexandru-rdx (approved):

PR #703, opened 2023-10-02
Add config to disable p2p discovery

LukasGasior1:


core/src/main/java/com/radixdlt/p2p/discovery/PeerDiscovery.java
```diff
@@ -179,23 +180,33 @@ public RemoteEventProcessor<NodeId, PeersResponse> peersResponseRemoteEventProce
...
   public RemoteEventProcessor<NodeId, GetPeers> getPeersRemoteEventProcessor() {
     return (sender, unused) -> {
-      final var peers =
-          Stream.concat(
-                  Stream.of(selfUri),
-                  this.addressBook.bestCandidatesToConnect().limit(MAX_PEERS_IN_RESPONSE - 1))
-              .collect(ImmutableSet.toImmutableSet());
+      final ImmutableSet<RadixNodeUri> peersToSend;
+
+      if (p2pConfig.discoveryDisabled()) {
```
jakrawcz-rdx, line 198: _(just wondering)_ I see that the same flag is used to disable dispatching discovery requests and to disable returning responses to such incoming requests. If you are sure this is the level of granularity we want, then all ok 👍

jakrawcz-rdx (approved):

core/src/main/java/com/radixdlt/p2p/discovery/PeerDiscovery.java
LukasGasior1: yup, that was my intention - disable it on both sides

PR #704, opened 2023-10-03
Release/babylon 1.0.4

LukasGasior1:
Update develop with the changes that only went to releaes/babylon-1.0.4


jakrawcz-rdx (approved):
(I only looked at "what commits go in")

PR #707, opened 2023-10-03
Adding a per-CF `raw_db_entries` and `raw_db_size` metric

jakrawcz-rdx:
Addresses https://docs.google.com/spreadsheets/d/1JPiisoLFOhrDBT1-qDj2uvVwIsojKQd8kHzf1CIhaFw/edit#gid=0&range=B2 .

Example output:
```
# HELP rn_raw_db_size An approximate size of the database, in bytes, by category of entries.
# TYPE rn_raw_db_size gauge
rn_raw_db_size{category="account_change_state_versions"} 0
rn_raw_db_size{category="committed_transaction_identifiers"} 0
rn_raw_db_size{category="epoch_ledger_proofs"} 0
rn_raw_db_size{category="executed_genesis_scenarios"} 0
rn_raw_db_size{category="extensions_data"} 0
rn_raw_db_size{category="intent_hashes"} 0
rn_raw_db_size{category="ledger_proofs"} 0
rn_raw_db_size{category="ledger_transaction_hashes"} 0
rn_raw_db_size{category="local_transaction_executions"} 9957188
rn_raw_db_size{category="notarized_transaction_hashes"} 0
rn_raw_db_size{category="raw_ledger_transactions"} 0
rn_raw_db_size{category="receipt_accu_tree_slices"} 44714797
rn_raw_db_size{category="stale_state_hash_tree_parts"} 77519824
rn_raw_db_size{category="state_hash_tree_nodes"} 1061135269
rn_raw_db_size{category="substate_node_ancestry_records"} 0
rn_raw_db_size{category="substates"} 3604035
rn_raw_db_size{category="transaction_accu_tree_slices"} 44719326
rn_raw_db_size{category="transaction_receipts"} 6064682
rn_raw_db_size{category="vertex_store"} 12592
```
_(and similar `raw_db_entries{}`)_

The readings seem to reflect actual raw `du` pretty closely: _(also based on the above output)_

image

_(`du` says 1.2 GB, `rn_raw_db_size` sums up to 1.24 GB)_


jakrawcz-rdx:
# Re: cost of measuring this

The `get_data_volume_statistics()` call took at most 93.583 µs on my machine (typical: 60-80 µs), when I had ~40 `.sst` files. At the same time a "commit write batch" was taking ~450 µs for me.

On prod, the "commit write batch" takes ~2ms, so even by linear proportion I don't expect the measurement will take more than ~0.5ms there. In fact, it should depend only on the file count (not workload, not actual data volume). And the file count should grow significantly sub-linearly.

To sum up: I think we are currently fine with measuring after every commit, and if we see any problem, I see easy mitigations (measure once per N commits, or once per M minutes, or "spend on measurement max 0.01% of wallclock")

core-rust/state-manager/src/jni/node_rust_environment.rs
```diff
@@ -199,3 +219,19 @@ impl JNINodeRustEnvironment {
...
+    database: Arc<RwLock<StateManagerDatabase>>,
+    metric_registry: &Registry,
+) -> ScheduleHandle {
+    let raw_db_metrics = RawDbMetrics::new(metric_registry);
+    let mut scheduler = Scheduler::new();
+    scheduler.every(RAW_DB_MEASUREMENT_INTERVAL).run(move || {
+        let statistics = database.read().get_data_volume_statistics();
+        raw_db_metrics.update(statistics);
+    });
+    scheduler.watch_thread(Duration::from_secs(1))
```
dhedey, line 236: What does this do out of interest?

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -553,6 +554,35 @@ impl ConfigurableDatabase for RocksDBStore {
...
+                warn!("could not get DB live files; returning 0: {:?}", err);
+                Vec::new()
+            }
+        };
+        for live_file in live_files {
+            let Some(statistic) = statistics.get_mut(&live_file.column_family_name) else {
+                warn!("LiveFile of unknown column family: {:?}", live_file);
+                continue;
+            };
+            statistic.add(live_file.num_entries, live_file.size);
```
dhedey, line 580: As we're iterating across SSTs, I guess these entries could be tombstones as well as "create/update" entries? Which is fine, just wanted to understand what we're measuring.

It would be nice to update the description on `raw_db_entries` to include if it includes tombstones, updates, and recreations which haven't yet been compacted?

dhedey (approved):

core-rust/state-manager/src/jni/node_rust_environment.rs
jakrawcz-rdx: it literally does `thread::spawn()`, and returns a super-nice handle that will stop the scheduler when dropped.
I would say that conceptually it finalizes the build of the `{scheduler + its tasks}` (since it consumes `self`).

core-rust/state-manager/src/store/rocks_db.rs
jakrawcz-rdx: The rocks' doc says `Number of entries/alive keys in the file`.

And I believe it is the actual number of alive keys (i.e. excludes deletions), because there is another field available:
```
    /// Number of deletions/tomb key(s) in the file
    pub num_deletions: u64,
```
(which I choose to not monitor)

jakrawcz-rdx: OFC the total that I am `.add()`ing is still an estimate, because of upserts. Lemme add some doc about it.

dhedey: Oooo interesting :D

I could see it be cool to return all three:
* `num_tombstones`
* `num_upserts`
* `num_entries ~ num_upserts - num_tombstones`

But anyway, not needed for 1.0.4

core-rust/state-manager/src/jni/node_rust_environment.rs
dhedey: And is the `Duration` is a "graceful delay before I kill you on stop command" thing?

jakrawcz-rdx: no, it is a lame "how often the scheduler thread wakes up to check whether any task run is due" a.k.a. "precision"
(which should not be needed at all in a non-lame scheduler implementation)

PR #711, opened 2023-10-06
Remove clokwerk in favour of a Tokio thread

malexandru-rdx:
This builds on top of https://github.com/radixdlt/babylon-node/pull/707 and removes `clokwerk` lib in favour of a tokio thread.


jakrawcz-rdx (commented):
I am very glad that this is achievable tokio-native :)

The code looks nice (although complicated by low-level interactions with tokio and async). I am 100% for using this, but - as discussed - we need a nice facade/api (i.e. `let my_handle = scheduler.run(every 10 sec, || my_task());`, so that we can immediately use it in existing places (i.e. drop-in replacement for the lame `clockwerk`).

PR #721, opened 2023-10-11
Tokio scheduler infra

malexandru-rdx:
This supersedes https://github.com/radixdlt/babylon-node/pull/711: adds a simple piece of Tokio infra to simplify periodic tasks spawn (and tracking) and also removes `clokwerk`.


core-rust/node-common/src/scheduler.rs
```diff
@@ -74,7 +95,16 @@ pub trait Scheduler {
     /// The consecutive runs should be started at approximately the given intervals - however, if
     /// a particular run takes longer than the interval, then it should not be (in any way) aborted,
     /// and no concurrent run should be started - the schedule should simply be delayed.
-    fn start_periodic(&mut self, interval: Duration, task: impl 'static + FnMut() + Send);
+    fn start_periodic_advanced(
+        &self,
+        duration: Duration,
+        task: impl 'static + FnMut() + Send,
+        shutdown_signal: oneshot::Receiver<()>,
```
jakrawcz-rdx, line 102: I first need to ask a question out of my ignorance: in what situations would the caller need to call the `_advanced()` version? Do you see foreseeable use-cases in our code?

And unrelated to the above I have 2 improvement ideas:
1. can we use some less-specific type (trait maybe?) for the `shutdown_signal`? (in java it would be something like `Future<>` or `Supplier<>` or `Callable<>` I assume). Or maybe a custom well-named type? (I'd like to avoid relying on `oneshot::` in the API)
2. can you add a piece of doc explaining how to use the `shutdown_signal`?
3. can this API be "inverted" somehow? (as in: not require the caller to pass the `shutdown_signal` as a param, but instead we return a `ScheduledTask` which allows to control the execution - in particular, cancel it)

malexandru-rdx: > I first need to ask a question out of my ignorance: in what situations would the caller need to call the _advanced() version? Do you see foreseeable use-cases in our code?

Obviously no use case right now, but I can easily imagine wanting to use something with more finer control than what the alternative is: drop the `TokioScheduler` and kill **all** tasks. 

> 1. can we use some less-specific type (trait maybe?) for the shutdown_signal? (in java it would be something like Future<> or Supplier<> or Callable<> I assume). Or maybe a custom well-named type? (I'd like to avoid relying on oneshot:: in the API)

Why? I don't consider "more like Java, less like Rust" an valid argument. I understand that it seems ugly at first, but please try and use it. It **is** the right abstraction: (non-`clone`able, non-`copy`able) single-use signal enforced by the compiler - this makes it hard to make mistakes (check answer to Q3 too).
The established pattern for this in Rust really is the `oneshot::`, all libraries built on top of tokio that need a shutdown signal use it.
I guess that I can wrap it again in another thin layer that doesn't really do anything but just hides the `oneshot`, but I don't see the point :).

> 2. can you add a piece of doc explaining how to use the shutdown_signal?
Sure, I can try. But my question is if we should add to all the occurrences in the future? How about [CoreApiServer](https://github.com/radixdlt/babylon-node/blob/7676526258c6be7cc99c20c3894bc43c2fe84b67/core-rust/core-api-server/src/jni/mod.rs#L133)?

> 3. can this API be "inverted" somehow? (as in: not require the caller to pass the shutdown_signal as a param, but instead we return a ScheduledTask which allows to control the execution - i.e. cancel it [or whatever the shutdown_signal is for])

You can easily use the `_advanced` and wrap the result in a `ShutdownOnDrop` (which is what the simpler method does). Note that in current usage of the Scheduler you can't return just a task handler and keep it in `StateManager` because the `StateManager` is `Clone` and also used by `core-api` - dropping the `core-api` will also trigger the `drop` on the "cloned" handlers and wrongfully stop the tasks.

core-rust/node-common/src/scheduler.rs
```diff
@@ -62,57 +62,227 @@
...
+    type VoidFuture<'a>: Future<Output = ()>
+    where
+        Self: 'a;
+
+    /// Be aware that cancellation is not an atomic operation. It is possible
+    /// for another thread running in parallel with a call to `cancel` to first
+    /// receive `true` from `is_cancelled` on one child node, and then receive
+    /// `false` from `is_cancelled` on another child node. However, once the
+    /// call to `cancel` returns, all child nodes have been fully cancelled.
+    fn cancel(&self);
```
jakrawcz-rdx, line 89: Pretty please, let's leave just this single method in the public interface of the scheduler facade v1 🙏 .

PR #723, opened 2023-10-16
Impl for GC-ing non-critical ledger proofs from configured old-enough epochs

jakrawcz-rdx:
Addresses https://docs.google.com/spreadsheets/d/1JPiisoLFOhrDBT1-qDj2uvVwIsojKQd8kHzf1CIhaFw/edit#gid=0&range=B15

The implementation here is quite similar to the "JMT GC", with notable exceptions:
- This GC goes epoch-by-epoch (not version-by-version)
  - This is because we do not want to prune the current epoch, which is still "live" (i.e. asking which proofs are critical gives different answers over time, until the epoch ends)
- This GC had no pre-existing "progress-tracking" CF (i.e. we don't have "stale proofs" CF)
  - Instead, I had to introduce the new, dedicated "proofs GC progress" CF, to persist the progress between runs.
- This GC _cannot_ use "historical, non-locked" DB access - it fully locks for read and write instead!
  - Please see the comment in the code for details
  - In general, I now think that our "DB locking" was rather naive from the beginning, and even the current "current vs historical state lock" is too weak to model all our data access dependencies. This GC works _fine_, but we should discuss locking more soon.

### DIFF
--- a/core-rust-bridge/src/main/java/com/radixdlt/transaction/REv2TransactionAndProofStore.java
+++ b/core-rust-bridge/src/main/java/com/radixdlt/transaction/REv2TransactionAndProofStore.java
@@ -73,7 +73,6 @@ import com.radixdlt.monitoring.Metrics;
 import com.radixdlt.monitoring.Metrics.MethodId;
 import com.radixdlt.sbor.Natives;
 import com.radixdlt.statecomputer.commit.LedgerProof;
-import com.radixdlt.utils.UInt32;
 import com.radixdlt.utils.UInt64;
 import java.util.Optional;
 
@@ -101,15 +100,9 @@ public final class REv2TransactionAndProofStore {
             .build(new TypeToken<>() {});
   }
 
-  public Option<TxnsAndProof> getTxnsAndProof(
-      long startStateVersionInclusive,
-      int maxNumberOfTxnsIfMoreThanOneProof,
-      int maxPayloadSizeInBytes) {
+  public Option<TxnsAndProof> getTxnsAndProof(long startStateVersionInclusive) {
     return this.getTxnsAndProof.call(
-        new TxnsAndProofRequest(
-            UInt64.fromNonNegativeLong(startStateVersionInclusive),
-            UInt32.fromNonNegativeInt(maxNumberOfTxnsIfMoreThanOneProof),
-            UInt32.fromNonNegativeInt(maxPayloadSizeInBytes)));
+        new TxnsAndProofRequest(UInt64.fromNonNegativeLong(startStateVersionInclusive)));
   }
 
   public Optional<LedgerProof> getLastProof() {
--- a/core-rust/node-common/src/config/limits.rs
+++ b/core-rust/node-common/src/config/limits.rs
@@ -72,20 +72,6 @@ pub const DEFAULT_MAX_TOTAL_VERTEX_EXECUTION_COST_UNITS_CONSUMED: u32 =
 pub const DEFAULT_MAX_TOTAL_VERTEX_FINALIZATION_COST_UNITS_CONSUMED: u32 =
     FINALIZATION_COST_UNIT_LIMIT * 2;
 
-/// Maximum number of transactions to return in a single "get transactions with their proof"
-/// response, but only if at least one proof can be used. In other words:
-/// - if the next proof covers more than [`MAX_TXNS_FOR_RESPONSES_SPANNING_MORE_THAN_ONE_PROOF`]
-///   transactions, then all those transactions can still be returned (given they fit under the
-///   [`MAX_TXN_BYTES_FOR_A_SINGLE_RESPONSE`] limit);
-/// - if the next proof contains less than [`MAX_TXNS_FOR_RESPONSES_SPANNING_MORE_THAN_ONE_PROOF`]
-///   transactions, then its subsequent proof can only be used if its transactions also fit under
-///   this limit.
-pub const MAX_TXNS_FOR_RESPONSES_SPANNING_MORE_THAN_ONE_PROOF: u32 = 1000;
-
-/// Maximum transactions size (in terms of their total byte size) to return in a single "get
-/// transactions with their proof" response.
-pub const MAX_TXN_BYTES_FOR_A_SINGLE_RESPONSE: u32 = 12 * 1024 * 1024;
-
 #[derive(Debug, Clone, Copy, ScryptoSbor)]
 pub struct VertexLimitsConfig {
     pub max_transaction_count: u32,
--- a/core-rust/node-common/src/config/limits.rs
+++ b/core-rust/node-common/src/config/limits.rs
@@ -72,6 +72,20 @@ pub const DEFAULT_MAX_TOTAL_VERTEX_EXECUTION_COST_UNITS_CONSUMED: u32 =
 pub const DEFAULT_MAX_TOTAL_VERTEX_FINALIZATION_COST_UNITS_CONSUMED: u32 =
     FINALIZATION_COST_UNIT_LIMIT * 2;
 
+/// Maximum number of transactions to return in a single "get transactions with their proof"
+/// response, but only if at least one proof can be used. In other words:
+/// - if the next proof covers more than [`MAX_TXNS_FOR_RESPONSES_SPANNING_MORE_THAN_ONE_PROOF`]
+///   transactions, then all those transactions can still be returned (given they fit under the
+///   [`MAX_TXN_BYTES_FOR_A_SINGLE_RESPONSE`] limit);
+/// - if the next proof contains less than [`MAX_TXNS_FOR_RESPONSES_SPANNING_MORE_THAN_ONE_PROOF`]
+///   transactions, then its subsequent proof can only be used if its transactions also fit under
+///   this limit.
+pub const MAX_TXNS_FOR_RESPONSES_SPANNING_MORE_THAN_ONE_PROOF: u32 = 1000;
+
+/// Maximum transactions size (in terms of their total byte size) to return in a single "get
+/// transactions with their proof" response.
+pub const MAX_TXN_BYTES_FOR_A_SINGLE_RESPONSE: u32 = 12 * 1024 * 1024;
+
 #[derive(Debug, Clone, Copy, ScryptoSbor)]
 pub struct VertexLimitsConfig {
     pub max_transaction_count: u32,
--- a/core-rust/state-manager/src/store/proofs_gc.rs
+++ b/core-rust/state-manager/src/store/proofs_gc.rs
@@ -87,6 +87,9 @@ pub struct LedgerProofsGcConfig {
     /// How often to run the GC, in seconds.
     /// Since this GC operates with an epoch precision, we do not need to run more often than epoch
     /// changes.
+    // TODO(after having some event-driven Rust infra): The entire `LedgerProofsGc` could be
+    // migrated away from `Scheduler` into `EventListener<EpochChangeCommittedEvent>` (as noted
+    // above - it wants to run async exactly once after each epoch).
     pub interval_sec: u32,
     /// How many most recent *completed* epochs should be left not GC-ed.
     /// Please note that the current epoch is never GC-ed.
@@ -124,8 +127,8 @@ impl LedgerProofsGc {
         self.interval
     }
 
-    /// Performs a single GC run, which is supposed to permanently delete *all* old-enough state
-    /// hash tree nodes marked as stale.
+    /// Performs a single GC run, which is supposed to permanently delete *all* non-critical ledger
+    /// proofs of configured old-enough epochs.
     pub fn run(&self) {
         // TODO(locks/snapshots): The GC's operation does not interact with the "current state", and
         // intuitively could use the "historical, non-locked" DB access. However, we have a (very
--- a/core-rust/state-manager/src/store/proofs_gc.rs
+++ b/core-rust/state-manager/src/store/proofs_gc.rs
@@ -64,9 +64,6 @@
 
 use radix_engine::types::{Categorize, Decode, Encode};
 
-use node_common::config::limits::{
-    MAX_TXNS_FOR_RESPONSES_SPANNING_MORE_THAN_ONE_PROOF, MAX_TXN_BYTES_FOR_A_SINGLE_RESPONSE,
-};
 use radix_engine_common::types::Epoch;
 use std::sync::Arc;
 use std::time::Duration;
@@ -79,6 +76,7 @@ use crate::store::traits::proofs::QueryableProofStore;
 
 use crate::store::StateManagerDatabase;
 
+use crate::jni::LedgerSyncLimitsConfig;
 use node_common::locks::StateLock;
 
 /// A configuration for [`LedgerProofsGc`].
@@ -104,21 +102,24 @@ pub struct LedgerProofsGc {
     database: Arc<StateLock<StateManagerDatabase>>,
     interval: Duration,
     most_recent_full_resolution_epoch_count: u64,
+    limits_config: LedgerSyncLimitsConfig,
 }
 
 impl LedgerProofsGc {
     /// Creates a new GC.
     pub fn new(
         database: Arc<StateLock<StateManagerDatabase>>,
-        config: LedgerProofsGcConfig,
+        gc_config: LedgerProofsGcConfig,
+        limits_config: LedgerSyncLimitsConfig,
     ) -> Self {
         Self {
             database,
-            interval: Duration::from_secs(u64::from(config.interval_sec)),
+            interval: Duration::from_secs(u64::from(gc_config.interval_sec)),
             most_recent_full_resolution_epoch_count: u64::try_from(
-                config.most_recent_full_resolution_epoch_count,
+                gc_config.most_recent_full_resolution_epoch_count,
             )
             .unwrap(),
+            limits_config,
         }
     }
 
@@ -143,9 +144,9 @@ impl LedgerProofsGc {
         // Read the GC's persisted state and initialize the run:
         let read_progress_database = self.database.read_current();
         let to_epoch = read_progress_database
-            .max_epoch()
-            .map(|current_epoch| current_epoch.number())
-            .and_then(|current| current.checked_sub(self.most_recent_full_resolution_epoch_count))
+            .max_completed_epoch()
+            .map(|max_completed_epoch| max_completed_epoch.number())
+            .and_then(|number| number.checked_sub(self.most_recent_full_resolution_epoch_count))
             .map(Epoch::of);
         let Some(to_epoch) = to_epoch else {
             // Nothing to GC ever, yet.
@@ -170,7 +171,7 @@ impl LedgerProofsGc {
 
         let mut retained_proofs = 0; // only for logging purposes
         loop {
-            let at_state_version = last_pruned_state_version
+            let batch_start_state_version_inclusive = last_pruned_state_version
                 .next()
                 .expect("state version overflow");
 
@@ -178,16 +179,17 @@ impl LedgerProofsGc {
             // the business logic responsible for outputting proven transactions:
             let locate_proof_database = self.database.read_current();
             let transactions_and_proof = locate_proof_database.get_txns_and_proof(
-                at_state_version,
-                MAX_TXNS_FOR_RESPONSES_SPANNING_MORE_THAN_ONE_PROOF,
-                MAX_TXN_BYTES_FOR_A_SINGLE_RESPONSE,
+                batch_start_state_version_inclusive,
+                self.limits_config
+                    .max_txns_for_responses_spanning_more_than_one_proof,
+                self.limits_config.max_txn_bytes_for_single_response,
             );
             drop(locate_proof_database);
 
             let Some((_transactions, proof)) = transactions_and_proof else {
                 error!(
-                    "A chain of transactions-without-proof from state version {} does not fit within the limits; aborting the GC",
-                    at_state_version,
+                    "A chain of transactions-without-proof from state version {} does not fit within the limits {:?}; aborting the GC",
+                    batch_start_state_version_inclusive, self.limits_config
                 );
                 return;
             };
@@ -196,8 +198,11 @@ impl LedgerProofsGc {
             // Delete all the proofs from the beginning of transactions listing up to (but
             // excluding) the returned proof:
             last_pruned_state_version = header.state_version;
-            let delete_database = self.database.read_current();
-            delete_database.delete_ledger_proofs_range(at_state_version, last_pruned_state_version);
+            let delete_database = self.database.write_current();
+            delete_database.delete_ledger_proofs_range(
+                batch_start_state_version_inclusive,
+                last_pruned_state_version,
+            );
 
             retained_proofs += 1;
             if let Some(next_epoch) = header.next_epoch {
--- a/core-rust/state-manager/src/store/proofs_gc.rs
+++ b/core-rust/state-manager/src/store/proofs_gc.rs
@@ -152,9 +152,15 @@ impl LedgerProofsGc {
             // Nothing to GC ever, yet.
             return;
         };
-        let progress_started_at: LedgerProofsGcProgress = read_progress_database
-            .get_progress()
-            .unwrap_or_else(LedgerProofsGcProgress::none);
+        let progress_started_at: LedgerProofsGcProgress =
+            read_progress_database.get_progress().unwrap_or_else(|| {
+                LedgerProofsGcProgress::new(
+                    read_progress_database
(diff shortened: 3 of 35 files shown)


core/src/test/java/com/radixdlt/rev2/LedgerProofsGcTest.java
```diff
@@ -0,0 +1,188 @@
...
+                        false,
+                        StateHashTreeGcConfig.forTesting(),
+                        new LedgerProofsGcConfig(
+                            UInt32.fromNonNegativeInt(1),
+                            UInt64.fromNonNegativeLong(mostRecentFullResolutionEpochCount)),
+                        false))));
+  }
+
+  @Test
+  public void node_prunes_non_critical_proofs_from_configured_old_enough_epochs() {
```
jakrawcz-rdx, line 135: _(note from author)_

This test takes 14s locally.
It is long because it wants to see that 2 proofs per epoch are left not-pruned. (and to achieve it, each epoch needs to excess the "1000 txns or 12 MB limit")

I am asking since I am not sure about our conventions on test timings.
Is 14s ok? Should I somehow mark it "slow"? Should I somehow make it an "integration" test? Or should I, instead, make the test run ~1s at the price of only testing super-small, single-proof epochs?

malexandru-rdx: Maybe we can test it with a lower `MAX_TXNS_FOR_RESPONSES_SPANNING_MORE_THAN_ONE_PROOF` & `MAX_TXN_BYTES_FOR_A_SINGLE_RESPONSE`?

malexandru-rdx (approved):

core-rust/state-manager/src/store/proofs_gc.rs
```diff
@@ -0,0 +1,223 @@
...
+                );
+                return;
+            };
+            let header = proof.ledger_header;
+
+            // Delete all the proofs from the beginning of transactions listing up to (but
+            // excluding) the returned proof:
+            last_pruned_state_version = header.state_version;
+            let delete_database = self.database.read_current();
+            delete_database.delete_ledger_proofs_range(at_state_version, last_pruned_state_version);
```
dhedey, line 200: Potential issue (!!) (EDIT: Not actually an issue I think, but still a point of confusion).

The Rust doc says "from inclusive, to exclusive" - but if we delete the proof `at_state_version` don't we delete the previous end-proof and break our invariant? 

Maybe I'm misunderstanding? EDIT: Okay, I see my mistake, I has assumed `at_state_version` was `last_pruned_state_version` but actually we do:
```rust
let at_state_version = last_pruned_state_version
                .next()
```

However I believe this naming is confusing - for me, a ledger state sync simulating "at state version X" would have its ledger tip at `X` and should be requesting with a `start_state_version_inclusive` of X + 1.

Could I suggest therefore renaming the `at_state_version` variable to `batch_start_state_version_inclusive` or similar? This then aligns with the naming of the corresponding parameter to `get_txns_and_proof` and `delete_ledger_proofs_range` and makes every line-by-line read correctly to me.

core-rust/state-manager/src/store/proofs_gc.rs
```diff
@@ -0,0 +1,223 @@
...
+        );
+        let mut last_pruned_state_version = progress_started_at.epoch_proof_state_version;
+
+        let mut retained_proofs = 0; // only for logging purposes
+        loop {
+            let at_state_version = last_pruned_state_version
+                .next()
+                .expect("state version overflow");
+
+            // Locate the next proof that we need to retain. We use the same method and limits as
```
dhedey, line 177: I like the re-use of this logic

core-rust/state-manager/src/store/proofs_gc.rs
```diff
@@ -0,0 +1,223 @@
...
+    pub fn interval(&self) -> Duration {
+        self.interval
+    }
+
+    /// Performs a single GC run, which is supposed to permanently delete *all* non-critical ledger
+    /// proofs of configured old-enough epochs.
+    pub fn run(&self) {
+        // TODO(locks/snapshots): The GC's operation does not interact with the "current state", and
+        // intuitively could use the "historical, non-locked" DB access. However, we have a (very
+        // related) use-case, which lists arbitrary past transactions and their proof - and it could
```
dhedey, line 135: Can we be more explicit? You're referring to ledger sync here I believe?

---

Also I'm wondering about if we could just have a `ledger_sync` lock, which is taken out as readable to serve ledger sync responses, and is taken out as writable by this process? (I think ledger extension doesn't need to pick up this lock as it's atomic - although we'd want to check that the `get_txns_and_proof` doesn't assume any such invariants).

This also decreases contention whereby ledger sync requests could slow down commit/consensus 👍

core-rust/state-manager/src/store/proofs_gc.rs
dhedey: Regardless, I'd like to see a test which can commit stuff, and then can run the GC then still serve ledger syncs.

_Or_ if it's easier, can commit a few epochs, then wipe the GC tracker, start again from scratch, and not hit the `"A chain of transactions-without-proof from state version {} does not fit within the limits; aborting the GC"` error.

Possibly updating the Java test might be easier.

core/src/test/java/com/radixdlt/api/core/TransactionStreamTest.java
```diff
@@ -331,15 +333,22 @@ public void test_previous_state_identifiers_and_proofs() throws Exception {
...
-      var firstPartResponseWithProofs =
-          getStreamApi()
-              .streamTransactionsPost(
-                  new StreamTransactionsRequest()
-                      .network(networkLogicalName)
-                      .includeProofs(true)
-                      .limit(100)
-                      .fromStateVersion(1L));
-      assertThat(firstPartResponseWithProofs.getProofs().size()).isEqualTo(13);
+      // An async "proofs pruning" is supposed to run over the genesis' epoch...
```
dhedey, line 336: How is this guaranteed? Is this because the test GC config runs quickly? This feels a little too implicit for my liking

core/src/test/java/com/radixdlt/api/core/TransactionStreamTest.java
```diff
@@ -331,15 +333,22 @@ public void test_previous_state_identifiers_and_proofs() throws Exception {
...
+              () -> {
+                var firstPartResponseWithProofs =
+                    getStreamApi()
+                        .streamTransactionsPost(
+                            new StreamTransactionsRequest()
+                                .network(networkLogicalName)
+                                .includeProofs(true)
+                                .limit(100)
+                                .fromStateVersion(1L));
+                // ... leaving 9 (out of original 13) ledger proofs.
```
dhedey, line 349: Why are 9 proofs kept instead of all of them being deleted? Are epochs short and these are the end of epoch proofs?

In any case, this assertion feels too specific, I'd rather we just assert some proofs are returned, or that the max gap between proofs is 1000 or something :)

core/src/test/java/com/radixdlt/rev2/LedgerProofsGcTest.java
dhedey: Hmm well yes and no, we have a whole section of our configuration for `protocol.*` variables, and sanity checks they're internally consistent. These shouldn't be changed, but are useful to change for (A) investigating what the correct values should be / optimizing and (B) tests.

I think it's reasonable to make these configurable, but have a default which shouldn't be changed in prod. We could even consider requiring `protocol.enable_unsafe_testing_only_protocol_constant_changes` or something to read them :D

dhedey: That said, 14 seconds isn't so bad 👍

core/src/test/java/com/radixdlt/rev2/LedgerProofsGcTest.java
```diff
@@ -0,0 +1,200 @@
...
+          .untilAsserted(
+              () -> {
+                var stateReader = test.getInstance(0, TestStateReader.class);
+                // - The epoch 5 has just started, hence it has no proofs
+                assertThat(stateReader.countProofsWithinEpoch(5)).isEqualTo(0);
+                // - The 2 most recent completed epochs contain all their proofs (i.e. not pruned)
+                assertThat(stateReader.countProofsWithinEpoch(4)).isEqualTo(epochLength);
+                assertThat(stateReader.countProofsWithinEpoch(3)).isEqualTo(epochLength);
+                // - This epoch was pruned, and it has 2 proofs (1 would not fit the 12MB limit)
+                assertThat(stateReader.countProofsWithinEpoch(2)).isEqualTo(2);
```
dhedey, line 160: Hmm okay maybe my bug spot was wrong in the gc code, but I'll leave it to you to work out what I wrongly observed.

core/src/test/java/com/radixdlt/rev2/LedgerProofsGcTest.java
```diff
@@ -0,0 +1,200 @@
...
+          .untilAsserted(
+              () -> {
+                // ... the "pruning window" has progressed (i.e. now epoch 3 got pruned)
+                var stateReader = test.getInstance(0, TestStateReader.class);
+                assertThat(stateReader.countProofsWithinEpoch(6)).isEqualTo(0);
+                assertThat(stateReader.countProofsWithinEpoch(5)).isEqualTo(epochLength);
+                assertThat(stateReader.countProofsWithinEpoch(4)).isEqualTo(epochLength);
+                assertThat(stateReader.countProofsWithinEpoch(3)).isEqualTo(2);
+                assertThat(stateReader.countProofsWithinEpoch(2)).isEqualTo(2);
+              });
```
dhedey, line 178: Might be nice to include a full node, but only turn it on now, and ensure that it can sync up to epoch 6 :D

core-rust/state-manager/src/store/proofs_gc.rs
```diff
@@ -0,0 +1,223 @@
...
+                    at_state_version,
+                );
+                return;
+            };
+            let header = proof.ledger_header;
+
+            // Delete all the proofs from the beginning of transactions listing up to (but
+            // excluding) the returned proof:
+            last_pruned_state_version = header.state_version;
+            let delete_database = self.database.read_current();
```
dhedey, line 199: !! Shouldn't this be `write_current()` so it contends with the ledger sync, as per your comment higher up?

core/src/test/java/com/radixdlt/rev2/LedgerProofsGcTest.java
```diff
@@ -0,0 +1,200 @@
...
+          .atMost(2 * GC_INTERVAL_SEC, TimeUnit.SECONDS)
+          .untilAsserted(
+              () -> {
+                var stateReader = test.getInstance(0, TestStateReader.class);
+                // - The epoch 5 has just started, hence it has no proofs
+                assertThat(stateReader.countProofsWithinEpoch(5)).isEqualTo(0);
+                // - The 2 most recent completed epochs contain all their proofs (i.e. not pruned)
+                assertThat(stateReader.countProofsWithinEpoch(4)).isEqualTo(epochLength);
+                assertThat(stateReader.countProofsWithinEpoch(3)).isEqualTo(epochLength);
+                // - This epoch was pruned, and it has 2 proofs (1 would not fit the 12MB limit)
```
dhedey, line 159: Ideally I'd like us to be able to test the 1000 transaction invariant too, if that'd be possible? Although obviously we should use much smaller transactions for this :)

dhedey (changes requested):
Looks mostly really good, can't see any critical issues, but I have a few picky comments / additional test case requests, so marking a request changes.

This is a potentially risky change and I'd like to check that the sync invariant is maintained in a more combinations of scenarios.

core-rust/node-common/src/config/limits.rs
```diff
@@ -72,6 +72,20 @@ pub const DEFAULT_MAX_TOTAL_VERTEX_EXECUTION_COST_UNITS_CONSUMED: u32 =
 pub const DEFAULT_MAX_TOTAL_VERTEX_FINALIZATION_COST_UNITS_CONSUMED: u32 =
     FINALIZATION_COST_UNIT_LIMIT * 2;
 
+/// Maximum number of transactions to return in a single "get transactions with their proof"
+/// response, but only if at least one proof can be used. In other words:
+/// - if the next proof covers more than [`MAX_TXNS_FOR_RESPONSES_SPANNING_MORE_THAN_ONE_PROOF`]
```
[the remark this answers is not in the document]
jakrawcz-rdx: > complexly defined

Indeed:

https://github.com/radixdlt/babylon-node/blame/d355c946ecae3a87df8fa9f1d6a27c09a043a962/core-rust/state-manager/src/store/rocks_db.rs#L982-L987

_(looks to me like "we have loosened the requirement" for the first proof from the requested range)_

> this deserves a wider conversation (with eg Lukasz and Josh)
> why do we still have this limit for transactions spanning more than one proof?

So at the moment @LukasGasior1 probably knows the most about it (see link above).

I strongly committed to **not** touching the ledger-sync rules or the `get_txns_and_proof()` logic while developing this PR. I still believe the most bulletproof way to implement the GC is to rely on the very same logic. If we want to change that logic (or its consts), we should do it in a separate PR. If we are changing it, then it's probably worth doing before the deployment of this GC (i.e. we have to pause this PR and discuss).

Let me start by chatting to @LukasGasior1 today.

core/src/main/java/com/radixdlt/messaging/MessagingModule.java
```diff
@@ -142,7 +142,7 @@ private int maxMessageSize(
     final var maxMessageSizeBaseline =
         Math.max(
             Math.max(
-                REv2TransactionsAndProofReader.MAX_TXN_BYTES_FOR_A_SINGLE_RESPONSE,
+                NodeConstants.MAX_TXN_BYTES_FOR_A_SINGLE_RESPONSE,
```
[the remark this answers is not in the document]
jakrawcz-rdx: That's true... as a consequence, the `proposalMaxUncommittedTransactionsPayloadSize` is not needed in this method at all.

Lemme add...
```
if (maxUncommittedTransactionsPayloadSize > maxTxnBytesForSingleResponse) {
  log.warn("you want it too high, we gotta use maxTxnBytesForSingleResponse instead");
  return maxTxnBytesForSingleResponse;
}
```
... to the `@ProposalMaxUncommittedTransactionsPayloadSize` provider.
_(since this seems to be our convention)_

core/src/main/java/com/radixdlt/rev2/REv2TransactionsAndProofReader.java
```diff
@@ -101,21 +90,12 @@ public REv2TransactionsAndProofReader(REv2TransactionAndProofStore transactionSt
...
     if (rawTxnsAndProofOpt.isEmpty()) {
       log.error(
-          "Impossible to build a chain of transactions from state version "
-              + startStateVersionInclusive
-              + " that ends with a proof that is: under "
-              + MAX_TXNS_FOR_RESPONSES_SPANNING_MORE_THAN_ONE_PROOF
-              + " transactions and under "
-              + MAX_TXN_BYTES_FOR_A_SINGLE_RESPONSE
-              + " total payload size!");
+          "Impossible to build a chain of txns from state version {} fitting within the limits",
```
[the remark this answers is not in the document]
jakrawcz-rdx: yeah the problem was only specific to Core API layer, `handle_stream_transactions()`, and only because it wanted to return `previous_state_identifiers` (see https://rdxworks.slack.com/archives/C01M90Y2448/p1695389384296069) and it got fixed in https://github.com/radixdlt/babylon-node/pull/688 and now is covered by an int-test `requesting_state_version_out_of_bounds_returns_error()`.

The `REv2TransactionsAndProofReader`'s method (and its rust impl `get_txns_and_proof()`) was always resilient to "get far future" (an empty proofs iter results in `return None;`) and I haven't touched it (only adjusted the error message).

core-rust/state-manager/src/store/traits.rs
```diff
@@ -310,6 +310,10 @@ pub mod proofs {
     #[enum_dispatch]
     pub trait QueryableProofStore {
         fn max_state_version(&self) -> StateVersion;
+        fn max_epoch(&self) -> Option<Epoch> {
+            self.get_last_epoch_proof()
+                .map(|proof| proof.ledger_header.epoch)
```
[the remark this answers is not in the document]
jakrawcz-rdx: renaming `max_completed_epoch` epoch 👍 _(it cannot be clearer, even if an unused `max_uncompleted_epoch` was right next to it)_

core/src/main/java/com/radixdlt/RadixNodeModule.java
```diff
@@ -437,6 +438,36 @@ private StateHashTreeGcConfig parseStateHashTreeGcConfig(RuntimeProperties prope
...
+    // changes.
+    var intervalSec = properties.get("ledger_proofs.gc.interval_sec", 300);
+    Preconditions.checkArgument(
+        intervalSec > 0, "ledger proofs GC interval must be positive: %s sec", intervalSec);
+
+    // How many most recent completed epochs should not be pruned?
+    // Set to 0 to discard non-critical proofs of every completed epoch as soon as possible.
+    // Set to Integer.MAX_VALUE to effectively disable this GC.
+    var mostRecentFullResolutionEpochCount =
+        properties.get("ledger_proofs.most_recent_full_resolution_epoch_count", 1000);
```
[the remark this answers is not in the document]
jakrawcz-rdx: changing to 12 👍

core-rust/state-manager/src/store/proofs_gc.rs
jakrawcz-rdx: > if we could just have a `ledger_sync` lock

Yes, that's a good direction, but I don't want to go there in this PR.

`<off-topic>`:
I think that the current shape of `StateLock` (our DB lock) clearly shows that we don't want to lock a single "DB the resource", but rather a set of abstractly defined "accesses to DB". E.g.:

- Lock 1: 
  - Access: "Consistent current top-of-ledger"
  - Typical reader: "get current state version and values of a couple substates" (i.e. state-reading Core API endpoints)
  - Typical writer: Ledger Commit

- Lock 2:
  - Access: "A sequence of transactions and their end-proof"
  - Typical reader: Ledger Sync responder
  - Typical writer: Ledger Proofs GC

_(possibly more)_

And the `&database` itself is simply always accessible without any lock. This means that readers/writers would have to be a lot more careful... (remember about locking R or W of the lock appropriate to your Access + be careful to never interact with the parts of DB guarded by locks of other Accesses).

TBH I'd refactor it right away, but I acknowledge the same area is touched by @malexandru-rdx's "db snapshots" work, so let's first see that.
`</off-topic>`

So, in this PR, let me (fix and) leave the full R/W-locking of entire DB for GC purposes.
(GC process organically releases & re-acquires the lock, so even during the backlog it should not starve the BFT process)

core/src/test/java/com/radixdlt/api/core/TransactionStreamTest.java
jakrawcz-rdx: right :( there is awaitility (2x the GC period used in tests), but it still quietly assumes that GC completes quickly enough.

Together with the next comment to the same test, I now think it may be better to NOT "await until GC", just assert that `proofs.size() > 0`.

core/src/test/java/com/radixdlt/api/core/TransactionStreamTest.java
jakrawcz-rdx: > Why are 9 proofs kept instead of all of them being deleted?

According to logs in this tests, we are at epoch 2, round 8. There are 8 proofs in epoch 2 (the GC never run over epoch 2, because it is not a concluded epoch), and a single proof from epoch 1 (i.e. only 1 is retained from original 5).

In general, this test is not about ledger proofs GC (and not even about ledger sync limits), so the best way to deal with confusion here is to relax the asserts (just "some proofs are returned") so that we don't have to consider the async pruning.

core/src/test/java/com/radixdlt/rev2/LedgerProofsGcTest.java
jakrawcz-rdx: hard, but possible 👍

core/src/test/java/com/radixdlt/rev2/LedgerProofsGcTest.java
jakrawcz-rdx: that's a good separate test 👍

jakrawcz-rdx:
Alright, dear reviewers, I addressed all the comments.

Besides the renames and the embarrassing write-lock fix, the most important changes are:
- I moved the ledger-sync limits (consts) back to java - see `LedgerSyncLimitsConfig`:
  - they are configurable on the Guice level (i.e. in tests, but not from external properties)
  - they are passed to every `getTxnsAndProof()` (as before) and now also to the ledger proofs GC config
- I added more tests (in the same `LedgerProofsGcTest`):
  - including the high-level sanity test about a full-node which ledger-syncs from a validator after GC

PTAL

core-rust-bridge/src/main/java/com/radixdlt/transaction/REv2TransactionAndProofStore.java
```diff
@@ -100,9 +100,11 @@ public REv2TransactionAndProofStore(Metrics metrics, NodeRustEnvironment nodeRus
             .build(new TypeToken<>() {});
   }
 
-  public Option<TxnsAndProof> getTxnsAndProof(long startStateVersionInclusive) {
+  public Option<TxnsAndProof> getTxnsAndProof(
```
dhedey, line 103: This is fine, although minorly surprising - I'd kinda expect this config to live in some kind of rust service responsible for ledger sync (and set from some kind of constructor).

I don't think we should change that in this PR necessarily, but just a thought for the future 👍

core/src/test/java/com/radixdlt/rev2/LedgerProofsGcTest.java
```diff
@@ -114,36 +129,42 @@ private DeterministicTest createTest(long mostRecentFullResolutionEpochCount, lo
...
-    /// ... 6 rounds * 3 transactions * 1 MB = 18MB; Limit is 12 MB - so we require 2 proofs
-    var epochLength = 6;
-    try (var test = createTest(mostRecentFullResolutionEpochCount, epochLength)) {
-      test.startAllNodes();
+  public void node_retains_enough_proofs_to_cover_max_transaction_size_in_old_enough_epochs() {
+    /// 6 rounds * 100 KB = 600KB
+    var roundsPerEpoch = 6;
+    var txnSize = 100 * 1024;
+    /// Limit is 400 KB - so we require 2 proofs to cover 600 KB
+    var maxTxnPayloadSizeUnderProof = 400 * 1024; // cannot be much lower due to genesis' size
```
dhedey, line 162: Thanks for the comment!

This is a really good point, and raises a possible issue. Big gaps in genesis are irrelevant for ledger sync, because genesis never needs to be synced... BUT it is a problem for the pruner unless we special case genesis. In particular, on mainnet, we have no proofs across all of genesis.

I think we may have to special case genesis in the pruner (basically ignore it / don't try to clear up proofs). This might also apply to protocol update transactions, but we can ignore that until discussions shortly.

dhedey (changes requested):
Really nice tests / tweaks, thanks! However just marking as "request chagnes" to check if we handle genesis correctly.

core-rust-bridge/src/main/java/com/radixdlt/transaction/REv2TransactionAndProofStore.java
jakrawcz-rdx: > to live in some kind of rust service responsible for ledger sync

Yeah, that's the problem - there is no Rust service responsible for ledger sync 🤷 (because Java is).

This PR merely wraps the 2 ints into a `LedgerSyncLimitsConfig`, but I agree that some coming-soon refactor could introduce a... "Rust service responsible for _picking transactions to be returned in_ ledger sync" - if it's not too narrow of a responsibility :(
A much better move will be possible only after migrating to 100% Rust.

core/src/test/java/com/radixdlt/rev2/LedgerProofsGcTest.java
jakrawcz-rdx: Huh, that's all true observation, but addressing it is larger than expected.
Let's track it as [another work item](https://docs.google.com/spreadsheets/d/1JPiisoLFOhrDBT1-qDj2uvVwIsojKQd8kHzf1CIhaFw/edit#gid=0&range=B19), and another PR.

dhedey (approved):
Really nice fix, thanks. Happy to merge

PR #722, opened 2023-10-12
Making our lock facade more configurable + Adding metrics capability to it

jakrawcz-rdx:
Every lock now gets these 4:
```
# HELP rn_locks_acquire_seconds Time spent waiting to acquire a specific lock.
# TYPE rn_locks_acquire_seconds histogram
rn_locks_acquire_seconds_bucket{lock="rn.state_computer.execution_cache",le="0.0001"} 10725
// other histo buckets
rn_locks_acquire_seconds_bucket{lock="rn.state_computer.execution_cache",le="+Inf"} 10725
rn_locks_acquire_seconds_sum{lock="rn.state_computer.execution_cache"} 0.0003370880000000202
rn_locks_acquire_seconds_count{lock="rn.state_computer.execution_cache"} 10725

# HELP rn_locks_hold_seconds Time spent holding a specific lock.
# TYPE rn_locks_hold_seconds histogram
rn_locks_hold_seconds_bucket{lock="rn.state_computer.execution_cache",le="0.0001"} 8332
// other histo buckets
rn_locks_hold_seconds_bucket{lock="rn.state_computer.execution_cache",le="+Inf"} 10725
rn_locks_hold_seconds_sum{lock="rn.state_computer.execution_cache"} 3.319986907999996
rn_locks_hold_seconds_count{lock="rn.state_computer.execution_cache"} 10725

# HELP rn_locks_holding_threads A number of threads currently holding a specific lock (may actually be >1 e.g. for a reader lock).
# TYPE rn_locks_holding_threads gauge
rn_locks_holding_threads{lock="rn.state_computer.execution_cache"} 0

# HELP rn_locks_waiting_threads A number of threads currently waiting to acquire a specific lock.
# TYPE rn_locks_waiting_threads gauge
rn_locks_waiting_threads{lock="rn.state_computer.execution_cache"} 0
```

PR #725, opened 2023-10-17
Making Scheduler facade configurable + Adding metrics capability to it

jakrawcz-rdx:
This is a sibling of https://github.com/radixdlt/babylon-node/pull/722 but for our `Scheduler` facade.

Unsurprisingly, the configuration API shape follows the approach we took in `LockFactory` facade. One notable difference comes from the “task tracker” (i.e. having to keep handles to the tasks in order to stop them at the right time).

The new metrics preview:
```
# HELP rn_scheduler_task_execute_seconds Time spent executing a single run of a specific task.
# TYPE rn_scheduler_task_execute_seconds histogram
rn_scheduler_task_execute_seconds_bucket{task="rn.raw_db_measurement",le="0.001"} 25
rn_scheduler_task_execute_seconds_bucket{task="rn.raw_db_measurement",le="0.005"} 26
rn_scheduler_task_execute_seconds_bucket{task="rn.raw_db_measurement",le="0.02"} 26
rn_scheduler_task_execute_seconds_bucket{task="rn.raw_db_measurement",le="0.1"} 26
rn_scheduler_task_execute_seconds_bucket{task="rn.raw_db_measurement",le="0.5"} 26
rn_scheduler_task_execute_seconds_bucket{task="rn.raw_db_measurement",le="2"} 26
rn_scheduler_task_execute_seconds_bucket{task="rn.raw_db_measurement",le="5"} 26
rn_scheduler_task_execute_seconds_bucket{task="rn.raw_db_measurement",le="+Inf"} 26
rn_scheduler_task_execute_seconds_sum{task="rn.raw_db_measurement"} 0.005004584
rn_scheduler_task_execute_seconds_count{task="rn.raw_db_measurement"} 26
rn_scheduler_task_execute_seconds_bucket{task="rn.state_hash_tree_gc",le="0.001"} 0
rn_scheduler_task_execute_seconds_bucket{task="rn.state_hash_tree_gc",le="0.005"} 0
rn_scheduler_task_execute_seconds_bucket{task="rn.state_hash_tree_gc",le="0.02"} 0
rn_scheduler_task_execute_seconds_bucket{task="rn.state_hash_tree_gc",le="0.1"} 0
rn_scheduler_task_execute_seconds_bucket{task="rn.state_hash_tree_gc",le="0.5"} 5
rn_scheduler_task_execute_seconds_bucket{task="rn.state_hash_tree_gc",le="2"} 5
rn_scheduler_task_execute_seconds_bucket{task="rn.state_hash_tree_gc",le="5"} 5
rn_scheduler_task_execute_seconds_bucket{task="rn.state_hash_tree_gc",le="+Inf"} 5
rn_scheduler_task_execute_seconds_sum{task="rn.state_hash_tree_gc"} 1.633424375
rn_scheduler_task_execute_seconds_count{task="rn.state_hash_tree_gc"} 5
```

### DIFF
--- a/core-rust/node-common/src/scheduler.rs
+++ b/core-rust/node-common/src/scheduler.rs
@@ -62,7 +62,10 @@
  * permissions under this License.
  */
 
+use std::cell::RefCell;
 use std::fmt::Display;
+
+use std::rc::Rc;
 use std::sync::Arc;
 
 use prometheus::{HistogramVec, Registry};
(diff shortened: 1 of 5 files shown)


core-rust/node-common/src/scheduler.rs
```diff
@@ -62,190 +62,216 @@
...
+        derived.running_task_tracker = Some(running_task_tracker);
+        derived
     }
-}
 
-impl<CT: CancellableToken> Drop for CancelOnDrop<CT> {
-    fn drop(&mut self) {
-        self.cancellation_token.cancel();
+    /// Configures the metrics collection for tasks *created from this point on*.
+    pub fn measured(&self, metrics_registry: &Registry) -> Self {
```
jakrawcz-rdx, line 135: _(note from author)_

one extra capability I imagine would be `.resume_after_panic()`

(I tested, and currently, a panic inside a task run is logged [by the default panic handler] and the task dies; maybe we would prefer that it resumes execution on the next interval? wdyt?)

core-rust/node-common/src/scheduler.rs
```diff
@@ -62,190 +62,216 @@
...
-        cancellation_token
     }
 }
 
-/// An auxiliary structure to be used along side a [`Scheduler`] in order to collect
-/// [`CancellableToken`]s for the spawned tasks in order to autostop them on drop.
-pub struct TaskTracker<CT: CancellableToken> {
-    tasks: Vec<CancelOnDrop<CT>>,
+/// A handle representing a running task started using the scheduler.
+/// Dropping *all* its clones cancels any subsequent executions of the task.
```
jakrawcz-rdx, line 196: _(note from author)_

I was considering whether we need an explicit `.cancel()` too (apart from "cancel on drop"). I didn't add it yet, to keep things simple (i.e. adding it in future is no big deal).

core-rust/node-common/src/scheduler.rs
```diff
@@ -62,190 +62,216 @@
...
-pub struct NoopCancellationFuture;
-
-impl Future for NoopCancellationFuture {
-    type Output = ();
-
-    fn poll(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<()> {
-        Poll::Ready(())
+    /// Creates a mock instance for testing purposes.
+    /// The "scheduled" tasks will not actually be started.
+    pub fn for_testing() -> Self {
```
jakrawcz-rdx, line 108: _(note from author)_

Obviously I liked the explicit `NoopScheduler` 10x more (than `runtime: None` 🤦 ).
But I was not able to make `trait Scheduler` work together with the "immutable builder pattern" - basically I had the same problem with `LockFactory`: the returned types became too wild and nested (or required `Box<dyn ...>` everywhere).

malexandru-rdx (changes requested):
This is a step backwards. Removing `trait`s in favour of a one-size-fits-all `struct` is not the way - I want us to leverage Rust static typing.

It is still possible to achieve the level of ease-of-use builder pattern without compromising this - of course with more code. And even if it is not possible 100%, then it is 99% a bad pattern to be forced into Rust.

core-rust/node-common/src/scheduler.rs
```diff
@@ -62,190 +62,216 @@
...
+        }
     }
 }
 
-impl<CT: CancellableToken> Default for TaskTracker<CT> {
-    fn default() -> Self {
-        Self::new()
+impl Drop for RunningTask {
+    fn drop(&mut self) {
+        if let Some(cancellation_token) = Arc::get_mut(&mut self.cancellation_token) {
```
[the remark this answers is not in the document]
jakrawcz-rdx: no, the `Arc::get_mut` only returns `Some` when it is _the only_ `Arc<>`. Since we are inside `drop()`, we additionally know that we have _the last_ `Arc<>`. So we cancel it and then the `Arc<>` is dropped, and its content (`CancellationToken`) gets dropped too and no-one cancels it anymore.

(AFAIU, the tokio async future holds a _clone_ of the `CancellationToken` contained here, but that's expected, because it wants to notice the cancelation that happens here, right?)

For context - my goal here is:
I want to hand out a couple of `RunningTask` handles (namely: one for the `TaskTracker` [if configured], one for the `.start_periodic()` caller). And the caller can also clone it if they have a reason.
And then, I want to cancel the task _not_ when the first of them is dropped, but after _all_ of them are dropped. That's it.

jakrawcz-rdx:
> It is still possible to achieve the level of ease-of-use builder pattern without compromising this - of course with more code.

I'll try. I plan to use this PR as an opportunity to learn something more regarding "OO design in rust".
My main idea for now is to actually separate out the "Scheduler core" (which would be a trait, responsible 100%-only for spawning a task) vs "Scheduler convenience configurator" (which is the "immutable builder", probably still a concrete struct). Wdyt?

core-rust/node-common/src/scheduler.rs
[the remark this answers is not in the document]
malexandru-rdx: Right! Thanks for explanation :D this is cool!

core-rust/state-manager/src/state_manager.rs
```diff
@@ -143,7 +143,7 @@ impl StateManager {
         mempool_relay_dispatcher: Option<MempoolRelayDispatcher>,
         lock_factory: &LockFactory,
         metrics_registry: &Registry,
-        scheduler: &impl Scheduler,
+        scheduler: &Scheduler<impl Spawner, impl Tracker, impl Metrics>,
```
jakrawcz-rdx, line 146: _(note from author)_

This is the least “convenient” spot of the scheduler facade that I ended up with.

Basically, I’d prefer to have just `&Scheduler` here - even `&Scheduler<_, _, _>` (if it compiled) would be sub-optimal (since I would have to update this line when I add 1 more capability to the facade).

(e.g. I considered putting a trait over the entire `Scheduler` struct. But its “builder” nature does not make it so easy, and AFAIS we would lose some type information [i.e. which capabilities are configured] also in the places where we _do_ care about them. I basically want a concise way of saying "give me a `Scheduler` and I don't care what type params it has").

Any suggestions are welcome, but worst-case scenario I can live with what we have here.

malexandru-rdx: Yeah, it looks a bit ugly. Unfortunately type aliases with `impl` are unstable (`type GenericScheduler = Scheduler<impl Spawner, impl Tracker, impl Metrics>`) - let's hope it doesn't take years to get in.

Because in `StateManager` you don't really care about any other capability than `Spawner` you could pass "something" else that only has that - but you lose the whole hierarchical (builder) nature and `named(..)`.

I actually like this pattern and how it ended up, even with this inconvenience. I am fine to go forward hoping that https://github.com/rust-lang/rust/issues/63063 will be sorted out.

malexandru-rdx (approved):
Nice!

jakrawcz-rdx:
The CI is red only because of:
```
✗ Critical severity vulnerability found in zlib/zlib1g
  Description: Integer Overflow or Wraparound
  Info: https://security.snyk.io/vuln/SNYK-DEBIAN12-ZLIB-6008963
  Introduced through: curl@7.88.1-10+deb12u4, openjdk-17/openjdk-17-jre-headless@17.0.8+7-1~deb12u1, apt@2.6.1, dash@0.5.12-2
  From: curl@7.88.1-10+deb12u4 > zlib/zlib1g@1:1.2.13.dfsg-1
  From: openjdk-17/openjdk-17-jre-headless@17.0.8+7-1~deb12u1 > zlib/zlib1g@1:1.2.13.dfsg-1
  From: apt@2.6.1 > apt/libapt-pkg6.0@2.6.1 > zlib/zlib1g@1:1.2.13.dfsg-1
  and 10 more...
  Image layer: Introduced by your base image (debian:12.1-slim)
```
_(I tried apply some quick-fix, but I've read "There is no fixed version for Debian:12 zlib." and that's where my skills end)_

Let me merge this PR and we will treat the required dep fix separately.

PR #726, opened 2023-10-23
Introduce vertex store size limit

LukasGasior1:
This adds a size limit to vertex store, to protect us from the risk of it growing so large that it couldn't be serialized.


core/src/integration/java/com/radixdlt/integration/steady_state/deterministic/consensus/DivergentExecutionLivenessBreakTest.java
```diff
@@ -0,0 +1,293 @@
...
+    return DeterministicTest.builder()
+        .addPhysicalNodes(
+            PhysicalNodeConfig.createSortedBatchWithFakeAddresses(
+                NUM_NODES, KeyComparator.instance()))
+        .messageSelector(MessageSelector.randomSelector(random))
+        .overrideWithIncorrectModule(
+            new AbstractModule() {
+              @Provides
+              @Singleton
+              private StateComputerLedger.StateComputer stateComputer(
```
dhedey, line 132: Does this somehow replace the StateComputer below? Why do we need this? To turn off verification somehow?

Could we add a comment explaining how this is supposed to work?

core/src/integration/java/com/radixdlt/integration/steady_state/deterministic/consensus/DivergentExecutionLivenessBreakTest.java
```diff
@@ -0,0 +1,293 @@
...
+  private static final Random random = new Random(123456);
+
+  @Rule public TemporaryFolder folder = new TemporaryFolder();
+
+  private DeterministicTest createTest() {
+    return DeterministicTest.builder()
+        .addPhysicalNodes(
+            PhysicalNodeConfig.createSortedBatchWithFakeAddresses(
+                NUM_NODES, KeyComparator.instance()))
+        .messageSelector(MessageSelector.randomSelector(random))
```
dhedey, line 127: Why is this needed? Could we explain?

dhedey (commented):
Really nice, although I think we should discuss the edge cases a little more.

core/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
```diff
@@ -580,10 +580,33 @@ private void processGetVerticesErrorResponse(NodeId sender, GetVerticesErrorResp
...
         syncToQC(
             response.highQC(), sender, HighQcSource.RECEIVED_IN_BFT_SYNC_VERTICES_ERROR_RESPONSE);
+      } else {
+        // The node we've asked doesn't have the vertex we've asked for, and it doesn't have a newer
+        // QC.
+        // This is most likely a case where the vertex has been pruned
+        // (due to reaching vertex store max size limit).
+        // We're falling back to request timeout (so no-op here) and will try with a different
+        // (random) peer there.
+        // TODO (most likely during Rust migration): better handle this corner case
```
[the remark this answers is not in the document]
LukasGasior1: That's correct. The `bftSyncing` entry is being cleaned up / removed when we handle the timeout.

LukasGasior1:
As discussed offline: we feel this approach is not adequate and its potential benefits don't counterweight the additional complexity and risks.
Other vertex store overflow mitigation measures will be applied in a follow up PR.

PR #729, opened 2023-10-30
Add a `new_read_only` constructor to `RocksDBStore`.

0xOmarA:
## Description

This PR adds a `new_read_only` constructor to `RocksDBStore` to allow for allow for reading the database while another process writes to it.

### DIFF
--- a/core-rust/state-manager/src/store/rocks_db.rs
+++ b/core-rust/state-manager/src/store/rocks_db.rs
@@ -460,8 +460,8 @@ impl RocksDBStore {
     /// [`ledger-tools`]: https://github.com/radixdlt/ledger-tools
     pub fn new_read_only(root: PathBuf) -> Result<RocksDBStore, DatabaseConfigValidationError> {
         let mut db_opts = Options::default();
-        db_opts.create_if_missing(true);
-        db_opts.create_missing_column_families(true);
+        db_opts.create_if_missing(false);
+        db_opts.create_missing_column_families(false);
 
         let column_families: Vec<ColumnFamilyDescriptor> = ALL_COLUMN_FAMILIES
             .iter()
--- a/core-rust/state-manager/src/store/rocks_db.rs
+++ b/core-rust/state-manager/src/store/rocks_db.rs
@@ -447,6 +447,32 @@ impl RocksDBStore {
         Ok(rocks_db_store)
     }
 
+    /// Creates a readonly [`RocksDBStore`] that allows reading from the store while some other
+    /// process is writing to it. Any write operation that happens against a read-only store leads
+    /// to a panic.
+    pub fn new_read_only(root: PathBuf) -> Result<RocksDBStore, DatabaseConfigValidationError> {
+        let mut db_opts = Options::default();
+        db_opts.create_if_missing(true);
+        db_opts.create_missing_column_families(true);
+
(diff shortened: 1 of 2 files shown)


core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -447,6 +447,32 @@ impl RocksDBStore {
         Ok(rocks_db_store)
     }
 
+    /// Creates a readonly [`RocksDBStore`] that allows reading from the store while some other
+    /// process is writing to it. Any write operation that happens against a read-only store leads
+    /// to a panic.
+    pub fn new_read_only(root: PathBuf) -> Result<RocksDBStore, DatabaseConfigValidationError> {
+        let mut db_opts = Options::default();
+        db_opts.create_if_missing(true);
+        db_opts.create_missing_column_families(true);
```
jakrawcz-rdx, line 456: I wonder how this 2x `true` really behave while `read_only` 🤔 

I suspect that regardless of `true` vs `false`, we would see a panic.
But with `false` this panic will be more accurate (i.e. "the logic expected [some CFs] but they don't exist in DB").

jakrawcz-rdx (approved):
Very cool 👍  I only see one potential very-weak error-surfacing improvement for a rare case when e.g. someone runs a newer version of `ledger-tools` on a DB opened by some older version of a Node.
(where the "version difference" == "some new CF was introduced")

core-rust/state-manager/src/store/rocks_db.rs
dhedey: Yeah, I think I'd rather these both be false I think?

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -447,6 +447,32 @@ impl RocksDBStore {
         Ok(rocks_db_store)
     }
 
+    /// Creates a readonly [`RocksDBStore`] that allows reading from the store while some other
+    /// process is writing to it. Any write operation that happens against a read-only store leads
+    /// to a panic.
```
dhedey, line 452: Could we explain _why_ we have this? / why we need to keep this?

As I don't believe any node developers think of the node repo as being a library, so we typically just use `pub` for `pub (crate)` and might just delete this bit of code thinking it's unused :D

dhedey (approved):

malexandru-rdx (changes requested):
What is the use case for this? I don't think the read only instance is giving any guarantees about data consistency. AFAIU, this view is always "catching up" with the write. This means that 2 sequent reads may be from different state versions, is this ok?

I'm marking this for request changes because I want to make sure this is acknowledged.

core-rust/state-manager/src/store/rocks_db.rs
0xOmarA: Good point! Will add to this doc comment explaining why its needed and why it has no usages in the node code.

core-rust/state-manager/src/store/rocks_db.rs
0xOmarA: Looks like `false` make more sense here. I've made this change.

malexandru-rdx (approved):
As discussed on slack, this is fine for the use cases. Later we will have snapshots, which will provide data consistency guarantees. Until then, a warning/note about this should be fine!

PR #736, opened 2023-11-03
Impl of `/browse/entity/iterator` endpoint (mostly: paging infra)

jakrawcz-rdx:
⚠️ This builds on top of https://github.com/radixdlt/babylon-node/pull/733 (re-uses some code) and ultimately wants to merge into `feature/browse_api` _(not `develop`!)_.

This implements https://radixdlt.atlassian.net/wiki/spaces/RPNV1/pages/3189211214/RNP-12+-+Browse+API+sub-API+within+Core+API#1.-Entity-Listing but the main focus of the PR is on re-usable listing-with-pages-on-continuation-tokens infra (please see the heavily-documented abstractions introduced in `paging.rs`).


core-rust/core-api-server/src/core_api/browse.rs
```diff
@@ -186,6 +194,100 @@ impl ObjectFieldMeta {
     }
 }
 
+/// A lister of Engine's Nodes.
+pub struct EngineNodeLister<'s, S> {
+    database: &'s S,
+}
+
+impl<'s, S: QueryableProofStore + ReadableTreeStore> EngineNodeLister<'s, S> {
```
jakrawcz-rdx, line 202: _(note from author)_

This StateHashTree-based impl may be regarded as temporary, if we want to migrate to NodeAncestryRecord-based impl: see discussion at https://radixdlt.atlassian.net/wiki/spaces/RPNV1/pages/3189211214/RNP-12+-+Browse+API+sub-API+within+Core+API#Implementation-details

Still, I consider it good-enough to be merged in v1 of this API.

core-rust/core-api-server/src/core_api/handlers/browse_entity_iterator.rs
```diff
@@ -0,0 +1,86 @@
...
+use radix_engine::types::*;
+
+use std::ops::Deref;
+use std::time::Duration;
+
+/// A default maximum page size (can be further limited by each request).
+const DEFAULT_MAX_PAGE_SIZE: usize = 10000;
+
+/// A maximum wallclock time spent on iteration.
+const MAX_ITERATION_DURATION: Duration = Duration::from_millis(100);
```
malexandru-rdx, line 12: Maybe we should do this for all paginated/stream endpoints?

malexandru-rdx (approved):
Nice!

core-rust/core-api-server/src/core_api/paging.rs
```diff
@@ -0,0 +1,277 @@
...
+use std::ops::Add;
+use std::time::{Duration, Instant};
+
+/// A sequence of deterministically-ordered elements, which knows how to efficiently start iteration
+/// from an arbitrarily requested item.
+pub trait RandomAccessIterable<T, I: Iterator<Item = T>> {
+    /// Turns itself into an iterator which:
+    /// - either starts at the beginning of the sequence (if `from.is_none()`),
+    /// - or starts at the given `from` element (if it is actually exists in the sequence),
+    /// - or at the first actually-existing element after the given `from` (if the from was to be
```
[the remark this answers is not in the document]
jakrawcz-rdx: basically in that last point I was trying to cover the following case:

The sequence is `[m, u, s, e]`.
I come in and say: "hi, please give me iterator starting from `o`".

I have no idea that `o` is not in the sequence. Maybe I thought it is there, maybe it was there in the past - doesn't matter.
Still, the question is: where should the iterator start?

The sequence has some order of elements (knows it internally), but I (from the outside) can make no assumptions about the order of elements in the sequence`*`.

And the contract defines the following response: "imagine that `o` existed in the sequence, and start iteration from the next element after that imaginary `o`". The `o` kinda "has its place", it is just not present there. So the starting `o` point is "exclusive", but only because `o` isn't there, so it can't be "inclusive".

OFC if I asked for "start from `s`", I would get `[s, e]` back (inclusive).

`*` This actually matches our impl e.g. in JMT (which uses a prefix of 20 hashed bytes for JMT's reasons).

PR #738, opened 2023-11-08
Refactor: TypedCf (and associated traits) use GATs

malexandru-rdx:
This is a pure refactor with no intended functional changes. We remove the generics from `TypedCf<K, V, KC, VC>` in favour of associated types. This way we (and more importantly the compiler) has a clearer understanding of the relationships between `K`, `V`, `KC` and `VC`. This also reduces the "boilerplate types" needed to be passed around everywhere we use `TypedCf`.

This is the first PR in the DB snapshotting epic, since it's required/simplifies the work a lot.


jakrawcz-rdx (approved):
cool!

I admire the small size of this PR, and that it cuts the refactor out from the business changes, and also that it happens before the business changes 👍

PR #739, opened 2023-11-09
Impl of `/browse/entity/info` endpoint

jakrawcz-rdx:
⚠️ This builds on top of https://github.com/radixdlt/babylon-node/pull/736 (refactors some stuff introduced there, in order to re-use it here) and ultimately wants to merge into `feature/browse_api` _(not `develop`!)_.

This implements https://radixdlt.atlassian.net/wiki/spaces/RPNV1/pages/3189211214/RNP-12+-+Browse+API+sub-API+within+Core+API#2.-Entity-Information .


core-rust/core-api-server/src/core_api/browse.rs
```diff
@@ -63,67 +116,348 @@ impl<'s, S: SubstateDatabase> EngineStateMetaLoader<'s, S> {
...
-                    .expect("guaranteed by max field count")
-            })
-            .map(|field_index| {
+        let resolver = TypeReferenceResolver::new(&self.reader);
+
+        let fields = fields
+            .into_iter()
+            .flat_map(|(_partition_description, fields)| fields)
+            .enumerate()
+            .map(|(index, schema)| {
```
jakrawcz-rdx, line 266: _(note from author)_

I just realized that this logic does NOT filter out the fields disabled by features. It is not that trivial (e.g. needs another DB read to support "feature on outer" condition), but the fix is coming (maybe even in this PR).

jakrawcz-rdx:
I am merging all these PRs into long-lived `feature/browse_api` (since it becomes too cumbersome to track the long PR dep chain)

PR #742, opened 2023-11-13
Fix: Key store recreation with empty root path and AwsSecrets CLI

dhedey:


cli-tools/src/test/java/com/radixdlt/cloud/AWSSecretsTest.java
```diff
@@ -0,0 +1,89 @@
...
+          "--enable-aws-secrets=false",
+          "--recreate-aws-secrets=false",
+          "--network-name=dumunet",
+          "--node-name-prefix=node",
+          "--node-number=0",
+          "--node-names=dumunet_ap_south_1_node0,dumunet_eu_west_1_node0,dumunet_eu_west_1_node1,dumunet_eu_west_1_node2,dumunet_eu_west_1_node3",
+          "--path="
+              + Paths.get(
+                  System.getProperty("java.io.tmpdir"),
+                  UUID.randomUUID().toString().replaceAll("[^a-zA-Z]", ""))
```
jakrawcz-rdx, line 85: do we already have the right deps here to do `@Rule public TemporaryFolder folder = new TemporaryFolder();` ?

If yes, then I recommend it.
(you will be able to do just `.newFile()` instead of inventing the "UUID name gen" + the file will auto-delete self)

cli-tools/src/main/java/com/radixdlt/cloud/AWSSecrets.java
```diff
@@ -260,6 +270,7 @@ private static void generateAndStoreKey(
             publicKeyFileAwsSecret, publicKeyFileSecretName, awsSecretsOutputOptions, false, false);
       } catch (Exception e) {
         System.out.println(e);
+        throw new RuntimeException(e);
```
jakrawcz-rdx, line 273: (nitpick)

If you re-throw, please skip the log.
(in 99% cases it only leads to a duplicated log)

jakrawcz-rdx (approved):
Interesting case of "null out of f*** ~~nowhere~~ JRE"; and thanks for the test 👍 

I only have some nitpicking comments there.

cli-tools/src/test/java/com/radixdlt/cloud/AWSSecretsTest.java
dhedey: Yes, totally agree - thanks

cli-tools/src/main/java/com/radixdlt/cloud/AWSSecrets.java
dhedey: Agreed, thanks. Good spot

PR #747, opened 2023-11-17
Impl for Programmatic JSON deserializer

jakrawcz-rdx:
⚠️ This builds on top of https://github.com/radixdlt/babylon-node/pull/744 and ultimately wants to merge into `feature/browse_api` _(not `develop`!)_.

This is in the "Browse API" PR-chain, but actually it is a pretty generic feature:
- Implemented Programmatic JSON **deser**ializer 
  - We need it now (yeah for Browse API), and Engine does not offer one (they only serialize)
  - The impl is stolen from RET (and then heavily adjusted for Node's specifics; but not beyond recognition)
  - The plan is to either move this impl to Engine (or implement the same on the Engine's side), and we'll then be able to remove this code from Node (and maybe even from RET?)
- created the Programmatic JSON **ser**ializer facade
  - Just so that it can live next to the **deser**ializer
  - ofc it is only facading the Engine's serialization impl
  - it is used to test the **deser**ializer for compatibility with Engine's **ser**ializer (the "round-trip codec test")

Note: The **ser**ializer facade is used right away, but the **deser**ializer is currently unused (and rightfully produces warnings) [except in tests]; it will soon be used by Browse API's "lookup kv store/collection" endpoints.

### DIFF
--- a/core-rust/Cargo.lock
+++ b/core-rust/Cargo.lock
@@ -33,6 +33,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
 
 [[package]]
+name = "android_system_properties"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


jakrawcz-rdx:
I am merging all these PRs into long-lived `feature/browse_api` (since it becomes too cumbersome to track the long PR dep chain)

PR #749, opened 2023-11-21
feat: add transaction balance changes opt-in

krzlabrdx:
- new opt-in for `TransactionFormatOptions`: `balance_changes`
- when opted-in `CommittedTransaction.balance_changes` property should be a practical equivalent of the `/lts/stream/transaction-outcomes` endpoint response
- used LTS types in the `/stream/transactions` endpoint


krzlabrdx:
Open question: shouldn't we bump API version from 1.0.4 to 1.0.5?

kofrdx:
Added a change on the Dockerfile since I noticed the failed build while analyzing workflow costs.
```
#12 26.01  openjdk-17-jdk : Depends: openjdk-17-jre (= 17.0.8+7-1~deb12u1) but 17.0.9+9-1~deb12u1 is to be installed
[2138](https://github.com/radixdlt/babylon-node/actions/runs/6942445992/job/18885449594#step:17:2143)
#12 26.01                   Depends: openjdk-17-jdk-headless (= 17.0.8+7-1~deb12u1) but 17.0.9+9-1~deb12u1 is to be installed
```

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -7146,6 +7149,8 @@ components:
           $ref: "#/components/schemas/LedgerTransaction"
         receipt:
           $ref: "#/components/schemas/TransactionReceipt"
+        outcomes:
```
dhedey, line 7152: I think "outcome" would be more appropriate (instead of `outcomes`) but even than I'm not sure of.

I think I'd either prefer (A) including the full LtsTransactionOutcome in this directly, or (B) just calling this balance changes, like we did in the Gateway.

I don't mind which, but I was never entirely happy with the name "outcome" and I'd rather not introduce another model which is different and also labelled as the outcome :D.

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -7350,6 +7355,39 @@ components:
         payload_hex:
           type: string
           description: The hex-encoded system transaction payload. Only returned if enabled in TransactionFormatOptions on your request.
+    TransactionOutcomes:
+      description: |
+        For the given transaction, contains individual entity resource balance changes.
+        The balance changes accounts for the fee payments as well.
+        Current implementation does not take into account recalls, but this will change in a future update.
```
dhedey, line 7362: Ohh both of these should be fixed now. Could this be fixed here/wherever you copy-pasted from?

dhedey (commented):
Thanks for picking this up! 👍 A few tweaks then happy to approve

Dockerfile
```diff
@@ -30,7 +30,7 @@ RUN apt-get update \
     wget=${WGET_VERSION} \
     software-properties-common=0.99.30-4 \
   && apt-get install -y --no-install-recommends \
-    openjdk-17-jdk=17.0.8+7-1~deb12u1 \
```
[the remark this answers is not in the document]
dhedey: Oh that was Kim's drive by - all good :D

> Added a change on the Dockerfile since I noticed the failed build while analyzing workflow costs.

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -5749,7 +5752,6 @@ components:
       description: |
         For the given transaction, contains the status, total fee summary and individual entity resource balance changes.
         The balance changes accounts for the fee payments as well.
-        Current implementation does not take into account recalls, but this will change in a future update.
         For failed transactions, current implementation does not return any balance changes (not even the fee payments).
```
dhedey, line 5755: This "for failed transactions" bit also isn't true and should be removed :)

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -7146,12 +7148,46 @@ components:
...
+          $ref: "#/components/schemas/CommittedTransactionBalanceChanges"
         proposer_timestamp_ms:
           type: integer
           format: int64
           minimum: 0
           maximum: 100000000000000
           description: An integer between `0` and `10^14`, marking the proposer timestamp in ms.
+    CommittedTransactionBalanceChanges:
+      description: |
+        For the given transaction, contains individual entity resource balance changes.
```
dhedey, line 7161: A lot of this isn't right - can we change this to:
```
The fee-related and non-fee-related balance changes of a transaction, aggregated by resource and global entity within which the change occurred.
```

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -7146,12 +7148,46 @@ components:
...
+        The balance changes accounts for the fee payments as well.
+        For failed transactions, current implementation does not return any balance changes (not even the fee payments).
+        This will also change in a future update.
+      type: object
+      required:
+        - fungible_entity_balance_changes
+        - non_fungible_entity_balance_changes
+        - resultant_account_fungible_balances
```
dhedey, line 7169: I think we could/should probably ditch `resultant_account_fungible_balances` - or move then to a separate opt-in in future - what do you think?

dhedey (commented):
Looking better! A few minor tweaks to go, then we need to regenerate the Java + Core API models, and check the tests compile/pass. Looks like there's currently compile errors with the Java generated models somehow

core-rust/core-api-server/core-api-schema.yaml
krzlabrdx: 1. Do we want to keep CommittedTransactionBalanceChanges and existing LtsEntityFungibleBalanceChanges descriptions identical? I believe they should be.
2. What you proposed seems like a valid description for `CommittedTransactionBalanceChanges.fungible_entity_balance_changes` not the root type itself.

core-rust/core-api-server/core-api-schema.yaml
krzlabrdx: Opt-in within opt-in? I'd be hesitant. We'd have to move those `resultant_account_fungible_balances` into its own area. But then again... the idea behind this PR was to reuse LTS models. But maybe we should abandon this and model non-LTS endpoint from scratch? That would result in some OAS type "duplication" I wanted to avoid.

core-rust/core-api-server/core-api-schema.yaml
dhedey: 1. Sounds sensible :+1:
2. I still think it's useful to highlight it on that type, to hint they need to look inside the `fungible_entity_balance_changes`

core-rust/core-api-server/core-api-schema.yaml
dhedey: I meant a separate opt-in, not an opt-in inside an opt-in.

Let's just remove `resultant_account_fungible_balances` from this model as it's not balance changes, and I can't see the product need right now.

core-rust/core-api-server/core-api-schema.yaml
krzlabrdx: Changed description / dropped resultant account balances

core-rust/core-api-server/core-api-schema.yaml
krzlabrdx: Changed description / dropped resultant account balances

jakrawcz-rdx (approved):
The PR name/description talks about `outcomes`, but I see it is about `balance_changes`.

But I assume this is the intention and this is what NG desperately needs. If so, then LGTM 👍 (and my 2 minor comments are non-blocking)

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -7152,12 +7152,36 @@ components:
           $ref: "#/components/schemas/LedgerTransaction"
         receipt:
           $ref: "#/components/schemas/TransactionReceipt"
+        balance_changes:
+          $ref: "#/components/schemas/CommittedTransactionBalanceChanges"
```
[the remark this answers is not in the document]
krzlabrdx: Added extra description.

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -7152,12 +7152,36 @@ components:
...
         proposer_timestamp_ms:
           type: integer
           format: int64
           minimum: 0
           maximum: 100000000000000
           description: An integer between `0` and `10^14`, marking the proposer timestamp in ms.
+    CommittedTransactionBalanceChanges:
+      description: |
+        The fee-related and non-fee-related balance changes of a transaction, aggregated by resource and global entity
```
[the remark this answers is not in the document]
krzlabrdx: Changed.

core-rust/core-api-server/core-api-schema.yaml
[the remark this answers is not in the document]
jakrawcz-rdx: ⚠️ if I understand the OpenAPI spec right, the `description: Only ...` you have here and the `description: All balance ...` (which is literally "kinda-included" by the power of `$ref:`) will NOT be merged together in any elegant way by the tools.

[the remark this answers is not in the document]
dhedey: Technically putting a description alongside a ref is invalid according to the Open API schema specification.
I typically discuss optional fields on the parent's description for this reason.

PR #760, opened 2023-12-04
Another approach to fixing the invalid key range calculation of a deleted partition

jakrawcz-rdx:
This achieves the same as https://github.com/radixdlt/babylon-node/pull/757 but in a more "local" way + added context in docs.


jakrawcz-rdx:
abandoning this in favor of https://github.com/radixdlt/babylon-node/pull/763

PR #761, opened 2023-12-05
Feature: Add secondary instance support

talekhinezh:
Added support for secondary instances with RocksDB (ie added `new_as_secondary` and `try_catchup_with_primary` functions to `RocksDB`).


LukasGasior1 (approved):
LGTM!

PR #763, opened 2023-12-06
Clean-up after the "delete range" bugfix 

jakrawcz-rdx:
This PR cleans up some minor aspects of https://github.com/radixdlt/babylon-node/pull/757:
- Simplifying the actual `prefix_to_range()` impl:
  - it was `[partition:0, next_partition:0)` (+ `to = 0xffff...` for special cases)
  - now: `[partition:0, partition:0xffff...)` (uniformly for all cases)
  - **I acknowledge this is a logic change (in a refactor PR)** but actually I use the logic which was already there for one of special cases (and it seems more bulletproof than calculating `next_partition`)
  - all the tests added in the fix's PR are passing unchanged
- Refining API of those `DbCodec`s "extra traits"
- Renaming them 
- Adding more rustdocs with clarifications and examples
- Moving codecs to separate file


core-rust/state-manager/src/store/typed_cf_api.rs
```diff
@@ -399,17 +399,69 @@ pub trait DbCodec<T> {
...
+///   - grouping broken: even though John and Ann belong to the same family, they *do not* share
+///     any well-specified prefix in their byte representations.
+///   - a lexicographically-ordered range covering all members of a family *cannot* be constructed
+///     under this encoding.
+///   - the grouping in this case *could* be preserved e.g. by encoding `<last_name> <first_name>`
+///     (with a variable-length prefix, defined as "everything before the first space character").
+///
+/// The grouping preservation is important for database *key* codecs of column families which need
+/// to support e.g. a batch delete operation of an entire group.
+pub trait GroupPreservingDbCodec {
```
dhedey, line 454: We're missing a `GroupedOrderPreservingDbCodec` which enables `delete_group_range` and `iterate_group_from`

core-rust/state-manager/src/store/typed_cf_api.rs
```diff
@@ -399,17 +399,69 @@ pub trait DbCodec<T> {
...
+///   - `1u32` <-> `[49]`,
+///   - `7u32` <-> `[55]`,
+///   - `259u32` <-> `[50, 53, 59]`,
+///   - order broken: the right side *does not* consistently increase lexicographically (the bytes
+///     starting with `[50, ...]` are lexicographically before `[55]`).
+///
+/// The order preservation is important for database *key* codecs of column families which need to
+/// support e.g. iteration of elements starting from a particular element, or any batch operations
+/// defined by `[from, to]` ranges.
 pub trait OrderPreservingDbCodec {}
```
dhedey, line 424: This should also power `iterate_from`

dhedey (commented):
This PR needs to address more IMO to properly fix the issues:

* `get_first`, `get_first_value`, `get_last`, `get_last_key`, `get_last_value`, `iterate`, `iterate_from`, `iterate_with_mode` all need moving from the general `TypedCfApi` impl into the impl scoped to keys which implement`OrderPreservingDbCodec` because they all suffer the same problem.
* We need a separate `GroupedOrderPreservingDbCodec` which supports similar methods under a given grouping. (Notably, the substate keys).

core-rust/state-manager/src/store/typed_cf_api.rs
```diff
@@ -261,13 +262,12 @@ impl<'db, 'wb, CF: TypedCf> TypedCfApi<'db, 'wb, CF> {
...
-impl<'db, 'wb, KC: PrefixableDbCodec, CF: TypedCf<KeyCodec = KC>> TypedCfApi<'db, 'wb, CF> {
-    /// Deletes the entries from the given prefix.
-    pub fn delete_all_under_prefix(&self, prefix: &KC::Prefix) {
-        let (from, to) = self.key_codec.encode_prefix_to_range(prefix);
-        // Frustratingly, RocksDB doesn't allow an open range, so we have to work around that
-        let to = to.unwrap_or_else(|| self.key_codec.encode_key_greater_than_all_possible(prefix));
-        self.write_buffer.delete_range(self.cf_handle, from, to);
+impl<'db, 'wb, KC: GroupPreservingDbCodec, CF: TypedCf<KeyCodec = KC>> TypedCfApi<'db, 'wb, CF> {
+    /// Deletes the entries from the given group.
+    pub fn delete_group(&self, group: &KC::Group) {
```
dhedey, line 267: I like this abstraction :D

core-rust/state-manager/src/store/typed_cf_api.rs
```diff
@@ -399,17 +399,69 @@ pub trait DbCodec<T> {
     fn decode(&self, bytes: &[u8]) -> T;
 }
 
-/// A marker trait which must only be implemented on Codecs which preserve lexicographic ordering.
+/// A marker trait which must only be implemented on [`DbCodec`]s which preserve the business-level
```
dhedey, line 402: Really nice comments, thank you! :D

core-rust/state-manager/src/store/codecs.rs
```diff
@@ -0,0 +1,435 @@
...
+
+    fn decode(&self, bytes: &[u8]) -> StateVersion {
+        StateVersion::from_be_bytes(bytes)
+    }
+}
+
+impl OrderPreservingDbCodec for StateVersionDbCodec {}
+
+#[derive(Default)]
+pub struct EpochDbCodec {}
```
dhedey, line 93: This can also be `OrderPreservingDbCodec`

core-rust/state-manager/src/store/codecs.rs
```diff
@@ -0,0 +1,435 @@
...
+        value.number().to_be_bytes().to_vec()
+    }
+
+    fn decode(&self, bytes: &[u8]) -> Epoch {
+        Epoch::of(u64::from_be_bytes(copy_u8_array(bytes)))
+    }
+}
+
+#[derive(Default)]
+pub struct ScenarioSequenceNumberDbCodec {}
```
dhedey, line 106: This can also be `OrderPreservingDbCodec`

dhedey (commented):
Additional comments which were meant to be a part of the previous review

jakrawcz-rdx (commented):
@dhedey thanks for thorough review!

I initially wanted to avoid introducing all of this in a single PR, but it turns out to be not that massive 👍 
One surprise is: I needed to add the high-level "iterate groups" operation (needed because of https://github.com/radixdlt/babylon-node/pull/709).

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -1231,10 +1231,7 @@ impl ListableSubstateDatabase for RocksDBStore {
         Box::new(
             self.open_db_context()
                 .cf(SubstatesCf)
-                .iterate(Direction::Forward)
-                .map(|(iter_key_bytes, _)| iter_key_bytes.0)
-                // Rocksdb iterator returns sorted entries, so ok to to eliminate
-                // duplicates with dedup()
+                .iterate_key_groups()
                 .dedup(),
```
dhedey, line 1235: I don't think we need the dedup any more, right?

dhedey (approved):
Nice! Much better 👍. One minor thing left. Happy for you to merge when you've fixed and are happy.

jakrawcz-rdx:
> Also - slight preference for `GroupedOrderPreservingDbCodec` over `GroupOrderPreservingDbCodec` because the latter suggests it's the GroupOrder being preserved, rather than the order inside the grouping.

true, but `GroupedOrderPreservingDbCodec` does not immediately suggest "preserves order inside the grouping" to me either :(

hence... `IntraGroupOrderPreservingDbCodec`?

core-rust/state-manager/src/store/rocks_db.rs
jakrawcz-rdx: ahh yes, neither the `Box::new()` :)

jakrawcz-rdx:
(only one `MultiNodeRebootTest` failed)

PR #772, opened 2023-12-12
Handle ledger sync queries for a higher state version than node's latest

LukasGasior1:


core/src/main/java/com/radixdlt/rev2/REv2TransactionsAndProofReader.java
```diff
@@ -101,6 +101,18 @@ public REv2TransactionsAndProofReader(REv2TransactionAndProofStore transactionSt
...
+    final var latestStateVersion =
+        transactionStore
+            .getLastProof()
+            .map(proof -> proof.ledgerHeader().stateVersion().toLong())
+            .orElse(0L);
+
+    // Early return if we got a query for a higher
+    // state version than our latest.
+    if (startStateVersionInclusive > latestStateVersion) {
+      return null;
```
jakrawcz-rdx, line 113: this will still log `[WARN] REMOTE_SYNC_REQUEST: ...` ([here](https://github.com/radixdlt/babylon-node/blob/0c342dc7fd8c8f51a0dddcfed7801c7c817227fd/core/src/main/java/com/radixdlt/sync/RemoteSyncService.java#L136)) so I guess not all Discord questions will disappear :D
_(but ofc better than error)_

jakrawcz-rdx (approved):

core/src/main/java/com/radixdlt/rev2/REv2TransactionsAndProofReader.java
LukasGasior1: Ah, good catch. I guess we should just remove that

LukasGasior1: or downgrade it to `trace`

dhedey: Maybe `debug` is better than trace?

LukasGasior1: this has changed a bit :) https://github.com/radixdlt/babylon-node/pull/790/files#diff-bcb79e17117d0fede36b9660b21ea61ab98802dd6d3e043ecc6481de01e67346R110
this is now present in metrics (UnfulfilledSyncRequest with a NOTHING_TO_SERVE_AT_THE_GIVEN_STATE_VERSION tag)

PR #773, opened 2023-12-12
Fix sbor readBytes Decoder for empty arrays

LukasGasior1:


common/src/main/java/com/radixdlt/sbor/coding/Decoder.java
```diff
@@ -242,6 +242,14 @@ public String readString() {
   public byte[] readBytes(int length) {
     var bytes = new byte[length];
 
+    // Early return if the length is 0.
+    // This is needed because, contrary to JavaDoc,
```
jakrawcz-rdx, line 246: This is **_FALSE ACCUSATION_**, they DO NOT "lie in the javadocs"...

... they do something MUCH WORSE: they redefine the contract of the overridden method :D

image

good catch 👍

jakrawcz-rdx (approved):

common/src/main/java/com/radixdlt/sbor/coding/Decoder.java
LukasGasior1: omg :man_facepalming: I updated the comment. We don't want no false accusations here

PR #781, opened 2024-01-02
Architecture: Remove mut requirement from hash tree

talekhinezh:


core-rust/state-manager/src/staging/result.rs
```diff
@@ -538,16 +538,16 @@ pub struct HashStructuresDiff {
 #[derive(Clone, Debug)]
 pub struct StateHashTreeDiff {
     pub new_root: StateHash,
-    pub new_nodes: Vec<(NodeKey, TreeNode)>,
-    pub stale_tree_parts: Vec<StaleTreePart>,
+    pub new_nodes: RefCell<Vec<(NodeKey, TreeNode)>>,
+    pub stale_tree_parts: RefCell<Vec<StaleTreePart>>,
```
jakrawcz-rdx, line 542: could we, instead, make `CollectingTreeStore` hold a `diff: RefCell<StateHashTreeDiff>`?

PR #790, opened 2024-01-09
Feature/coordinated protocol updates execution

LukasGasior1:


cli-tools/src/main/java/com/radixdlt/shell/LedgerFileSync.java
```diff
@@ -1,192 +0,0 @@
...
-import java.io.FileInputStream;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.nio.ByteBuffer;
-import java.util.List;
-import java.util.Objects;
-import javax.annotation.concurrent.Immutable;
-
-/** Utility class to write/restore ledger sync data from a file. */
-public final class LedgerFileSync {
```
LukasGasior1, line 90: This is no longer used/needed.

core/src/main/java/com/radixdlt/api/system/routes/HealthHandler.java
```diff
@@ -66,22 +66,21 @@
...
 
 public final class HealthHandler extends SystemGetJsonHandler<HealthResponse> {
   private final HealthInfoService healthInfoService;
-  private final ProtocolVersion currentProtocolVersion;
+  private final RustStateComputer rustStateComputer;
 
   @Inject
-  HealthHandler(
-      HealthInfoService healthInfoService, @Current ProtocolVersion currentProtocolVersion) {
+  HealthHandler(HealthInfoService healthInfoService, RustStateComputer rustStateComputer) {
```
talekhinezh, line 80: Any way we can remove this dependency on `RustStateComputer` for example by using messages? Seems pretty bad to have this API contend with the protocol state lock with the state computer.

Otherwise, this is probably not too bad of a hack but we should add a comment.

core-rust/state-manager/src/protocol/flash_templates.rs
```diff
@@ -0,0 +1,32 @@
...
+    ConsensusManagerField, VersionedConsensusManagerConfiguration,
+};
+use radix_engine::system::system_substates::{FieldSubstate, FieldSubstateV1, LockStatus};
+use radix_engine::track::StateUpdates;
+use radix_engine_common::prelude::{scrypto_encode, CONSENSUS_MANAGER};
+use radix_engine_interface::blueprints::consensus_manager::ConsensusManagerConfig;
+use radix_engine_interface::prelude::MAIN_BASE_PARTITION;
+use radix_engine_store_interface::interface::DatabaseUpdate;
+
+pub fn consensus_manager_config_flash(new_config: ConsensusManagerConfig) -> StateUpdates {
```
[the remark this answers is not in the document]
LukasGasior1: This is also used for the real update, where's the scrypto version?

core-rust/state-manager/src/transaction/ledger_transaction.rs
```diff
@@ -130,9 +130,26 @@ const GENESIS_TRANSACTION_SYSTEM_TRANSACTION_DISCRIMINATOR: u8 = 1;
...
 #[derive(Debug, Clone, PartialEq, Eq, ManifestCategorize, ManifestEncode, ManifestDecode)]
 pub struct FlashTransactionV1 {
+    pub nonce: u64,
     pub state_updates: StateUpdates,
 }
 
+impl FlashTransactionV1 {
+    pub fn prepare(&self) -> Result<PreparedFlashTransactionV1, PrepareError> {
+        let mut data_to_hash = scrypto_encode(&self.state_updates).unwrap();
+        data_to_hash.extend_from_slice(self.nonce.to_le_bytes().as_slice());
```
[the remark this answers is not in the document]
LukasGasior1: not sure I understand, do you mean to use the `HashAccumulator` (as in `RoundUpdateTransactionV1.prepare()`), or something else?

core/src/main/java/com/radixdlt/api/system/routes/HealthHandler.java
LukasGasior1: Good point! Gonna take a look at that

core-rust-bridge/src/main/java/com/radixdlt/protocol/ProtocolUpdateEnactmentCondition.java
```diff
@@ -0,0 +1,131 @@
...
+        codecs ->
+            EnumCodec.fromPermittedRecordSubclasses(
+                ProtocolUpdateEnactmentCondition.class, codecs));
+  }
+
+  static ProtocolUpdateEnactmentCondition singleReadinessThresholdBetweenEpochs(
+      long minEpoch,
+      long maxEpoch,
+      Decimal requiredPercentageStakeSupported,
+      long requiredConsecutiveEpochsOfSupport) {
```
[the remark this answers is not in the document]
LukasGasior1: Sure, I can add a comment. This doesn't have to be true, protocol configuration is independent of the current ledger state.

core-rust/state-manager/src/protocol/mainnet_updates/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+use crate::{
+    NoStateUpdatesProtocolUpdater, ProtocolUpdater, ProtocolUpdaterFactory, StateManagerDatabase,
+    ANEMONE_PROTOCOL_VERSION, GENESIS_PROTOCOL_VERSION,
+};
+use node_common::locks::StateLock;
+use radix_engine_common::network::NetworkDefinition;
+use std::sync::Arc;
+
+pub struct MainnetProtocolUpdaterFactory {
+    network: NetworkDefinition,
```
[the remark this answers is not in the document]
LukasGasior1: We'll want to reuse this config (or parts of it) on testnets. In fact, that's what `TestingDefaultProtocolUpdaterFactory` does: it delegates the execution to `MainnetProtocolUpdaterFactory` if the protocol version name matches (even if it was enacted due to a different condition). I guess we could rename it to `ProductionProtocolUpdaterFactory` or something?

core-rust/state-manager/src/protocol/protocol_config.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub required_consecutive_completed_epochs_of_support: u64,
+}
+
+impl ProtocolConfig {
+    pub fn mainnet() -> ProtocolConfig {
+        Self {
+            genesis_protocol_version: GENESIS_PROTOCOL_VERSION.to_string(),
+            protocol_updates: vec![ProtocolUpdate {
+                next_protocol_version: ANEMONE_PROTOCOL_VERSION.to_string(),
+                enactment_condition:
```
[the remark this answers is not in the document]
LukasGasior1: Yup, once the update occurs we'll be able to update this to `EnactUnconditionallyAtEpoch` (formerly known as "fixed epoch") but it's not required, just an optimization / for clarity.

core-rust/state-manager/src/protocol/protocol_state.rs
```diff
@@ -0,0 +1,511 @@
...
+
+// This file contains types and utilities for
+// managing the (dynamic) protocol state of a running node.
+
+#[derive(Clone, Debug, Eq, PartialEq, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
+pub struct ProtocolState {
+    pub current_epoch: Option<Epoch>,
+    pub current_protocol_version: String,
+    pub enacted_protocol_updates: BTreeMap<StateVersion, String>,
+    pub unenacted_protocol_updates: Vec<UnenactedProtocolUpdate>,
```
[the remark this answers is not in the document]
LukasGasior1: Yeah, I think it's actually a better name (I considered it, but rejected for the reason I no longer remember).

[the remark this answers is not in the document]
LukasGasior1: I know, I considered it in "ProtocolConfig", but "pending" doesn't really fit there because at the (static) configuration level the node doesn't know if it's "pending" or "past". Okay to use it in (dynamic) state though

core-rust/state-manager/src/protocol/protocol_update.rs
```diff
@@ -0,0 +1,489 @@
...
+        Self {
+            protocol_version_name,
+            state_computer_configurator: StateComputerConfigurator::default(network),
+        }
+    }
+}
+
+impl ProtocolUpdater for NoStateUpdatesProtocolUpdater {
+    fn protocol_version_name(&self) -> String {
+        self.protocol_version_name.clone()
```
jakrawcz-rdx, line 122: (just a hunch) Whenever I see the same method impl in each struct implementing a trait, I suspect the responsibility should be moved one level higher.

core-rust/state-manager/src/state_computer.rs
```diff
@@ -1210,12 +1278,55 @@ where
...
+
         Ok(CommitSummary {
             validator_round_counters: round_counters,
             num_user_transactions,
         })
     }
 
+    pub fn apply_protocol_update(
+        &self,
+        protocol_updater: &dyn ProtocolUpdater,
```
jakrawcz-rdx, line 1293: (minor) if we wanted to limit the spread of `dyn`s, we could do `protocol_updater: &P` here, and it would work, right? (`<P: ProtocolUpdater + ?Sized>`)

core-rust/state-manager/src/protocol/protocol_update.rs
```diff
@@ -0,0 +1,489 @@
...
+        };
+
+        let mut series_executor = TransactionSeriesExecutor::new(
+            read_store.deref(),
+            &execution_cache,
+            &self.execution_configurator,
+            dummy_protocol_state,
+        );
+
+        let mut committed_transaction_bundles = Vec::new();
```
jakrawcz-rdx, line 320: The ~70 lines starting here mimic the `StateComputer.commit()` (quite understandably :p).
We definitely were incrementally adding stuff to that logic, and it may happen that we add more. Forgetting about protocol update's "copy" would be very sad (and possibly only caught at next protocol update).

Hence I think it is a good moment to actually address the TODO left there: https://github.com/radixdlt/babylon-node/blob/8445409a552b176a0d7c23b220208b25603f243c/core-rust/state-manager/src/state_computer.rs#L1135 (and then re-use the resultant "committer" helper here)

Since we are in a rush, we can live with the copy-paste + another `TODO(urgent)` here, for a moment.

core-rust/state-manager/src/protocol/flash_templates.rs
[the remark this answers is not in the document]
talekhinezh: https://github.com/radixdlt/radixdlt-scrypto/pull/1680

core-rust-bridge/src/main/java/com/radixdlt/protocol/ProtocolUpdateEnactmentCondition.java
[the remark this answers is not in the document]
talekhinezh: Ah, what I mean is, does a threshold pass even if we're passed the maxEpoch and we're just waiting for number of consecutive epochs?

core-rust-bridge/src/main/java/com/radixdlt/statecomputer/ProtocolState.java
```diff
@@ -0,0 +1,121 @@
...
+            EnumCodec.fromPermittedRecordSubclasses(UnenactedProtocolUpdateState.class, codecs));
+
+    codecMap.register(
+        UnenactedProtocolUpdateState.SignalledReadinessThresholdState.class,
+        codecs ->
+            StructCodec.fromRecordComponents(
+                UnenactedProtocolUpdateState.SignalledReadinessThresholdState.class, codecs));
+  }
+
+  public record UnenactedProtocolUpdate(
```
talekhinezh, line 105: I'm not quite following this structure. Why the need for `UnenactedProtocolUpdateState` if `ProtocolUpdate` already has a condition?

core-rust-bridge/src/main/java/com/radixdlt/statecomputer/RustStateComputer.java
```diff
@@ -119,4 +130,22 @@ public Result<CommitSummary, InvalidCommitRequestError> commit(CommitRequest com
       commitFunc;
 
   private static native byte[] commit(NodeRustEnvironment nodeRustEnvironment, byte[] payload);
+
+  public String newestProtocolVersion() {
```
talekhinezh, line 134: Is this the newest pending or the current version?

core-rust-bridge/src/main/java/com/radixdlt/transaction/GetSyncableTxnsAndProofError.java
```diff
@@ -0,0 +1,87 @@
...
+
+public sealed interface GetSyncableTxnsAndProofError {
+  static void registerCodec(CodecMap codecMap) {
+    codecMap.register(
+        GetSyncableTxnsAndProofError.class,
+        codecs ->
+            EnumCodec.fromPermittedRecordSubclasses(GetSyncableTxnsAndProofError.class, codecs));
+  }
+
+  record RefusedToServeGenesis(LedgerProof refusedProof) implements GetSyncableTxnsAndProofError {}
```
talekhinezh, line 79: "Refused" sounds like a network client is not responding but this sounds like it should be more of a database error?

core-rust/state-manager/src/state_manager.rs
```diff
@@ -127,31 +129,42 @@ impl StateManagerConfig {
...
     pub database: Arc<StateLock<StateManagerDatabase>>,
     pub pending_transaction_result_cache: Arc<RwLock<PendingTransactionResultCache>>,
     pub mempool: Arc<RwLock<PriorityMempool>>,
     pub mempool_manager: Arc<MempoolManager>,
-    pub committability_validator: Arc<CommittabilityValidator<StateManagerDatabase>>,
-    pub transaction_previewer: Arc<TransactionPreviewer<StateManagerDatabase>>,
+    pub execution_configurator: Arc<RwLock<ExecutionConfigurator>>,
+    pub committability_validator: Arc<RwLock<CommittabilityValidator<StateManagerDatabase>>>,
+    pub transaction_previewer: Arc<RwLock<TransactionPreviewer<StateManagerDatabase>>>,
+    pub protocol_updater_factory: Arc<dyn ProtocolUpdaterFactory + Sync + Send>,
```
talekhinezh, line 149: It doesn't seem like this needs to be `dyn`, is it possible to just use a generic?

core-rust/state-manager/src/store/proofs_gc.rs
```diff
@@ -175,60 +181,346 @@ impl LedgerProofsGc {
...
-
-            let Some((_transactions, proof)) = transactions_and_proof else {
-                error!(
-                    "A chain of transactions-without-proof from state version {} does not fit within the limits {:?}; aborting the GC",
-                    batch_start_state_version_inclusive, self.limits_config
-                );
-                return;
+            let Some(next_prune_range) = self.locate_next_prune_range(last_pruned_state_version)
+            else {
+                break;
```
jakrawcz-rdx, line 189: is it an opportunity to use that cool Rust idiom `while let Some(x) = potentially_get_next_x() {` ?

core-rust/state-manager/src/jni/node_rust_environment.rs
```diff
@@ -141,10 +146,18 @@ impl JNINodeRustEnvironment {
             .track_running_tasks()
             .measured(metric_registry.deref());
 
+        let protocol_updater_factory: Box<dyn ProtocolUpdaterFactory + Send + Sync> =
```
talekhinezh, line 149: Probably cleaner if we are more explicit about what `ProtocolUpdaterFactory` is and implement it as an enum rather than dynamic trait.

core-rust/state-manager/src/jni/transaction_store.rs
```diff
@@ -114,6 +115,82 @@ extern "system" fn Java_com_radixdlt_transaction_REv2TransactionAndProofStore_ge
...
+        |_no_args: ()| -> Option<LedgerProof> {
+            let database = JNINodeRustEnvironment::get_database(&env, j_rust_global_context);
+            let proof = database.read_current().get_latest_epoch_proof();
+            proof
+        },
+    )
+}
+
+#[no_mangle]
+extern "system" fn Java_com_radixdlt_transaction_REv2TransactionAndProofStore_getLatestProtocolUpdateInitProof(
```
talekhinezh, line 145: Could this just be called `latestProtocolUpdateProof`?

core-rust/state-manager/src/store/proofs_gc.rs
```diff
@@ -130,7 +135,8 @@ impl LedgerProofsGc {
 
     /// Performs a single GC run, which is supposed to permanently delete *all* non-critical ledger
     /// proofs of configured old-enough epochs.
-    pub fn run(&self) {
+    /// Returns proof ranges that have been pruned.
```
jakrawcz-rdx, line 138: AFAIS, this is only needed for a test assert.

If so, then I suggest to unit-test some extracted lower-level helper - even the existing `fn locate_next_prune_range()` might be enough?

Alternatively, if you prefer a higher-level test, then how about asserting that expected ledger proofs were really deleted? (like, `db.get_proof_by_state_version(x) == None` after `.run()`)
_(after all, you never know whether `.run()` impl deletes exactly the ranges it returns, and has no off-by-1 errors etc. :p)_

core-rust/state-manager/src/protocol/mainnet_updates/anemone.rs
```diff
@@ -0,0 +1,70 @@
...
+use node_common::locks::StateLock;
+use radix_engine::prelude::dec;
+use radix_engine::utils::generate_seconds_precision_state_updates;
+use radix_engine_common::prelude::NetworkDefinition;
+use std::ops::Deref;
+use std::sync::Arc;
+
+pub struct AnemoneProtocolUpdater {
+    pub network: NetworkDefinition,
+    pub store: Arc<StateLock<StateManagerDatabase>>,
```
talekhinezh, line 16: Feels like to me that an "updater" should not have direct access to some store as it becomes stateful. Rather the store should be passed in to the execute or update function.

core-rust/state-manager/src/protocol/protocol_update.rs
```diff
@@ -0,0 +1,489 @@
...
+}
+
+/// Protocol update consists of two events:
+/// 1) Updating the current (state computer) configuration ("transaction processing rules").
+///    This includes: transaction validation, execution configuration, etc
+/// 2) Executing arbitrary state updates against the current database state.
+///    While the abstraction is quite flexible, the only concrete implementation at the moment
+///    only modifies the state through committing system transactions (e.g. substate flash).
+pub trait ProtocolUpdater {
+    fn protocol_version_name(&self) -> String;
```
talekhinezh, line 47: This name could probably be an associated constants

core-rust/state-manager/src/protocol/protocol_update.rs
```diff
@@ -0,0 +1,489 @@
...
+        }
+        ExecutionConfigurator::new(&self.network, &self.logging_config, costing_parameters)
+    }
+}
+
+pub trait StateUpdateExecutor {
+    /// Executes these state updates associated with the given protocol version
+    /// that haven't yet been applied
+    /// (hence "remaining", e.g. if node is restarted mid-protocol update).
+    fn execute_remaining_state_updates(&self);
```
talekhinezh, line 101: Feels a bit of code smell that the "remaining logic" part is hidden behind a trait rather than implemeted as explicit logic.

core-rust/state-manager/src/protocol/protocol_state.rs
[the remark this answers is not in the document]
talekhinezh: Cool, I like `applied_protocol_updates` and `pending_protocol_updates`

core-rust/state-manager/src/protocol/mainnet_updates/anemone.rs
```diff
@@ -0,0 +1,70 @@
...
+            self.state_computer_configurator.clone(),
+        );
+
+        while let Some(next_batch_idx) = txn_committer.next_committable_batch_idx() {
+            match next_batch_idx {
+                0 => {
+                    // Batch 0: flash consensus manager config update
+                    let mut config = self.store.read_current().get_consensus_manager_config();
+                    config.validator_creation_usd_cost = dec!("100");
+                    txn_committer.commit_flash(consensus_manager_config_flash(config));
```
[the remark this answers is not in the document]
talekhinezh: yup exactly, this is already on radixdlt-scrypto `develop` branch

core-rust-bridge/src/main/java/com/radixdlt/transaction/GetSyncableTxnsAndProofError.java
LukasGasior1: hmm I think it conveys the message quite well:
"You have asked to sync the genesis/protocol update transactions. I have them, but refuse to return because I believe you shouldn't need to sync them."
It's not an error

core-rust/state-manager/src/protocol/protocol_update.rs
talekhinezh: Furthermore, it's weird that we both have a mapping from protocol_version_name to `ProtocolUpdater`, and the `ProtocolUpdater` knowing it's own name. Perhaps minor but we could probably get away with the the Updater not knowing it's name.

core-rust/state-manager/src/protocol/protocol_update.rs
```diff
@@ -0,0 +1,489 @@
...
+
+/// Protocol update consists of two events:
+/// 1) Updating the current (state computer) configuration ("transaction processing rules").
+///    This includes: transaction validation, execution configuration, etc
+/// 2) Executing arbitrary state updates against the current database state.
+///    While the abstraction is quite flexible, the only concrete implementation at the moment
+///    only modifies the state through committing system transactions (e.g. substate flash).
+pub trait ProtocolUpdater {
+    fn protocol_version_name(&self) -> String;
+    fn state_computer_configurator(&self) -> StateComputerConfigurator;
```
talekhinezh, line 48: This could also be an associated constant to make it clear that this shouldn't change.

core-rust/state-manager/src/protocol/protocol_update.rs
```diff
@@ -0,0 +1,489 @@
...
+    LoggingConfig, ProtocolState, StateManagerDatabase,
+};
+
+pub trait ProtocolUpdaterFactory {
+    fn supports_protocol_version(&self, protocol_version_name: &str) -> bool;
+
+    fn updater_for(
+        &self,
+        protocol_version_name: &str,
+        store: Arc<StateLock<StateManagerDatabase>>,
```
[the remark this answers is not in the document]
talekhinezh: Agreed and relates to my comment that the `Updater` should not have a stateful store in it's struct but that it should rather be a stateless/immutable object.

core-rust/state-manager/src/protocol/protocol_config.rs
```diff
@@ -0,0 +1,178 @@
...
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
+pub struct ProtocolConfig {
+    pub genesis_protocol_version: String,
+    pub protocol_updates: Vec<ProtocolUpdate>,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
+pub enum ProtocolUpdateEnactmentCondition {
```
talekhinezh, line 55: Perhaps just `ProtocolUpdateCondition`?

core-rust/state-manager/src/protocol/protocol_update.rs
```diff
@@ -0,0 +1,489 @@
...
+///    While the abstraction is quite flexible, the only concrete implementation at the moment
+///    only modifies the state through committing system transactions (e.g. substate flash).
+pub trait ProtocolUpdater {
+    fn protocol_version_name(&self) -> String;
+    fn state_computer_configurator(&self) -> StateComputerConfigurator;
+    fn state_update_executor(&self) -> Box<dyn StateUpdateExecutor>;
+}
+
+#[derive(Clone, Debug)]
+pub struct StateComputerConfigurator {
```
talekhinezh, line 53: How about `StateComputerConfig` or `StateComputerConfiguration`

core-rust/state-manager/src/protocol/protocol_update.rs
```diff
@@ -0,0 +1,489 @@
...
+
+impl StateUpdateExecutor for NoOpStateUpdateExecutor {
+    fn execute_remaining_state_updates(&self) {
+        // No-op
+    }
+}
+
+pub struct FixedFlashProtocolUpdater {
+    protocol_version_name: String,
+    store: Arc<StateLock<StateManagerDatabase>>,
```
talekhinezh, line 144: similarly feels wrong to have this `store` in an Updater, rather it should be passed in into the `update` method

core-rust/state-manager/src/state_computer.rs
```diff
@@ -113,40 +114,57 @@ pub struct StateComputer<S> {
...
         vertex_limits_config: VertexLimitsConfig,
         store: Arc<StateLock<S>>,
         mempool_manager: Arc<MempoolManager>,
-        execution_configurator: Arc<ExecutionConfigurator>,
+        execution_configurator: Arc<RwLock<ExecutionConfigurator>>,
         pending_transaction_result_cache: Arc<RwLock<PendingTransactionResultCache>>,
         logging_config: LoggingConfig,
         metrics_registry: &Registry,
         lock_factory: LockFactory,
+        initial_protocol_updater: Box<dyn ProtocolUpdater>,
```
talekhinezh, line 150: I don't think you need dynamic trait for the `initial_protocol_updater` here

core-rust/state-manager/src/state_computer.rs
```diff
@@ -1210,12 +1271,55 @@ where
         self.committed_transactions_metrics
             .update(transactions_metrics_data);
 
+        let mut locked_protocol_state = self.protocol_state.write();
+        *locked_protocol_state = new_protocol_state;
+        drop(locked_protocol_state);
```
talekhinezh, line 1283: Rather than using `drop` better to just scope:
```
    {
        let mut locked_protocol_state = self.protocol_state.write();
        *locked_protocol_state = new_protocol_state;
    }
```

core-rust/state-manager/src/state_computer.rs
```diff
@@ -1210,12 +1271,55 @@ where
...
 
+    pub fn apply_protocol_update(
+        &self,
+        protocol_updater: &dyn ProtocolUpdater,
+    ) -> ProtocolUpdateResult {
+        let mut locked_ledger_transaction_validator = self.ledger_transaction_validator.write();
+        *locked_ledger_transaction_validator = protocol_updater
+            .state_computer_configurator()
+            .ledger_transaction_validator();
+        drop(locked_ledger_transaction_validator);
```
talekhinezh, line 1299: Using scoping rather than `drop` for all of these:
```
  {
        let mut locked_ledger_transaction_validator = self.ledger_transaction_validator.write();
        *locked_ledger_transaction_validator = protocol_updater
            .state_computer_configurator()
            .ledger_transaction_validator();
  }
```

core-rust/state-manager/src/state_computer.rs
```diff
@@ -1210,12 +1271,55 @@ where
...
         Ok(CommitSummary {
             validator_round_counters: round_counters,
             num_user_transactions,
         })
     }
 
+    pub fn apply_protocol_update(
+        &self,
+        protocol_updater: &dyn ProtocolUpdater,
+    ) -> ProtocolUpdateResult {
```
talekhinezh, line 1294: Minor: It feels a little weird that we need to pass in an Updater into a method of StateComputer. Perhaps better would be to only pass in the required state of StateComputer to Updater for updates?

core-rust/state-manager/src/state_computer.rs
```diff
@@ -1270,6 +1374,14 @@ where
             ),
             new_substate_node_ancestry_records: commit.new_substate_node_ancestry_records,
         });
+
+        // Protocol updates aren't allowed during genesis,
+        // so no need to handle an update here
+        // just assign the latest protocol state.
+        let mut locked_protocol_state = self.protocol_state.write();
+        *locked_protocol_state = series_executor.protocol_state();
+        drop(locked_protocol_state);
```
talekhinezh, line 1390: Also use scoping here

core-rust/state-manager/src/state_manager.rs
```diff
@@ -264,21 +297,66 @@ impl StateManager {
...
+            .protocol_updater_factory
+            .updater_for(protocol_version_name, self.database.clone());
+        let state_computer_configurator = protocol_updater.state_computer_configurator();
+
+        let new_execution_configurator =
+            state_computer_configurator.execution_configurator(self.config.no_fees);
+
+        let mut locked_execution_configurator = self.execution_configurator.write();
+        *locked_execution_configurator = new_execution_configurator;
+        drop(locked_execution_configurator);
```
talekhinezh, line 335: Use scoping for these drops :)

core-rust/state-manager/src/state_manager.rs
```diff
@@ -264,21 +297,66 @@ impl StateManager {
...
             mempool,
             mempool_manager,
+            execution_configurator,
             committability_validator,
             transaction_previewer,
+            protocol_updater_factory: protocol_updater_factory.into(),
         }
     }
+
+    pub fn apply_protocol_update(&self, protocol_version_name: &str) -> ProtocolUpdateResult {
```
talekhinezh, line 324: Is there a reason why this must be called from the outside? It seems like protocol updates should be in the realm of StateManager. But this is minor and we can refactor this later.

core-rust/state-manager/src/transaction/ledger_transaction.rs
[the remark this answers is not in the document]
talekhinezh: Yes and I believe there is a prepended version as well

core/src/main/java/com/radixdlt/consensus/ConsensusHasher.java
```diff
@@ -113,6 +113,10 @@ public static HashCode toHash(
         } else {
           outputStream.writeByte(0); // 1 byte
         }
+        if (committedHeader.nextProtocolVersion().isPresent()) {
```
talekhinezh, line 116: Can use `ifPresent`?

core/src/main/java/com/radixdlt/consensus/Vertex.java
```diff
@@ -260,17 +260,17 @@ public HashCode getParentVertexId() {
...
 
   public BFTHeader getGrandParentHeader() {
     return getQCToParent().getParentHeader();
   }
 
-  public boolean touchesGenesis() {
-    return this.getRound().isGenesis()
-        || this.getParentHeader().getRound().isGenesis()
-        || this.getGrandParentHeader().getRound().isGenesis();
+  public boolean touchesEpochInitialRound() {
```
talekhinezh, line 270: Can we do this renaming in another PR? This PR has gotten too big I think and having trouble finding the actual logic updates vs renames.

core-rust/state-manager/src/state_computer.rs
jakrawcz-rdx: I am still very new to Rust, so maybe I am not aware of all the consequences, but to me `drop` has advantages:
- [syntax] scope indents code
  - if you do a lot of it, your code visually moves too much to the right [just inconvenience]
  - also, my eyes kinda expect `if` or `for` whenever they see indented section, and a scope confuses them
- AFAIU, `drop` gives more control (e.g. over the order in which things are dropped)
- `drop` also seems very explicit: the intention is super-clear, someone wanted to free this resource at this very point (I mean: I understand that a scope also unambiguously does it, but feels more "silent", and I often witnessed an additional code comment explaining what the scope drops - while `drop` is so obvious that is does not need commenting)

core-rust/state-manager/src/protocol/protocol_state.rs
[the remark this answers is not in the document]
LukasGasior1: (renamed)

core-rust/state-manager/src/protocol/protocol_update.rs
LukasGasior1: Yeah, I agree

LukasGasior1: (removed)

core-rust/state-manager/src/protocol/protocol_update.rs
```diff
@@ -0,0 +1,489 @@
...
+    ExecutionConfigurator, FlashTransactionV1, LedgerTransaction, LedgerTransactionValidator,
+    PreparedLedgerTransaction, TransactionSeriesExecutor,
+};
+use crate::{
+    CommittedTransactionIdentifiers, ExecutionCache, LedgerHeader, LedgerProof, LedgerProofOrigin,
+    LoggingConfig, ProtocolState, StateManagerDatabase,
+};
+
+pub trait ProtocolUpdaterFactory {
+    fn supports_protocol_version(&self, protocol_version_name: &str) -> bool;
```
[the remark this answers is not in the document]
LukasGasior1: sounds good, they are lightweight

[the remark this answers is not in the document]
LukasGasior1: nah, I think it's actually uglier without this extra method because `updater_for` requires to provide a store (which isn't necessary if we just want to sanity check the updater factory)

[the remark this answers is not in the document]
jakrawcz-rdx: @LukasGasior1 but [some other comments](https://github.com/radixdlt/babylon-node/pull/790#discussion_r1453378751) (from me and @talekhinezh) already point out that `updater_for()` maybe should not really take `store` as param.

core-rust-bridge/src/main/java/com/radixdlt/statecomputer/ProtocolState.java
LukasGasior1: (as discussed separately) this also keeps the state of a condition

core-rust-bridge/src/main/java/com/radixdlt/statecomputer/RustStateComputer.java
LukasGasior1: newest pending (or enacted if there are no pending)

core-rust/state-manager/src/protocol/protocol_update.rs
LukasGasior1: agreed, removed this fn

core-rust/state-manager/src/protocol/protocol_update.rs
```diff
@@ -0,0 +1,489 @@
...
+/// Protocol update consists of two events:
+/// 1) Updating the current (state computer) configuration ("transaction processing rules").
+///    This includes: transaction validation, execution configuration, etc
+/// 2) Executing arbitrary state updates against the current database state.
+///    While the abstraction is quite flexible, the only concrete implementation at the moment
+///    only modifies the state through committing system transactions (e.g. substate flash).
+pub trait ProtocolUpdater {
+    fn protocol_version_name(&self) -> String;
+    fn state_computer_configurator(&self) -> StateComputerConfigurator;
+    fn state_update_executor(&self) -> Box<dyn StateUpdateExecutor>;
```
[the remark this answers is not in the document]
LukasGasior1: This originated from the following model:
Protocol update consists of:
a) changing the configuration
b) applying state changes
each action got its own component (StateComputerConfigurator, StateUpdateExecutor)
and ProtocolUpdater simply combines them.

Happy to flatten it out though.

core-rust/state-manager/src/protocol/protocol_config.rs
LukasGasior1: I don't have a strong opinion. What do others think? @jakrawcz-rdx @dhedey ?

jakrawcz-rdx: If you use "enactment, enact, enacted" widely throughout this feature, then `Enactment` part here is very relevant. And long names are actually our convention ;p
However, I've seen some discussions about other renames (like, "unenacted -> pending" [here](https://github.com/radixdlt/babylon-node/pull/790#discussion_r1452829020)). If you go this way, then I am for removing `Enactment` from here too.

core-rust/state-manager/src/store/traits.rs
```diff
@@ -323,8 +334,14 @@ pub mod proofs {
...
-        fn get_last_proof(&self) -> Option<LedgerProof>;
-        fn get_last_epoch_proof(&self) -> Option<LedgerProof>;
+        fn get_latest_proof(&self) -> Option<LedgerProof>;
+        fn get_latest_epoch_proof(&self) -> Option<LedgerProof>;
+        fn get_closest_epoch_proof_on_or_before(
+            &self,
+            state_version: StateVersion,
+        ) -> Option<LedgerProof>;
+        fn get_latest_protocol_update_init_proof(&self) -> Option<LedgerProof>;
+        fn get_latest_protocol_update_execution_proof(&self) -> Option<LedgerProof>;
```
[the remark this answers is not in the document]
LukasGasior1: `protocol_update_trigger_proof` / `protocol_update_local_proof` ?

core/src/main/java/com/radixdlt/consensus/ConsensusHasher.java
LukasGasior1: I'd have to handle `write`'s IOException, I think it's better as is

core-rust/state-manager/src/protocol/protocol_config.rs
dhedey: Slight preference for `ProtocolUpdateEnactmentCondition` but I don't mind really.

I think `protocol_update.enactment_condition` reads slightly better than `protocol_update.condition` but again, not a massive deal either way.

core/src/main/java/com/radixdlt/api/system/routes/HealthHandler.java
LukasGasior1: Done: I reused the LedgerUpdate event

talekhinezh (approved):

core-rust/state-manager/src/transaction/ledger_transaction.rs
```diff
@@ -130,9 +130,33 @@ const GENESIS_TRANSACTION_SYSTEM_TRANSACTION_DISCRIMINATOR: u8 = 1;
...
+                TransactionDiscriminator::V1Flash as u8,
+            ])
+            .update(state_updates_hash)
+            .update(self.nonce.to_le_bytes().as_slice())
+            .finalize();
+
+        Ok(PreparedFlashTransactionV1 {
+            state_updates: self.state_updates.clone(),
+            summary: Summary {
+                effective_length: 0,
```
dhedey, line 152: These two fields aren't correct... Probably doesn't matter a ton as I'm not sure they're used, BUT imo it's better to get rid of this method, move the content to `prepare_as_full_body_child`, and use:

```
let ((name, state_updates), summary) = ConcatenatedDigest::prepare_from_transaction_child_struct::<(
                SummarizedRawFullBody<String>,
                SummarizedRawFullBody<StateUpdates>,
            )>(decoder, TransactionDiscriminator::V1Flash)?;
```

It's more efficient, consistent, and the summary is more accurate.

See e.g. the implementation for SystemTransaction:
```
impl TransactionFullChildPreparable for PreparedSystemTransactionV1 {
    fn prepare_as_full_body_child(decoder: &mut TransactionDecoder) -> Result<Self, PrepareError> {
        let ((prepared_instructions, blobs, pre_allocated_addresses, hash_for_execution), summary) =
            ConcatenatedDigest::prepare_from_transaction_child_struct::<(
                PreparedInstructionsV1,
                PreparedBlobsV1,
                PreparedPreAllocatedAddresses,
                PreparedHash,
            )>(decoder, TransactionDiscriminator::V1System)?;
        Ok(Self {
            encoded_instructions: manifest_encode(&prepared_instructions.inner.0)?,
            references: prepared_instructions.references,
            blobs,
            pre_allocated_addresses,
            hash_for_execution,
            summary,
        })
    }
}
```

core-rust/state-manager/src/transaction/ledger_transaction.rs
```diff
@@ -130,9 +130,33 @@ const GENESIS_TRANSACTION_SYSTEM_TRANSACTION_DISCRIMINATOR: u8 = 1;
...
+impl FlashTransactionV1 {
+    pub fn prepare(&self) -> Result<PreparedFlashTransactionV1, PrepareError> {
+        let state_updates_hash = hash(scrypto_encode(&self.state_updates).unwrap());
+        let flash_hash = HashAccumulator::new()
+            .update([
+                TRANSACTION_HASHABLE_PAYLOAD_PREFIX,
+                TransactionDiscriminator::V1Flash as u8,
+            ])
+            .update(state_updates_hash)
+            .update(self.nonce.to_le_bytes().as_slice())
```
dhedey, line 146: It's weird to serialize for storage as little endian....

And actually I'd quite strongly rather we named the FlashTransaction rather than gave it a nonce. Would that be okay?

The name can then be used to describe the content of the transaction. Eg `anenome-pool-code-update` or whatever.

I guess we'd also need to adjust the Core API models.

dhedey (commented):
A few comments on `ledger_transaction.rs`

core-rust/state-manager/src/transaction/ledger_transaction.rs
dhedey: (although if we want to add a nonce of the state version too to be extra sure (as well as the name) we can?) - if we're worried the commit step wouldn't detect the duplicate hash?

core-rust/state-manager/src/protocol/protocol_config.rs
```diff
@@ -0,0 +1,178 @@
...
+    EnactWhenSupportedAndWithinBounds {
+        lower_bound: Epoch,
+        upper_bound: Epoch,
```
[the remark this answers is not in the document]
LukasGasior1: > This might be a little clearer as EnactAtStartOfAnEpochIfConditionsMet ?

I personally like it less specific, it leaves the doors open for "at state version" updates (by extending the bounds to `enum ProtocolUpdateBound { Epoch(..), StateVersion(..) }`.

I'll add a comment :+1: both are actually inclusive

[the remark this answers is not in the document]
jakrawcz-rdx: > both are actually inclusive

@LukasGasior1 just to avoid confusion, I'd vote to do the classic "from inclusive, to exclusive". The use of `Range<>` would achieve it while boosting the OOP score even further :p

Alternatively, if it conceptually makes LOTS of sense to operate on "to inclusive", then the best names would be `min, max` (since "max" hints "maximum possible" or "maximum that is still allowed", and won't be confused with "exclusive").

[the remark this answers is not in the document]
dhedey: > I personally like it less specific, it leaves the doors open for "at state version" updates (by extending the bounds to enum ProtocolUpdateBound { Epoch(..), StateVersion(..) }.

I disagree - when the enactment is checked is key to the condition. Having it enact only at a start of an epoch is different to having it enact at any particular state version.

I assume this condition can only be used to enact at start of epoch, right? Like we can't have this thing trigger mid-epoch?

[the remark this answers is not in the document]
jakrawcz-rdx: > Like we can't have this thing trigger mid-epoch?

I personally hope we will never have the "trigger protocol update mid-epoch" feature. I'd prefer to have a way of forcing a premature epoch change (with the Engine's cooperation, which Josh confirmed seems possible).

core-rust/state-manager/src/transaction/ledger_transaction.rs
LukasGasior1: I'm okay with named transactions, will change.

Commit should detect duplicates :+1:

LukasGasior1: (I'll address this in: https://github.com/radixdlt/babylon-node/pull/803)

core-rust/state-manager/src/transaction/ledger_transaction.rs
LukasGasior1: :+1: done in https://github.com/radixdlt/babylon-node/pull/803

core-rust/state-manager/src/protocol/protocol_config.rs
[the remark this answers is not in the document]
LukasGasior1: >  assume this condition can only be used to enact at start of epoch, right? Like we can't have this thing trigger mid-epoch?

Currently yes, but with state version updates the same "enact when supported" condition could be reused. One part of the condition would be met at the beginning of an epoch (the support sub-condition) and the other part possibly mid-epoch (min/max bound sub-condition).

Anyway, happy to rename that for now and revisit if/when we actually introduce other conditions.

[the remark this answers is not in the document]
LukasGasior1: but just for the sake of discussion :smile: 

> I disagree - when the enactment is checked is key to the condition. Having it enact only at a start of an epoch is different to having it enact at any particular state version.

I don't think when the enactment is checked is key. I treat the condition as being continously evaluated: if at *any* time (i.e. after any txn) all its components are met, then it passes. It just so happens that currently the state that influences the evaluation changes only at epoch boundaries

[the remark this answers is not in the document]
LukasGasior1: (as discussed, will rename in a follow up PR)

PR #794, opened 2024-01-13
Architecture: Update Scrypto Lib

talekhinezh:
* Update Scrypto Lib
* Add `VMBoot` substate to API


core-rust/state-manager/src/transaction/executable_logic.rs
```diff
@@ -197,7 +197,7 @@ impl<'a, S: SubstateDatabase> TransactionLogic<S> for ConfiguredExecutable<'a> {
                 substate_schema_mapper.add_for_all_individually_updated(&state_updates);
                 let substate_system_structures = substate_schema_mapper.done();
 
-                // Sanity check that all updates are to existing nodes so that
+                // Sanity check that all updates are to existing entities or bootstrap so that
```
jakrawcz-rdx, line 200: unintended change?

jakrawcz-rdx (commented):
This is surprisingly clean, at least from Node's perspective.

BTW. I noticed a lot of formatting/style changes, and I assume they all come from `spotlessApply`, and I figured it's time to finally update my toolchain :D
Let's merge my https://github.com/radixdlt/babylon-node/pull/798 and then this PR will become even clearer?

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -2016,6 +2016,7 @@ components:
     EntityModule:
       type: string
       enum:
+        - BootLoader
```
dhedey, line 2019: Very minor, but we should add this to the bottom so it aligns with the partition ordering. We should also update https://radixdlt.atlassian.net/wiki/spaces/S/pages/2979430761/REP-60+Persisted+State+Model - which number did we choose for boot loader partition? 0x10 to leave room for some more modules?

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -3895,6 +3899,25 @@ components:
...
+    BootLoaderModuleFieldVmBootSubstate:
+      allOf:
+        - $ref: '#/components/schemas/Substate'
+        - type: object
+          required:
+            - value
+          properties:
+            value:
+              $ref: '#/components/schemas/VmBootSubstate'
+    VmBootSubstate:
```
dhedey, line 3911: This doesn't follow the naming convention. This should be `BootLoaderModuleFieldVmBootValue`

core-rust/core-api-server/src/core_api/conversions/substates/type_info_module.rs
```diff
@@ -3,8 +3,26 @@ use super::*;
...
 use radix_engine::types::*;
+use radix_engine::vm::VmBoot;
 use radix_engine_queries::typed_substate_layout::*;
 
+pub fn to_api_vm_boot_substate(
+    _context: &MappingContext,
+    _state_mapping_lookups: &StateMappingLookups,
+    substate: &VmBoot,
+) -> Result<models::Substate, MappingError> {
+    let substate = match substate {
```
dhedey, line 14: Very minor/picky, but for consistency, this is referred to as `value` rather than the substate.

For consistency, you could also create a `locked_versioned_system_field_substate` macro, but that's possibly overkill. The macro does prevent inconsistent naming though.

dhedey (approved):
A few mark-ups from Jakub and I, then happy for you to merge. On my end, only the `BootLoaderModuleFieldVmBootValue` rename is mandatory, the other tweaks are optional.

core-rust/state-manager/src/transaction/executable_logic.rs
talekhinezh: yup good catch :)

core-rust/core-api-server/core-api-schema.yaml
talekhinezh: Good point, 0x20 (32) was chosen to leave room for more modules

PR #795, opened 2024-01-15
Split off Browse API from Core API

jakrawcz-rdx:
☺️ Please do not fear the PR size: it is caused by moving very large parts of OpenApi model spec to a separate YAML file.

This PR addresses https://radixdlt.atlassian.net/browse/NODE-578 :
- Renames the project `core-api-server` to `node-http-apis`, where both servers now live,
- Splits `browse-api-schema.yaml` from `core-api-schema.yaml`,
- Adjusts the OpenApi generator .py script,
- Fixes the generic aspects of Browse API's impl to follow [the new API conventions](https://radixdlt.atlassian.net/wiki/spaces/RPNV1/pages/3236986940/Umbrella+2nd+Generation+API+Consistency): most notably error-handling


jakrawcz-rdx:
The last commit fixes the hotlinked logo url (in all places).

For reference, now the local `$ npx @redocly/cli build-docs core-rust/node-http-apis/browse-api-schema.yaml` looks like this:
image

jakrawcz-rdx:
I suspect the long-lived feature branch will be easier to review with this than without. Lemme merge there.

PR #801, opened 2024-01-18
Simplifying the paging specification (and paging policy impl)

jakrawcz-rdx:
As discussed with @les-sheppard:

The page size driven by request's handling (wallclock) time was nice to the DB lock, but confusing to consumers.
We prefer to have a lower limit (i.e. 100, which should never hold the DB lock for too long) but simpler rules (i.e. "if you don't specify any page size, you get our max allowed, 100").


jakrawcz-rdx:
let me merge down all the stacked changes to the long-lived branch (I want to pull `develop` there)

PR #802, opened 2024-01-18
Add readiness signal status to health API

LukasGasior1:


talekhinezh (approved):

core/src/main/java/com/radixdlt/api/system/health/HealthInfoServiceImpl.java
```diff
@@ -106,4 +122,46 @@ public RecentSelfProposalMissStatistic recentSelfProposalMissStatistic() {
...
   }
+
+  @Override
+  public HealthResponse.ReadinessSignalStatusEnum readinessSignalStatus() {
+    return this.cachedReadinessSignalStatus;
+  }
+
+  @Override
+  public EventProcessor<HealthReadinessSignalStatusUpdateTrigger>
+      readinessSignalStatusUpdateTriggerEventProcessor() {
```
jakrawcz-rdx, line 133: maybe I don't get the full motivation here, but... why the `{trigger(10s) + processor(read DB) + read from cache on request}`, instead of a simple `read DB on request`?

core/src/main/java/com/radixdlt/api/system/health/HealthInfoServiceImpl.java
```diff
@@ -106,4 +122,46 @@ public RecentSelfProposalMissStatistic recentSelfProposalMissStatistic() {
...
+    return unusedEvent ->
+        selfValidatorInfo
+            .bftValidatorId()
+            .ifPresent(
+                selfValidatorId -> {
+                  final var protocolState = inMemorySystemInfo.getProtocolState();
+                  final var signalRequired =
+                      protocolState.pendingProtocolUpdates().stream()
```
dhedey, line 141: It feels weird to me that this is per-protocol update, but then we flatten it to a single boolean, rather than per pending protocol update?

Perhaps instead we just cache our current vote? And then can compute the `readinessSignalStatus` for each pending update?

core/src/main/java/com/radixdlt/api/system/health/HealthInfoServiceImpl.java
dhedey: I kinda agree with Jakub, this feels a little like overkill? Out of interest, what was the push to do it this way?

core-rust/state-manager/src/jni/state_reader.rs
```diff
@@ -0,0 +1,110 @@
...
+    request_payload: jbyteArray,
+) -> jbyteArray {
+    jni_sbor_coded_fallible_call(
+        &env,
+        request_payload,
+        |validator_address: ComponentAddress| -> JavaResult<Option<String>> {
+            let result = JNINodeRustEnvironment::get(&env, j_node_rust_env)
+                .state_manager
+                .database
+                .read_current()
```
dhedey, line 99: We could consider using `access_non_locked_historical` to avoid locking if we were worried about locking on the health endpoint.

dhedey (commented):
A few misc thoughts

core/src/main/java/com/radixdlt/api/system/health/HealthInfoServiceImpl.java
LukasGasior1: just to avoid locking on health calls

LukasGasior1: hmh but maybe that's not really an issue if I use `access_non_locked_historical` as you suggest below

jakrawcz-rdx: yup, I suggest `access_non_locked_historical` for now.
Later we will read it from a DB snapshot.

We have far too many DB-reading endpoints already to care about such short-lived lock.
If you want to be super-careful (due to the public nature of /health endpoint), you can wrap it in a simple Guava cache (e.g. expire 10s after read). It would lock less often than the current solution.

core/src/main/java/com/radixdlt/api/system/health/HealthInfoServiceImpl.java
LukasGasior1: I wanted to flatten it for the end user to a just a single status (VOTE_REQUIRED if there exists at least one pending protocol update that requires a vote)

core/src/main/java/com/radixdlt/api/system/health/HealthInfoServiceImpl.java
LukasGasior1: (done)

core/src/main/java/com/radixdlt/api/system/health/HealthInfoServiceImpl.java
dhedey: Hmm but if there is more than one protocol update it can never be true? Like you either be ready for one or the other? So it would always be VOTE_REQUIRED.

Also, VOTE_REQUIRED is old terminology, we should use the term readiness signalled.

So it doesn't seem to make sense really.

I think it makes sense instead to have a `readinessSignalStatus` for each pending protocol update, and either `NO_SIGNAL_REQUIRED` (if you're not a validator or the protocol update doesn't need a signal) or `READINESS_NOT_SIGNALLED` or `READINESS_SIGNALLED`.

LukasGasior1: (done)

jakrawcz-rdx (approved):

PR #805, opened 2024-01-19
DO-2037 / use selfhosted runners for ci.yml

kofrdx:
## Description

Use self-hosted EC2 runners for all workflows in the ci.yml.
This aims to reduce costs and give appropriate resource requests to the individual workflows.
Compared to the current self-hosted k8s runners we should see some benefits

Pro k8s | Con k8s
-- | --
Multi jobs on 1 machine | Mixed resources. Random oom-kill
Relatively low waiting time before job gets picked up | 10min idle time before shutdown. Wasted money
  | Runs in docker, extra abstraction, might slow down, changes available tooling
  | Same runner type for each job. Same request/limit. Inefficient utilisation. Often less than 20%

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ on:
 jobs:
   snyk-scan-deps-licences:
     name: Snyk deps/licences scan
-    runs-on: ubuntu-latest
+    runs-on: selfhosted-ubuntu-22.04
     permissions:
       id-token: write
       pull-requests: read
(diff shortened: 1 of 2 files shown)


stringhandler:
Be aware of this when using self-hosted runners https://johnstawinski.com/2024/01/05/worse-than-solarwinds-three-steps-to-hack-blockchains-github-and-ml-through-github-actions/

kofrdx:
These workflows have a significant time in high CPU usage, so I tried executing them with the 16 CPU runners and noticed a runtime improvement. 

- Targeted integration tests
- Test core-rust docker build for local development
- Cross compile to Windows


Screenshot 2024-01-19 at 13 40 51

marek-karwacki-rdx (approved):

shambupujar (approved):

PR #807, opened 2024-01-19
Tweak/core api well known addresses

dhedey:
Add transaction_tracker well known address and regenerate all APIs using `./regenerate-apis.sh` (including Typescript SDK which looks like was missed).


jakrawcz-rdx (approved):

jakrawcz-rdx:
huh, that's funny:

image
(https://github.com/radixdlt/babylon-node/actions/runs/7585835903/job/20662592619?pr=807#step:9:523)

PR #809, opened 2024-01-19
Make protocol update upper bound exclusive; Only update the protocol state at epoch change

LukasGasior1:


core-rust/state-manager/src/protocol/protocol_config.rs
```diff
@@ -53,16 +53,20 @@ pub struct ProtocolConfig {
 
 #[derive(Clone, Debug, Eq, PartialEq, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
 pub enum ProtocolUpdateEnactmentCondition {
-    EnactWhenSupportedAndWithinBounds {
+    /// The enactment only proceeds if it's the start of epoch X,
+    /// at least one readiness threshold is met, and X satisfies
+    /// `lower_bound_inclusive <= X < upper_bound_exclusive`.
+    EnactAtStartOfAnEpochIfSupportedAndWithinBounds {
```
LukasGasior1, line 59: it's quite long, but hopefully still acceptable? :smile:

dhedey: :D 

Will change these to `EnactAtStartOfEpochConditionally` and `EnactAtStartOfEpochUnconditionally` in my follow up PR - what do you think?

dhedey (approved):

core-rust/state-manager/src/protocol/protocol_config.rs
LukasGasior1: happy with either of these :+1:

PR #810, opened 2024-01-22
Adding substate and partition delete support to FlashLedgerTransaction in Core API

jakrawcz-rdx:
An omission reported at https://rdxworks.slack.com/archives/C01M90Y2448/p1705913560366799

Intuitively, the `FlashTransaction` could be represented as `StateUpdates`.
However, our existing `StateUpdates` Core API output is very receipt-focused (e.g. it includes previous value for substate updates) and it is hard to render for `FlashTransaction`. So I had to re-create a similar structure.


core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -7413,10 +7413,10 @@ components:
         - $ref: "#/components/schemas/LedgerTransaction"
         - type: object
           required:
-            - flash_transaction
+            - flashed_state_updates
```
jakrawcz-rdx, line 7416: _(note from author)_

A flash transaction, apart from state updates, also has a `name`.
Do we want to include it in the Core API response?

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -1360,7 +1360,6 @@ impl SubstateDatabase for RocksDBStore {
             self.open_db_context()
                 .cf(SubstatesCf)
                 .iterate_group_from(&(partition_key.clone(), from_sort_key), Direction::Forward)
-                .take_while(move |((next_key, _), _)| next_key == &partition_key)
```
jakrawcz-rdx, line 1363: _(note from author)_

This is just a drive-by mentioned at https://rdxworks.slack.com/archives/D04CPLNN81Y/p1705591790950219

core-rust/core-api-server/core-api-schema.yaml
dhedey: Yes please :)

jakrawcz-rdx: (already done - see `- name` one line above)

PR #816, opened 2024-01-23
refactor: Improved configuration of protocol updates

dhedey:
## Summary

Basically it's a big refactor of the rust side of the code, roughly:

* I've tried to re-organise things into clearer files/folders.
* I've done some renames and some slight reworkings to clarify certain concepts:
  * The `ProtocolUpdateFactory` trait has become a single `ProtocolDefinitionResolver` implementation. This means we actually test our production code rather than using various stubs which seem to miss some of the key parts of the implementation. I think it's also easier to follow.
  * `ProtocolUpdate` has become `ProtocolUpdateTrigger`
  * `UpgradableStateComputerConfig` has become `ProtocolStateComputerConfig`
  * I've introduced a `ProtocolUpdateDefinition` which gives access to a static config and the ability to create a (boxed) `ProtocolUpdater` (which is no longer responsible for defining the static config).
* I've added some more comments
* I've hard coded `babylon-genesis` as the defacto standard genesis version everywhere. Attempting to use something else will fail. This reduces the potential for issues. (although still allows us to define a different genesis name with a different config if we want!)

In terms of features/useful changes:
* Different network protocol configs now live in rust, and testnets now attempt to run anemone at the start of epoch 3.
* `ProtocolConfig` now has a field to define some overrides for protocol updating (which depends on the ProtocolUpdater). Don't worry test team - we try to deserialize the old model so it shouldn't break your tests. We should probably just make it versioned from some point to make the code a little less hacky.
* Any protocol updates starting `test-` will inject a single empty test transaction.
* Any protocol updates starting `custom-` will read flash transactions from the raw overrides to inject; else will inject nothing - as requested by @les-sheppard. I've pivoted across `test.rs` (which should probably be renamed).


core-rust-bridge/src/main/java/com/radixdlt/protocol/ProtocolConfigV0.java
```diff
@@ -0,0 +1,86 @@
...
+
+package com.radixdlt.protocol;
+
+import com.google.common.collect.ImmutableList;
+import com.radixdlt.sbor.codec.CodecMap;
+import com.radixdlt.sbor.codec.StructCodec;
+import java.util.Map;
+
+/**
+ * A legacy format still used in some tests. We attempt to decode this as well as the new format.
```
LukasGasior1, line 73: lol we've got a legacy format even before the feature is released :laughing:

core-rust/state-manager/src/protocol/protocol_updates/definitions/test_definition.rs
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    fn state_computer_config(
+        network_definition: &NetworkDefinition,
+    ) -> ProtocolStateComputerConfig {
+        ProtocolStateComputerConfig::default(network_definition.clone())
+    }
+}
+
+struct TestBatchGenerator {
```
jakrawcz-rdx, line 44: the `ArbitraryBatchGenerator` with predefined contents could do this job as well?

core-rust/state-manager/src/protocol/protocol_updates/protocol_update_definition.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    /// Must return `Ok` if the `raw_config` is None or passed validate.
+    fn create_updater_with_raw_overrides(
+        &self,
+        new_protocol_version: &str,
+        network_definition: &NetworkDefinition,
+        raw_config: Option<&[u8]>,
+    ) -> Result<Box<dyn ProtocolUpdater>, DecodeError>;
+
+    fn validate_raw_overrides(&self, raw_config: &[u8]) -> Result<(), DecodeError>;
```
jakrawcz-rdx, line 47: (just a hunch)

shouldn't it be `fn validate(raw) -> ParsedValidOverrides`?
wouldn't it make `create_updater_with_raw_overrides()` infallible?

alternatively: maybe this method does not really belong in the trait?

core-rust-bridge/src/main/java/com/radixdlt/protocol/ProtocolConfig.java
```diff
@@ -65,22 +65,52 @@
...
   public static void registerCodec(CodecMap codecMap) {
     codecMap.register(
         ProtocolConfig.class,
         codecs -> StructCodec.fromRecordComponents(ProtocolConfig.class, codecs));
   }
 
+  public static ProtocolConfig sborDecodeWithFallbackForOldVersions(byte[] encoded) {
+    try {
+      return NodeSborCodecs.decode(encoded, NodeSborCodecs.resolveCodec(new TypeToken<>() {}));
+    } catch (SborDecodeException ex) {
```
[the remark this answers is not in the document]
dhedey: > What if it deserializes [by chance] and just has a different meaning?
It can't - the ProtocolConfig structs have different lengths

> Do we need the backwards compatibility that desperately?
I put this in as a help to the test team because they've already generated about 20 of these, and it would likely be a pain to regenerate this week.

That said, if you feel strongly we can talk to the test team and see if they're comfortable to change; and going forwards, we could make ProtocolConfig a versioned enum.

core-rust-bridge/src/test/java/com/radixdlt/rev2/ProtocolConfigGeneratorTest.java
```diff
@@ -84,7 +84,7 @@ public void generateProtocolConfig() {
         new ProtocolConfig(
             "babylon-genesis",
```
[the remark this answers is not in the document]
dhedey: Fixed, sorry.

core-rust/state-manager/src/protocol/protocol_config.rs
```diff
@@ -1,62 +1,207 @@
...
     }
 }
 
-#[derive(Clone, Debug, Eq, PartialEq, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
-pub struct ProtocolConfig {
-    pub genesis_protocol_version: String,
-    pub protocol_updates: Vec<ProtocolUpdate>,
+/// Returns an ASCII protocol version name truncated/left padded to canonical length (16 bytes)
+pub fn ascii_protocol_version_name_len_16(protocol_version_name: &str) -> String {
+    let filtered_version_name: String = protocol_version_name
```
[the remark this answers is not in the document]
dhedey: Yeah it looks like we already do validation on the thing that it's ascii, so will tighten that up and remove this code.

core-rust/state-manager/src/protocol/protocol_updates/protocol_update_definition.rs
dhedey: Well if the raw overrides are incorrect, we'd want to know at config time I think? Rather than parsing them incorrectly?

But maybe I'm not following you?

LukasGasior1 (approved):
Looks good!

core-rust/state-manager/src/protocol/protocol_updates/protocol_update_definition.rs
dhedey: Hmm I'm going to merge `ProtocolDefinitionResolver` into `ProtocolConfig` I think.

It just makes more sense I think having one place to validate everything.

jakrawcz-rdx: > we'd want to know at config time I think?

Yes,

I was just hoping that `Unwrap is allowed because we have already validated the raw config` can be avoided (by the power of type-safety, my favourite feature of all serious languages).
However, I played a bit with it right now, and I now think that either the `ConfigurableProtocolUpdateDefinition` is a bit artificial OR the `Overrides` associated type brings unneeded-flexibility (static `Vec<TxBatch>` instead?) at the price of excessive-complexity.

These aren't huge problems, but since this PR does some cool simplifications already, I only wanted to point out one more opportunity. Ofc `TODO: future refactor` is cool.

core-rust/state-manager/src/protocol/protocol_config.rs
```diff
@@ -1,62 +1,207 @@
...
+            ));
+        }
+
+        if !name.is_ascii() {
+            return Err(format!(
+                "Protocol version ({name}) can't use non-ascii characters"
+            ));
+        }
+
+        if !protocol_definition_resolver.recognizes(name) {
```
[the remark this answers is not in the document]
dhedey: I've reworked this all and ditched the resolver now. It all lives as part of `ProtocolConfig` and feels a lot less smelly now. But interested in your input  :)

core-rust/state-manager/src/protocol/protocol_updates/protocol_update_definition.rs
dhedey: I've tweaked it a little as you suggest (i.e. it's infallible, but has an explicit documented precondition that it will panic if the provided content is invalid)

A core issue here is that different protocol update definitions have different types for their config, and so we can't have a typed Overrides.

We could have a ValidatedRawOverrides, but this still wouldn't necessarily capture validated with respect to the particular protocol definition.

dhedey: The overrides will be used in quite a few places in future I think:
* If/when we migrate genesis to be a protocol update
* Any protocol updates which touch consensus manager constants; or otherwise might want to be configured differently in test environments
* The `custom-` protocol updates for testing

jakrawcz-rdx (approved):

core-rust/state-manager/src/protocol/protocol_config.rs
```diff
@@ -1,62 +1,310 @@
...
+        }
+        let passes_char_check = self.0.chars().all(|c| match c {
+            _ if c.is_ascii_alphanumeric() => true,
+            '_' | '-' => true,
+            _ => false,
+        });
+        if !passes_char_check {
+            return Err(ProtocolVersionNameValidationError::CharsInvalid {
+                invalid_name: self.0.clone(),
+                allowed_chars: "[A-Za-z0-9] and -".to_string(),
```
LukasGasior1, line 228: and `_` :)

core-rust/state-manager/src/protocol/protocol_config.rs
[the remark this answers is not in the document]
LukasGasior1: :+1:

PR #818, opened 2024-01-23
Returning protocol update readiness signal(s) in Core API's state endpoints

jakrawcz-rdx:
As requested at https://rdxworks.slack.com/archives/C01M90Y2448/p1705924939276209

For `/state/validator`: this simply outputs the raw field `Substate`, same as what happens for all other fields (we could say it was mistakenly omitted until now).
For `/state/consensus-manager`: this returns an "interpreted" state of each validator from active set (i.e. requires N substate fetches). And hence is a new "ugly mixed abstraction" part of the Core API. It is hidden behind a request flag.

### DIFF
--- a/core-rust/core-api-server/core-api-schema.yaml
+++ b/core-rust/core-api-server/core-api-schema.yaml
@@ -7527,6 +7527,9 @@ components:
       properties:
         network:
           $ref: "#/components/schemas/NetworkIdentifier"
+        include_readiness_signals:
+          description: Whether to include protocol update readiness signals of active validator set (default false).
+          type: boolean
     StateConsensusManagerResponse:
       type: object
       required:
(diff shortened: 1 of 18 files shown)


core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -7553,6 +7556,32 @@ components:
...
+          description: |
+            Protocol versions signalled by current Validator set.
+            Only returned if enabled by `include_readiness_signals` on your request.
+          type: array
+          items:
+            $ref: "#/components/schemas/ProtocolVersionReadiness"
+    ProtocolVersionReadiness:
+      type: object
+      required:
+        - signalled_protocol_version
```
dhedey, line 7569: In terms of our tracking, I think it would actually be helpful to include the validators who have no signalled protocol version, and also the stake proportion of each validator.

(As importantly, in a way, it matters more who _hasn't_ signalled yet than who has)

dhedey: Although thinking about this a little more, I quite like the grouping...

Okay, so conclusion I think - how about we keep the grouping, but:
* We include a grouping with empty signalled protocol version
* We include a `total_active_stake_proportion` on the `ProtocolVersionReadiness` (so across the groups, they should add to something very close to 1)
* We change `signalling_validator_addresses` to `signalling_validators` (of a `SignallingValidator` item) which includes the `validator_address` as well as an `active_stake_proportion` for that validator.

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -7553,6 +7556,32 @@ components:
           $ref: "#/components/schemas/Substate"
         current_time_rounded_to_minutes:
           $ref: "#/components/schemas/Substate"
+        current_validator_readiness_signals:
+          description: |
+            Protocol versions signalled by current Validator set.
```
dhedey, line 7561:
```suggestion
            Protocol versions signalled by the current validator set.
```

dhedey (approved):
Nice, thanks! Have a small suggestion for a hopefully easy tweak.

core-rust/core-api-server/core-api-schema.yaml
jakrawcz-rdx: hah, I thought about this, and my classic "don't return easily-computable fields" mental barrier kicked in...
(the same response contains a full active validator set, with their stakes, so everything you mentioned will be there)

But ofc we have to answer what is "easily-computable" here:
**A)** if we expect people to call this endpoint manually in order to quickly glance at the readiness field: then I agree to all your suggestions
**B)** if we expect this will be mostly used to render some dashboard: then I say "2 lines of code on the caller's side".

❓ 

_I ack that we are already doing the "convenience" response here (after all, the caller could loop through validators and do 100 more calls themselves) so I won't be surprised if you confirm **A)** above :)_

jakrawcz-rdx: (addressed according to **A)**)

PR #819, opened 2024-01-24
Add stream proofs api to Core API

dhedey:
-


core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -7072,6 +7144,184 @@ components:
...
+    StreamProofsFilterAll:
+      allOf:
+        - $ref: "#/components/schemas/StreamProofsFilter"
+        - type: object
+          properties:
+            from_state_version:
+              type: integer
+              format: int64
+              minimum: 0
+              maximum: 10000000000
```
krzlabrdx, line 7198: I believe `state_version` maxes out at `100000000000000` (applies to the `description` too)

krzlabrdx (changes requested):

core-rust/core-api-server/core-api-schema.yaml
dhedey: Fixed 👍

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -7072,6 +7144,175 @@ components:
...
+                If empty, it starts from 0.
+    StreamProofsFilterProtocolUpdateExecution:
+      allOf:
+        - $ref: "#/components/schemas/StreamProofsFilter"
+        - type: object
+          properties:
+            protocol_version:
+              type: string
+              description: |
+                The protocol version name to filter to. This can be returned from looking at the protocol updates.
```
jakrawcz-rdx, line 7231: I don't understand the "... returned from looking at ..." part.

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -7125,7 +7366,7 @@ components:
           format: int32
           minimum: 0
           maximum: 10000
-          description: An integer between `0` and `10000`, giving the total count of transactions in the returned response
+          description: An integer between `0` and `10000`, giving the total count of proofs in the returned response
```
jakrawcz-rdx, line 7369: AFAIS, this is a part of `StreamTransactionsResponse` and it rightfully mentioned `transactions` count.

core-rust/core-api-server/src/core_api/conversions/pagination.rs
```diff
@@ -0,0 +1,83 @@
...
+) -> Result<usize, ExtractionError> {
+    let SizeRange { min, default, max } = size_range;
+    match provided {
+        Some(provided) => {
+            if provided < 0 {
+                None
+            } else {
+                let provided = provided as usize;
+                if provided < min || provided > max {
+                    None
```
jakrawcz-rdx, line 23: `return Err()` from here would be the most readable option for me

core-rust/core-api-server/src/core_api/conversions/pagination.rs
```diff
@@ -0,0 +1,83 @@
...
+
+use crate::core_api::*;
+
+pub struct SizeRange {
+    pub min: usize,
+    pub default: usize,
+    pub max: usize,
+}
+
+pub fn extract_valid_size(
```
jakrawcz-rdx, line 11: is this secretly a `size_range.resolve_valid(Option<i32>) -> Result`?

core-rust/core-api-server/src/core_api/conversions/pagination.rs
```diff
@@ -0,0 +1,83 @@
...
+            }
+        }
+        None => Some(default),
+    }
+    .ok_or(ExtractionError::InvalidSize { min, max })
+}
+
+pub fn extract_continuation_token<T: ScryptoDecode>(
+    continuation_token: Option<String>,
+) -> Result<Option<T>, ExtractionError> {
```
jakrawcz-rdx, line 36: this kind of functions become more general-purpose when you drop both the input and output `Option<>`.
(i.e. `extract_continuation_token(String) -> Result<T, E>`)

it is typically easy enough for the caller to do `option.map(extract_continuation_token)` anyways.

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -7072,6 +7144,175 @@ components:
...
+      type: object
+      required:
+        - type
+      properties:
+        type:
+          $ref: "#/components/schemas/StreamProofsFilterType"
+      discriminator:
+        propertyName: type
+        mapping:
+          # NOTE: These need to match StreamProofsFilterType
```
LukasGasior1, line 7184: `ProtocolUpdateInitializations` doesn't match (StreamProofsFilterType is `ProtocolUpdateTriggers` )

core-rust/core-api-server/src/core_api/handlers/stream_proofs.rs
```diff
@@ -0,0 +1,211 @@
...
+    continue_from_state_version: Option<StateVersion>,
+    from_epoch: Epoch,
+) -> Result<
+    Box<dyn Iterator<Item = LedgerProof> + 'a>,
+    ResponseError<models::StreamProofsErrorDetails>,
+> {
+    let continuation_next_epoch = match continue_from_state_version {
+        Some(state_version) => match database.get_proof_iter(state_version).next() {
+            Some(proof) => Some(proof.ledger_header.epoch.next().unwrap()),
+            None => Err(client_error("continuation_token is not valid"))?,
```
jakrawcz-rdx, line 103: here - yes, this means malformed continuation token

but in general, it can happen to a valid token, when elements are deleted from the end of the collection. So maybe an empty return is more appropriate than error?

On a related note: I can imagine an impl in which the continuation token for epoch-proof-listing is `continue_from_epoch: Epoch` (instead of state version). But we don't have to pursue that if too cumbersome for "Rust typing" reasons.

core-rust/core-api-server/src/core_api/handlers/stream_proofs.rs
```diff
@@ -0,0 +1,211 @@
...
+
+fn iterate_all_proofs<'a>(
+    database: &'a StateManagerDatabase,
+    continue_from_state_version: Option<StateVersion>,
+    from_state_version: StateVersion,
+) -> Result<
+    Box<dyn Iterator<Item = LedgerProof> + 'a>,
+    ResponseError<models::StreamProofsErrorDetails>,
+> {
+    let start = optional_max(from_state_version, continue_from_state_version);
```
jakrawcz-rdx, line 87: couldn't we just disregard the `from_state_version` when continuation token is present?

(for bonus points: `if from_state_version > continue_from_version { return Err("you tried to use the continuation token with different filters!") }`)

core-rust/core-api-server/src/core_api/handlers/stream_proofs.rs
```diff
@@ -0,0 +1,211 @@
...
+
+    let iter = database.get_protocol_update_execution_proof_iter(start);
+
+    Ok(match protocol_version {
+        Some(protocol_version) => Box::new(iter.filter(move |proof| {
```
jakrawcz-rdx, line 140: the `.filter()` here suggests that maybe we shouldn't try hard to support the protocol_version filter in the API?
_(unless it is a killer feature from convenience PoV)_

core-rust/core-api-server/src/core_api/handlers/stream_proofs.rs
```diff
@@ -0,0 +1,211 @@
...
+fn iterate_end_of_epoch_proofs<'a>(
+    database: &'a StateManagerDatabase,
+    continue_from_state_version: Option<StateVersion>,
+    from_epoch: Epoch,
+) -> Result<
+    Box<dyn Iterator<Item = LedgerProof> + 'a>,
+    ResponseError<models::StreamProofsErrorDetails>,
+> {
+    let continuation_next_epoch = match continue_from_state_version {
+        Some(state_version) => match database.get_proof_iter(state_version).next() {
```
LukasGasior1, line 101: looks like we could just use next epoch as a continuation token for this filter

LukasGasior1: but that requires a separate `to_api_page` call for different filter types above, so prolly not worth it :)

LukasGasior1 (approved):
Nice!

core-rust/core-api-server/core-api-schema.yaml
dhedey: Good spot!

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -7072,6 +7144,175 @@ components:
...
+          $ref: "#/components/schemas/StreamProofsFilter"
+        max_page_size:
+          description: If specified, the maximum number of proofs that will be returned.
+          type: integer
+        continuation_token:
+          $ref: "#/components/schemas/ContinuationToken"
+    StreamProofsFilterType:
+      type: string
+      enum:
+        - All
```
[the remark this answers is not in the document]
dhedey: Hmm it gets quite long then, and inconsistent with the rest of the naming schemes.

Maybe `Any` might be better than `All`? Will change that, but let me know what you think

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -7072,6 +7144,175 @@ components:
...
+############################
+# REQUEST: /stream/proofs #
+###########################
+    StreamProofsRequest:
+      description: |
+        A request to retrieve a sublist of proofs.
+      type: object
+      required:
+        - network
+        - filter
```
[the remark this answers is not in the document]
dhedey: I guess we can do. Implicitly defaults to "Any" with no `from_state_version`?

I'm not sure I understand your `from_state_version` point though :)

core-rust/core-api-server/core-api-schema.yaml
dhedey: Will clarify, thanks 👍

core-rust/core-api-server/core-api-schema.yaml
dhedey: Ooops, good spot, thanks!

core-rust/core-api-server/src/core_api/conversions/pagination.rs
dhedey: Possibly, but would slightly prefer to keep it as-is, for conformity / discoverability.

(Currently I use `to_api_` and `extract_` autocompletes in my IDE to find relevant helper mappers and extractors.)

core-rust/core-api-server/src/core_api/conversions/pagination.rs
dhedey: Hmm so I want to build a toolkit which mandates/encourages correct use of the new API specs. In that case, continuation token is always optional; so I think this is correct. Will add a comment 👍

core-rust/core-api-server/src/core_api/handlers/stream_proofs.rs
dhedey: > but in general, it can happen to a valid token, when elements are deleted from the end of the collection. So maybe an empty return is more appropriate than error?

Agreed in general, but I think either's fine here. It should never happen in practice, and is a user error, so I think it's fine to 400. This code is all just for converting state version to next epoch anyway.

> On a related note: I can imagine an impl in which the continuation token for epoch-proof-listing is continue_from_epoch: Epoch (instead of state version). But we don't have to pursue that if too cumbersome for "Rust typing" reasons.

Yeah I considered making the continuation token the epoch in this case....

I was torn, but in the end, decided that viewing proofs as having a primary key of state version, regardless of which index we read from was conceptually nicer. I'm honestly not totally sure if that was the right choice, but it was quite arbitrary and honestly I don't think it matters _too_ much. So would be tempted to leave unless you feel strongly.

core-rust/core-api-server/src/core_api/handlers/stream_proofs.rs
dhedey: True, we could. I honestly don't think it matters. We agreed that behaviour is undefined if the user changes their filters.

I think this feels the most natural way of handling such a case, but I don't mind too much.

core-rust/core-api-server/src/core_api/handlers/stream_proofs.rs
dhedey: There shouldn't be too many such proofs, so I don't think it's _too_ lacking in performance. I think it's a useful feature but if you're concerned about perf we could consider dropping it?

core-rust/core-api-server/src/core_api/handlers/stream_proofs.rs
dhedey: Yeah I considered it! Maybe more natural :D. (see thread: https://github.com/radixdlt/babylon-node/pull/819#discussion_r1465796089 )

But I don't think worth changing for now given we're trying to get in for anemone.

core-rust/core-api-server/src/core_api/handlers/stream_proofs.rs
jakrawcz-rdx: > viewing proofs as having a primary key of state version, regardless of which index we read from was conceptually nicer.

From one point of view ("the exact index used is only an impl detail"), I agree.
But then again, the contents of ContinuationToken is (very explicitly) an opaque impl detail too... It would clearly be more convenient for the DB layer if ContinuationToken contained the epoch directly (i.e. it would save an extra DB roundtrip [which I point out for conceptual reasons, not performance]). But I understand it is less convenient to the controller (for Rust typing reasons, or code reusability reasons).

So yeah I don't feel strongly in this case, we can keep "whatever is easier" here, and just allow "DB-driven ContinuationToken contents" in general. After all, this _is_ an impl detail, and our "API conventions" don't even need to force anything here.

core-rust/core-api-server/src/core_api/handlers/stream_proofs.rs
jakrawcz-rdx: > We agreed that behaviour is undefined if the user changes their filters.

hah, `continuation_token.or_else(from_state_version)` is "undefined behavior" too... it is a shorter impl, and ofc can confuse user in a different way than `optional_max()`.

TBH the more I think, the more I am convinced that `HTTP 400: please don't tamper with continuation token and filters` is worth it (and AFAIU, Krzysztof too). But that's a topic for a future API conventions adjustment.

core-rust/core-api-server/src/core_api/handlers/stream_proofs.rs
jakrawcz-rdx: > I think it's a useful feature

that's enough for me 👍

jakrawcz-rdx (approved):
Very cool 👍

core-rust/core-api-server/src/core_api/handlers/stream_proofs.rs
```diff
@@ -0,0 +1,211 @@
...
+    assert_matching_network(&request.network, &state.network)?;
+    let mapping_context = MappingContext::new(&state.network);
+
+    let filter = request.filter.unwrap_or(Box::new(StreamProofsFilterAny {
+        from_state_version: None,
+    }));
+
+    let page_size = extract_valid_size(
+        request.max_page_size,
+        SizeRange {
```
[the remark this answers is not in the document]
dhedey: Different endpoints currently have different max sizes (partly due to how large we expect the responses to be). Perhaps [1, 20, 100] is reasonably standard, but I'll wait till we have more endpoints to see.

PR #822, opened 2024-01-24
Add protocol-related metrics

LukasGasior1:


core-rust/state-manager/src/metrics.rs
```diff
@@ -330,6 +344,202 @@ impl CommittedTransactionsMetrics {
...
+                ),
+                &["readiness_signal_name"],
+            )
+            .registered_at(registry),
+            pending_update_threshold_required_ratio: GaugeVec::new(
+                opts(
+                    "protocol_update_pending_threshold_required_ratio",
+                    "Required readiness ratio for the given protocol update threshold.",
+                ),
+                &["protocol_version_name", "readiness_signal_name", "threshold_index"],
```
dhedey, line 363: Maybe we should include a `threshold_epochs` label too in these :)

core-rust/state-manager/src/metrics.rs
```diff
@@ -330,6 +344,202 @@ impl CommittedTransactionsMetrics {
...
+                .registered_at(registry),
+            pending_update_threshold_required_consecutive_epochs: IntGaugeVec::new(
+                opts(
+                    "protocol_update_pending_threshold_required_consecutive_epochs",
+                    "Required number of consecutive epochs of support for the given protocol update threshold.",
+                ),
+                &["protocol_version_name", "readiness_signal_name", "threshold_index"],
+            )
+                .registered_at(registry),
+            pending_update_threshold_matching_consecutive_epochs: IntGaugeVec::new(
```
dhedey, line 374: Ohh nice idea! :)

We should have a `pending_update_threshold_required_consecutive_epochs` as well as a `pending_update_threshold_current_consecutive_epochs` so we can draw a graph and check the current vs the required.

core-rust/state-manager/src/metrics.rs
```diff
@@ -330,6 +344,202 @@ impl CommittedTransactionsMetrics {
...
+                    "protocol_update_pending_upper_bound_epoch",
+                    "Upper bound epoch for the given protocol update (exclusive)",
+                ),
+                &["protocol_version_name", "readiness_signal_name"],
+            )
+                .registered_at(registry),
+            enacted_protocol_update_state_version: IntGaugeVec::new(
+                opts(
+                    "protocol_update_enacted_state_version",
+                    "State version of enacted protocol update",
```
dhedey, line 401: Maybe worth clarifying it's the state version of the init of the protocol update?

dhedey (approved):
Have just reviewed the definitions, not the implementation, but it looks good :). Have a few suggestions!

core-rust/state-manager/src/metrics.rs
LukasGasior1: yup, that's the "current" (I'll rename it)

core-rust/state-manager/src/metrics.rs
LukasGasior1: hmh it's already there as a separate gauge, not sure why we would need it as a label?

dhedey: Fair - I was just thinking if you want a single graph of these lines, along with the current ratios, being able to label with the threshold epochs is probably more informative than the threshold index

LukasGasior1: ah, gotcha :+1: okay, I'll add it

PR #823, opened 2024-01-24
feature: Properly configure protocol configs

dhedey:


core-rust/state-manager/src/protocol/protocol_configs/config_printer.rs
```diff
@@ -0,0 +1,238 @@
...
+            SignalledReadinessThreshold {
+                required_ratio_of_stake_supported: threshold,
+                required_consecutive_completed_epochs_of_support: epochs,
+            }
+        })
+        .collect::<Vec<_>>();
+
+    let trigger = ProtocolUpdateTrigger::of(
+        version,
+        ProtocolUpdateEnactmentCondition::EnactAtStartOfEpochIfValidatorsReady {
```
jakrawcz-rdx, line 57: I remember there was some "validate" (or "sanity check"?) method available on these configs (which checks `end >= start` etc). If yes, then worth calling?

core-rust/state-manager/src/protocol/protocol_configs/config_printer.rs
```diff
@@ -0,0 +1,238 @@
...
+    } else if hours > 0 {
+        format!("{hours} hours, {mins} mins, {secs} secs")
+    } else if mins > 0 {
+        format!("{mins} mins, {secs} secs")
+    } else if secs > 0 {
+        format!("{secs} secs")
+    } else {
+        "Immediate".to_string()
+    }
+}
```
jakrawcz-rdx, line 177: (optional) I've seen a few crates doing "human-readable duration".
I know it is 20 lines, but personally I'd accept an extra dev-only dependency (in exchange for avoiding a bug in a hand-crafted time math).

jakrawcz-rdx (approved):
Nice 👍 

The config matches the plan at https://rdxworks.slack.com/archives/C01M90Y2448/p1705924939276209 and the numbers (and comments :p) outputted by the calculator make sense.

PR #825, opened 2024-01-25
Implementing filters for entity listing endpoint

jakrawcz-rdx:
The "filters" part of https://radixdlt.atlassian.net/wiki/spaces/RPNV1/pages/3233153027/RNP-14+-+Browse+API+-+Slice+2#Filtering-%2F-sorting-for-the-Entity-Listing-endpoint .

This PR allows to filter `/entity/iterator` by:
- system type (object or KV-store)
- entity type
- blueprint (package + name pair)

Sorting is not addressed at all (i.e. still the only supported sort is by creation asc).
I left one TODO: for validating that filters were not changed across pages.

I also acknowledge that paging infra could use some simplification, but let's _not_ make it a scope of this PR (of course, suggestions are welcome).


jakrawcz-rdx:
lemme merge this to the long-lived branch, since next work stacks on top of it

PR #826, opened 2024-01-25
Clear peer bans when booting near protocol update

LukasGasior1:


dhedey (approved):
LGTM. Although can we change to merge into `release/anemone` instead please?

I assume this is because, from the old node, the new unknown field in the sync response, failing validation? So for safety, the response is not processed (and the sender is reasonably banned)?

(Side note - this whole "StartProcessorOnRunner" feels a little magic to me, but I guess we can make things a little less magic / more configuration-oriented when we rebuild in rust).

LukasGasior1:
> I assume this is because, from the old node, the new unknown field in the sync response, failing validation? So for safety, the response is not processed (and the sender is reasonably banned)?

Unknown field is fine, this goes deeper :smile: 
On new version, vote hash (what is being signed) includes `next_protcol_version`, while old nodes don't include it when they calculate the hash they use to verify the signature, so the error is actually validator signature mismatch.

PR #827, opened 2024-01-25
Less disturbing divergent execution log if the cause is a protocol update mismatch

LukasGasior1:


core/src/main/java/com/radixdlt/consensus/PendingVotes.java
```diff
@@ -124,17 +124,32 @@ private void checkForDivergentVertexExecution(Vote vote) {
...
-            vote,
-            otherVote);
+        if (!voteLedgerHeader
+            .nextProtocolVersion()
+            .equals(otherLedgerHeader.nextProtocolVersion())) {
+          log.info(
+              """
+              Received BFT votes for conflicting protocol update headers.
+              Validator {} enacts {}, while validator {} enacts {}.
+              This indicates that one or more validators run an outdated node version.
```
dhedey, line 134: This isn't always true - when it comes to the next update, it could be that the other ledger header is the one with the protocol version and therefore the one that has been updated.

I suggest splitting the cases of:
* Ours has a next protocol version, their's has no protocol version => Likely they are running an outdated version, all good. Probably a debug, not even an info(!)
* Theirs has protocol version, ours doesn't match => Possibly we're running an outdated version? Or potentially it's just more generally a malicious node messing with us. We probably want a variant of the `warn` message, with some initial emphasis that we might be running an outdated version.

Other issues:
* Also I notice the `checkForDivergentVertexExecution(vote)` happens before we even check the vote is from someone in the validator set. We should probably move it a step later.
* These logs may need limiting in some manner? Otherwise we could have a lot of spam if we're logging for every vote in every round! Although I guess ideally this will only happen for one round before the other validators crash... So maybe this is okay.
* If the validator set actually commits the next protocol update (which I imagine will be the standard here - most of the validator set will have updated!), we're going to crash anyway when trying to sync the last transactions of the previous epoch, with a divergent resolved proof. Can we double check we do do this assertion when we sync, and that we have an explicit log for this case before we crash / panic? e.g. something along the lines of "FATAL: The validator set have enacted anemone, but this node does not know about anemone." _or_ "FATAL: The validator set have enacted anemone, but this node does not expect anemone to trigger at this point.". I'm sure we do have that check, but I just want to sanity check.

LukasGasior1: > This isn't always true - when it comes to the next update, it could be that the other ledger header is the one with the protocol version and therefore the one that has been updated.

Well, this is _almost_ always true, IMO true-enough for this log message :smile:
If there are two votes that produce distinct `next_protcol_version` then they must use different `ProtocolConfig`s, of which one must be newer - unless there's a bug or someone has deliberately misconfigured their nodes (which could be considered "outdated"), but there's no need to provide such details here. The message just says that "one or more validators [in the validator set] run an outdated node", it doesn't specify if it's this node or some other.

There isn't really distinction between "ours" and "theirs". These are just two votes from two validators in the validator set. Although maybe it's worth adding it and extending the messages, as you suggest :thinking: 


> Also I notice the checkForDivergentVertexExecution(vote) happens before we even check the vote is from someone in the validator set. We should probably move it a step later.

Hmh seems that we actually check that twice (first time in `BFTEventStatelessVerifier`).
But agreed, I'll move it a step later, which is more correct from `PendingVotes`s perspective.

> These logs may need limiting in some manner? Otherwise we could have a lot of spam if we're logging for every vote in every round! Although I guess ideally this will only happen for one round before the other validators crash... So maybe this is okay.

Good point, I'll make sure this isn't too spammy.

> If the validator set actually commits ...  but I just want to sanity check.

We have:
`resultant protocol update at version {} differs from the proof ({:?} != {:?})`.
Which isn't too friendly, I'll update it.

dhedey: > There isn't really distinction between "ours" and "theirs". These are just two votes from two validators in the validator set. Although maybe it's worth adding it and extending the messages, as you suggest 🤔

Ahhh right, I see! So validator X and Y aren't necessarily either you :D.
I think this is fine, and sensible :). We then get a log of all the different validators which run an outdated version before they then drop out of consensus next epoch.

From a validator runner's perspective, there is a rather big difference between some other validator is out of date (I don't care), and your node is out of date (I do care). But as mentioned, it will crash at commit anyway :) - so I think that's the better place for this log message, as we're doing.

dhedey: Basically, I'm happy with your above message / proposals.

core-rust/state-manager/src/state_computer.rs
```diff
@@ -1270,6 +1268,58 @@ where
...
+        next_version_from_proof: &Option<ProtocolVersionName>,
+    ) {
+        match (next_version_from_local_execution, next_version_from_proof) {
+            (Some(local_version), Some(proof_version)) if local_version == proof_version => {
+                // All good, local execution enacts the same protocol version as proof. No-op.
+            }
+            (Some(local_version), Some(proof_version)) => {
+                // We're enacting a different version than the proof
+                panic!(
+                    "FATAL: At state version {} this node enacts {}, but this doesn't match the proof, which enacts {} (i.e. the validator set have enacted a different version). Make sure your node is running the latest software version.",
```
dhedey, line 1285: To avoid confusion, can we change this to "this node wants to enact" and make clear we're terminating without enacting anything?

I want to avoid node runners being concerned that they think their ledger has corrupted from this.

core-rust/state-manager/src/state_computer.rs
```diff
@@ -1270,6 +1268,58 @@ where
...
+                    commit_state_version,
+                    local_version,
+                    proof_version,
+                );
+            }
+            (Some(local_version), None) => {
+                // We're enacting locally, but the proof doesn't enact anything
+                panic!(
+                    "FATAL: At state version {} this node enacts {}, but this doesn't match the proof, which doesn't enact anything (i.e. the validator set haven't enacted it). Make sure your node is running the latest software version.",
```
dhedey, line 1294: (Similar here)

core-rust/state-manager/src/state_computer.rs
```diff
@@ -1270,6 +1268,58 @@ where
...
+                    commit_state_version,
+                    local_version,
+                );
+            }
+            (None, Some(proof_version)) => {
+                // The proof enacts, but we locally don't
+                if new_protocol_state.has_pending_update(proof_version) {
+                    panic!(
+                        "FATAL: At state version {} the validator set have enacted {}, this node has {} in its pending updates, but didn't expect it to happen at this state version. Make sure your node is running the latest software version.",
```
dhedey, line 1303: Suggestion - this branch is perhaps confusing (a node runner doesn't know what pending means in the context of the protocol state).

Instead, I suggest let's just have a single panic here saying  that either this node doesn't node about X or doesn't expect it to enact at this point.

dhedey: Separately - can we add a short rust doc to Protocol Update fields explaining what e.g. pending_x means :)

dhedey (approved):

dhedey (approved):

PR #828, opened 2024-01-26
Refactor: Simplify paging infra within Engine State API

jakrawcz-rdx:
As announced in https://github.com/radixdlt/babylon-node/pull/825.

This PR:
- removes the `Pager` trait (we only have one impl)
- switches the impl from "last key of returned page" to much simpler "first key of next page"
- removes the unnecessarily-flexible, too-complex "iterable" abstraction (in favor of directly getting iterator in higher layer)
- _[DRIVE-BY] [FEATURE]_ returning HTTP 400 after detecting a filter changed across pages (via hash)


core-rust/node-http-apis/src/engine_state_api/handlers/mod.rs
```diff
@@ -26,68 +27,88 @@ pub(crate) use object_collection_iterator::*;
...
+    pub fn new<F: serde::Serialize>(
+        max_page_size: Option<i32>,
+        continuation_token: Option<String>,
+        filter: &F,
+    ) -> Self {
         Self {
             max_page_size,
-            continuation_token_string,
+            requested_continuation_token_string: continuation_token,
+            requested_filter_hash: hash(serde_json::to_vec(filter).expect("it was serialized")),
```
krzlabrdx, line 53: Is this hash guaranteed to be stable? JSON serialization can produce different output based on a configuration settings which might change over time:
```
{"filter":"abćs"}
{\n  "filter": "abćs"\n}
{"filter":"ab\u0107"}
{\n\t"filter": "abćs"\n}
```

I guess this won't be a problem as long as we run the exact same Node version with the exact same configuration on the exact same platform. But none of that can be guaranteed.

jakrawcz-rdx: > I guess this won't be a problem as long as we run the exact same Node version with the exact same configuration on the exact same platform.

this is true, but luckily we only need these to be guaranteed for a moment (i.e. across paging).

I am not sure we have this documented, but in our discussions so far, we assumed that continuation token has a lifespan of at most ~minutes (i.e. only long enough to actually iterate through pages). We plan to use this e.g. when implementing consistent paging using RocksDB snapshots (i.e. a [super-cheap] DB snapshot is taken on start of iteration, and subsequent pages can then read from the same DB state, and it is available until some predefined timeout, and after that the API returns "HTTP 400 continuation token expired" or something).

A similar concern can be raised about "what if we refactor some `filter` field name or extend its schema". Ofc such token would stop working after a deploy. And we basically do not care, our continuation tokens are not considered "persistent".

jakrawcz-rdx: However, you just made me realize that this can be problematic when someone routes subsequent page requests to different Node instances.

(I mean: this "JSON serialization hash" itself won't be an issue, but any kind of "DB snapshot ID" will be!)

We definitely have to be careful in this area 👍 

But for the "filter hash", I am ok with best-effort solution which breaks across deploys. (after all, worst that can happen is: User says "huh, some weird HTTP 400, let me just restart listing from the beginning")

krzlabrdx: As I said I also thought this is acceptable tradeoff. My point was that it can get easily overlooked compared to say `filter` change as it's not so obvious.

jakrawcz-rdx:
lemme merge this to the long-lived branch, since next work stacks on top of it

PR #830, opened 2024-01-30
Better logs/metrics when divergent vertex execution is detected

LukasGasior1:


core/src/main/java/com/radixdlt/consensus/DivergentVertexExecutionDetector.java
```diff
@@ -0,0 +1,157 @@
...
+    final var ledgerHeadersByVertexId =
+        votesByVertexId.computeIfAbsent(
+            vote.getVoteData().getProposed().getVertexId(), unused -> new HashMap<>());
+    final var authorsByLedgerHeader =
+        ledgerHeadersByVertexId.computeIfAbsent(
+            vote.getVoteData().getProposed().getLedgerHeader(), unused -> new HashSet<>());
+    authorsByLedgerHeader.add(vote.getAuthor());
+  }
+
+  public void logAndUpdateMetrics(Metrics metrics, Round round) {
```
jakrawcz-rdx, line 106: Looking at the log message (`"In round %s ..."`), I assume that we only ever analyze/log these over a single round.

If this is true, I'd integrate `logAndUpdateMetrics() + clear()` as a single `logSummaryAfterRound()` _(name up to discussion)_.
Just so that forgetting about `clear()` is impossible.

core/src/main/java/com/radixdlt/consensus/DivergentVertexExecutionDetector.java
```diff
@@ -0,0 +1,157 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+/**
+ * A utility for detecting divergent vertex execution, i.e. a situation where two (or more)
+ * validators from the validator set have voted for the same vertexId, but their resultant ledger
+ * headers were different.
+ */
+@SuppressWarnings("UnstableApiUsage")
+public final class DivergentVertexExecutionDetector {
```
jakrawcz-rdx, line 83: After the refactors, this class now looks like a top-level injectable service:
- (highly recommended) the `Metrics` should be its injected dependency;
- (optional) maybe it could listen on `Vote` and/or `Round` events (instead of relying on those methods being explicitly called from the right places); _I actually dunno if this is feasible because I am still lost around all these events in Java land._

LukasGasior1: I think it's a bit clearer to have the detector being instantiated per consensus instance (which we create manually)

core/src/main/java/com/radixdlt/consensus/DivergentVertexExecutionDetector.java
LukasGasior1: sounds good :+1: will change

core/src/main/java/com/radixdlt/consensus/DivergentVertexExecutionDetector.java
LukasGasior1: but I'll move `metrics` to constructor

core/src/main/java/com/radixdlt/consensus/DivergentVertexExecutionDetector.java
LukasGasior1: perhaps `summarizeAfterRoundAndReset` ?

core/src/main/java/com/radixdlt/consensus/DivergentVertexExecutionDetector.java
jakrawcz-rdx: ahhh the manual instantiation; ok then 👍

core/src/main/java/com/radixdlt/consensus/DivergentVertexExecutionDetector.java
jakrawcz-rdx: "summarize" would suggest that a summary is returned :(

maybe `logRoundAndReset()`? `logAfterRoundAndReset()`? `onRoundFinished()` + "the logging and resetting is impl detail" approach?

LukasGasior1: > "summarize" would suggest that a summary is returned :(

maybe `finalize` then? `finalizeAfterRoundAndReset`
 
`log` indicates that it's the primary action but updating the metrics is equally important IMO, so either we include both in the name or use something more generic

jakrawcz-rdx: `finalize` is hated in Java Land already :D

ok, "summarize" does not seem that bad anymore 👍

jakrawcz-rdx (approved):

PR #841, opened 2024-02-08
tweak: Add test to calculate the start of a target epoch

dhedey:
Just a small change I had locally to help with calculation of target enactment epoch time before we have that information in the API.

### DIFF
--- a/core-rust/state-manager/src/protocol/protocol_configs/config_printer.rs
+++ b/core-rust/state-manager/src/protocol/protocol_configs/config_printer.rs
@@ -131,6 +131,20 @@ fn print_protocol_config_code() {
 }
 
 #[test]
+fn calculate_start_of_epoch() {
+    let calculator = CalculationParameters {
+        expected_epoch_length: Duration::minutes(5),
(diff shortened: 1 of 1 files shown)


core-rust/state-manager/src/protocol/protocol_configs/config_printer.rs
```diff
@@ -130,6 +130,20 @@ fn print_protocol_config_code() {
     println!("{base_indent}}},");
 }
 
+#[test]
+fn calculate_start_of_epoch() {
+    let calculator = CalculationParameters {
+        expected_epoch_length: Duration::minutes(5),
+        // This data can come from the Core API /core/state/consensus-manager response
+        base_epoch: Epoch::of(69946),
+        base_epoch_effective_start: DateTime::<Utc>::from_str("2024-02-05T11:55:57.229Z").unwrap(), //
```
LukasGasior1, line 139: there's an extra empty comment at the end of this line :smile:

LukasGasior1 (approved):
Looks good :+1:

PR #848, opened 2024-02-14
Add keygen binary to node dist (to Anemone)

LukasGasior1:
This PR adds a `keygen` (Unix) and `keygen.bat` (Windows) binaries from the `:keygen` subproject to `:core`'s output distribution zip. It also adds `keygen` as a dependency to `core`.

The resultant dist zip will contain all four files in its `bin` folder:

```
├── bin
│   ├── core
│   ├── core.bat
│   ├── keygen
│   └── keygen.bat
└── lib
```

This will remove the need to use a keygen docker image when setting up a node (e.g. if one doesn't want to use docker).


dhedey:
Okay, merged, and then merged `release/anemone => main` and `main => develop`.

PR #849, opened 2024-02-15
Bump scrypto dep to `develop-839137f1` + Fix imports for good

jakrawcz-rdx:
## Summary

This addresses https://radixdlt.atlassian.net/browse/NODE-596.

## Details

Apart from pulling the newest scrypto dep, this PR contains a slight convenience-only refactor:

Each of our crates now contains a top-level `pub(crate) mod scrypto_prelude { <lots of engine imports relevant to that crate> }`.
From this point on, all other files in these crates are supposed to only import a single `use crate::scrypto_prelude::*;` (instead of many independent engine imports).

## Testing

No Node feature / behavior change here - existing tests are sufficient.

### DIFF
--- a/core-rust/Cargo.lock
+++ b/core-rust/Cargo.lock
@@ -207,6 +207,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "blueprint-schema-init"
+version = "1.2.0-dev"
+source = "git+https://github.com/radixdlt/radixdlt-scrypto?tag=develop-839137f1#839137f1b34f45273c51429459221080b4d41b1d"
+dependencies = [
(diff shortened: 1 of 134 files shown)


dhedey (approved):
Very nice! :D

However can I request a slightly different name - e.g. `engine_prelude` rather than `scrypto_prelude`?

Because `scrypto::prelude::*` is already a thing - it's specifically the prelude used by scrypto developers building scrypto packages -- whereas we're importing the engine rather than scrypto.

jakrawcz-rdx:
alright, renamed; gonna merge when CI goes 🟢

PR #851, opened 2024-02-23
Implement `SubstateDatabase`-at-specific-version backed by JMT

jakrawcz-rdx:
## Summary

Addresses the first part of https://radixdlt.atlassian.net/browse/LZ-18.

## Testing

The impl is unit-tested (albeit against a mock storage of historically-upserted values).
None of the code here is referenced by production logic.

### DIFF
--- a/core-rust/state-manager/src/store/historical_state.rs
+++ b/core-rust/state-manager/src/store/historical_state.rs
@@ -1,0 +1,653 @@
+/* Copyright 2021 Radix Publishing Ltd incorporated in Jersey (Channel Islands).
+ *
+ * Licensed under the Radix License, Version 1.0 (the "License"); you may not use this
+ * file except in compliance with the License. You may obtain a copy of the License at:
+ *
+ * radixfoundation.org/licenses/LICENSE-v1
+ *
+ * The Licensor hereby grants permission for the Canonical version of the Work to be
+ * published, distributed and used under or by reference to the Licensor’s trademark
+ * Radix ® and use of any unregistered trade names, logos or get-up.
+ *
+ * The Licensor provides the Work (and each Contributor provides its Contributions) on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied,
+ * including, without limitation, any warranties or conditions of TITLE, NON-INFRINGEMENT,
+ * MERCHANTABILITY, or FITNESS FOR A PARTICULAR PURPOSE.
+ *
+ * Whilst the Work is capable of being deployed, used and adopted (instantiated) to create
+ * a distributed ledger it is your responsibility to test and validate the code, together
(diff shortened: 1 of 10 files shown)


core-rust/state-manager/src/store/traits.rs
```diff
@@ -222,6 +222,33 @@ pub mod substate {
...
+        /// - `get_value("X", 14) -> None`
+        ///
+        /// In other words, it only records the *acts of upserting* key-value pairs. It also
+        /// deliberately disregards deletes.
+        fn get_value(
+            &self,
+            partition_key: &DbPartitionKey,
+            sort_key: &DbSortKey,
+            upserted_at_state_version: StateVersion,
+        ) -> Option<DbSubstateValue>;
```
jakrawcz-rdx, line 250: _(note from author)_

@dhedey @talekhinezh the API of this storage kinda assumes that we are NOT pursuing the ["general-purpose DbTable<Hash, SubstateValue>"](https://radixdlt.atlassian.net/wiki/spaces/RPNV1/pages/3313532929/RNP-15+-+Historical+State#Alternative-%E2%80%9CDecouple-Substate-Hash-%E2%86%92-Value%E2%80%9C-Idea) approach. If you find it tempting enough, now is time to discuss it.

core-rust/state-manager/src/store/historical_state.rs
```diff
@@ -0,0 +1,657 @@
...
+) -> Box<dyn Iterator<Item = ResolvedLeaf> + 't> {
+    let Some(node) = tree_store.get_node(&at_key) else {
+        panic!("{:?} referenced but not found in the storage", at_key);
+    };
+    match node {
+        TreeNode::Internal(internal) => {
+            let mut child_from_nibbles = from_nibbles;
+            let from_nibble = child_from_nibbles
+                .pop_front()
+                .unwrap_or_else(|| Nibble::from(0));
```
LukasGasior1, line 234: Super minor comment, but my slight preference would be to use an Option here (and e.g. `from_nibble.map_or(true, |fn| child.nibble >= fn)` below).
I know this totally doesn't matter (it's just my OCD :smile: ) as both cases are really the same, but in the `if ` below we could go the `mem:take` path if we're actually comparing against the last `0` from `from_nibbles` and  `VecDeque::new()` if it's None (or a "default 0"). Currently we go the mem:take path in both cases.
But again, that's just a personal-preference-based nitpicking and doesn't influence the correctness, feel free to ignore.

LukasGasior1 (approved):
Really clean impl :heart: LGTM

core-rust/state-manager/src/store/historical_state.rs
jakrawcz-rdx: Done; it turns out that handling the Option here is actually even less code than the `Nibble::from(0)` 👍

PR #854, opened 2024-02-27
Add vertex store size limit; cleanup vertex store events

LukasGasior1:
This PR includes:

1) The primary feature
- added a size limit to the vertex store. The limit applies to the size of a dson-serialized vertex store state. The limit is "soft", in a sense that in some circumstances it might be slightly exceeded (e.g. inserting a newer QC/TC that has more signatures than the previous one) - for practicality reasons. The main goal is to protect against SBOR overflow (e.g. when encoding a commit request containing the vertex store) and in practice there should always be a lot of room between this limit and the critical SBOR limit - enough room not to worry about such small values. The limits defaults to 150 Mb and can be configured with `bft.vertex_store.max_serialized_size_bytes` (min value is 10 Mb).

2) Additional related changes
- vertex store serialization has been moved from state computer to the vertex store itself
- updating vertex store metrics has been moved to the vertex store itself
- vertex store events now include the serialized state to be persisted

3) Drive-by
- small refactor to vertex store events: they're now `record`s and there's no longer a separate `BFTCommittedUpdate` event. Since a (consensus) commit can only happen as a result of QC insertion, this has been captured in `BFTHighQCUpdate`, which allowed us to get rid of the weird logic in `dispatchPostQcInsertionEvents` in VertexStoreAdapter
- added vertex store persist after inserting a TC (which was missing)


core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreJavaImpl.java
```diff
@@ -350,43 +431,15 @@ private void removeVertexAndPruneInternal(HashCode vertexId, HashCode skip) {
     var children = vertexChildren.remove(vertexId);
     if (children != null) {
       for (HashCode child : children) {
-        if (!child.equals(skip)) {
-          removeVertexAndPruneInternal(child, null);
+        if (!skip.map(child::equals).orElse(false)) {
```
jakrawcz-rdx, line 434:
```suggestion
        if (!Optional.of(child).equals(skip)) {
```

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreJavaImpl.java
```diff
@@ -302,45 +340,88 @@ public InsertVertexChainResult insertVertexChain(VertexChain vertexChain) {
...
   }
 
-  private void removeVertexAndPruneInternal(HashCode vertexId, HashCode skip) {
-    vertices.remove(vertexId);
+  private void removeVertexAndPruneInternal(HashCode vertexId, Optional<HashCode> skip) {
+    Optional.ofNullable(vertices.remove(vertexId))
+        .flatMap(
+            removedVertex ->
+                Optional.ofNullable(vertexChildren.get(removedVertex.vertex().getParentVertexId())))
+        .ifPresent(siblings -> siblings.remove(vertexId));
```
jakrawcz-rdx, line 423: some lines could be a lot less awkward if `this.vertexChildren` was really a `Multimap<HashCode, HashCode>` (e.g. `= HashMultimap.create()`)

jakrawcz-rdx (approved):
The vertex store size limit feature (i.e. `break vertices_loop;`) looks very much ok 👍 (I mean, I was able to understand it, and it seems to work [outside of corner-cases where "really small stuff is added to vertex store without checking size"]). 

I can't say I understood 100% of the rest (i.e. the refactor) :(
I left some minor "technical Java" comments.
It does look more structured than before, but I definitely lack the Consensus knowledge to spot any new subtle bugs.

If you have faith in our Consensus regression tests - feel free to merge with my 🟢 .
Otherwise - I recommend waiting for more reviewers.

common/src/main/java/com/radixdlt/monitoring/Metrics.java
```diff
@@ -233,7 +233,12 @@ public record Sync(
         Counter invalidEpochInitialQcSyncStates) {}
 
     public record VertexStore(
-        Gauge size, Counter forks, Counter rebuilds, Counter indirectParents) {}
+        Gauge size,
```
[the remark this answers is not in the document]
LukasGasior1: sounds good :+1: 
note to self: update this on the dashboard later (if it's used)

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreJavaImpl.java
LukasGasior1: Splendid idea! Done.

core/src/test-core/java/com/radixdlt/harness/simulation/monitors/consensus/LivenessInvariant.java
```diff
@@ -99,14 +98,14 @@ public Observable<TestInvariantError> check(RunningNetwork network) {
...
                   },
                   BFTHighQCUpdate.class);
               nodeEvents.addListener(
-                  (node, committed) -> {
-                    emitter.onNext(committed.vertexStoreState().getHighQC().highestQC());
+                  (node, highQcUpdate) -> {
+                    emitter.onNext(highQcUpdate.newHighQc().highestQC());
                   },
-                  BFTCommittedUpdate.class);
+                  BFTHighQCUpdate.class);
```
dhedey, line 108: I think this code is now a duplicate so should be deleted?

core/src/main/java/com/radixdlt/ledger/StateComputerLedger.java
```diff
@@ -162,7 +161,8 @@ StateComputerPrepareResult prepare(
         List<RawNotarizedTransaction> proposedTransactions,
         RoundDetails roundDetails);
 
-    LedgerProofBundle commit(LedgerExtension ledgerExtension, VertexStoreState vertexStore);
+    LedgerProofBundle commit(
+        LedgerExtension ledgerExtension, Option<byte[]> serializedVertexStoreState);
```
dhedey, line 165: Out of interest, why is this `Option<byte[]>` and not `Option<WrappedByteArray>`?

core/src/integration/java/com/radixdlt/integration/steady_state/deterministic/consensus/DivergentExecutionLivenessBreakTest.java
```diff
@@ -0,0 +1,356 @@
...
+ *    while staying below the critical SBOR limit)
+ *
+ * The test consists of 3 phases:
+ * - Phase 1 (rounds 0 to LIVENESS_BREAK_START_ROUND): initial liveness (to verify that the test setup is correct)
+ * - Phase 2 (from LIVENESS_BREAK_START_ROUND): liveness break due to divergent vertex execution (different hashes)
+ * - Phase 3 Recovery. The nodes reboot with a new configuration (slightly increased vertex store size limit)
+*            and the forced divergent vertex execution in `prepare` is reverted.
+ */
+// spotless:on
+public final class DivergentExecutionLivenessBreakTest {
```
dhedey, line 125: Nice test 👍

dhedey (commented):
Generally really nice changes. I like the event fixing, the metrics tweaks, and the use of sealed interfaces for better error propagation.

The code in VertexStoreImpl isn't the prettiest, but then again, the whole class isn't great to start with... I wonder if we should change the size check to a simple check on current (cached) serialized size before we try inserting each vertex; rather than a check of the state-after-inserting-the-vertex? This would reduce the work done on the happy path, and mean we can be happier in not checking the size on the other code paths which just update the highQcs. Anyway, just an idea.

I don't think any of these notes are blockers, but ideally some of the more important ones could do with a small fix before we merge.

core/src/test-core/java/com/radixdlt/harness/simulation/monitors/consensus/LivenessInvariant.java
LukasGasior1: yup, good catch

core/src/main/java/com/radixdlt/ledger/StateComputerLedger.java
LukasGasior1: no reason at all, I'll change to `Option<WrappedByteArray>`

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreJavaImpl.java
```diff
@@ -302,91 +339,102 @@ public InsertVertexChainResult insertVertexChain(VertexChain vertexChain) {
...
       return;
     }
 
-    var children = vertexChildren.remove(vertexId);
-    if (children != null) {
-      for (HashCode child : children) {
-        if (!child.equals(skip)) {
-          removeVertexAndPruneInternal(child, null);
-        }
+    final var children = vertexChildren.removeAll(vertexId);
```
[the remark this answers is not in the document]
LukasGasior1: Yup, good point. I prefer a single delete too, for readability. :+1: changing this to `get`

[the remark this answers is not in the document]
LukasGasior1: and actually that was a bug! the if statement in the loop below could leave vertex store inconsistent if we removed the `skip` vertex from `vertexChildren`

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreJavaImpl.java
```diff
@@ -302,91 +339,102 @@ public InsertVertexChainResult insertVertexChain(VertexChain vertexChain) {
...
     }
 
     final var vertex = vertexWithHash.vertex();
     if (!this.containsVertex(vertex.getParentVertexId())) {
       throw new MissingParentException(vertex.getParentVertexId());
     }
 
-    return insertVertexInternal(vertexWithHash);
-  }
+    // Before we execute the vertex, let's check if we can fit it into the store...
```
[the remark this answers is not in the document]
LukasGasior1: Hmh, there isn't really any extra burden during normal operation though. We only serialize once (before insert, before execution), and if everything is good we use that state later on.
It seems that what you suggest with `recordHighQcOrVertexChange` is to increase the readability for the price of less strict limit enforcement, is that so? I'd personally prefer to keep the validation as accurate as we can.

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreJavaImpl.java
```diff
@@ -302,91 +339,102 @@ public InsertVertexChainResult insertVertexChain(VertexChain vertexChain) {
...
+      if (insertRes.isSuccess()) {
+        bftInsertUpdates.add(insertRes.unwrap());
+      } else {
+        switch (insertRes.unwrapError()) {
+          case ALREADY_PRESENT, PREPARE_FAILED -> {
+            // No-op, continue iterating the vertices
+          }
+          case VERTEX_STORE_SIZE_EXCEEDED -> {
+            // Stop if we hit the size limit
+            break vertices_loop;
```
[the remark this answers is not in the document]
LukasGasior1: Yeah this isn't ideal but it's implicitly handled in bft sync, it just goes on with the sync process (that is: `syncToQC`) regardless of the result. I'll add a comment so that we don't forget to handle that better in the next bft sync implementation :smile:

core/src/main/java/com/radixdlt/rev2/REv2StateComputer.java
```diff
@@ -283,15 +284,15 @@ public StateComputerLedger.StateComputerPrepareResult prepare(
...
+      LedgerExtension ledgerExtension, Option<WrappedByteArray> serializedVertexStoreState) {
     final var proof = ledgerExtension.proof();
     final var header = proof.ledgerHeader();
 
     var commitRequest =
         new CommitRequest(
             ledgerExtension.transactions(),
             proof,
-            serializedVertexStoreState,
+            serializedVertexStoreState.map(WrappedByteArray::value),
```
dhedey, line 295: We should just always handle it as a `WrappedByteArray` I think? It's just safer than passing around a big `byte[]`.

If we need to, let's add an SborCodec for `WrappedByteArray`:

```
  public static void registerCodec(CodecMap codecMap) {
    codecMap.register(
        WrappedByteArray.class,
        codecs ->
            new CustomByteArrayCodec<>(
                WrappedByteArray::value,
                WrappedByteArray::new));
  }
```

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreJavaImpl.java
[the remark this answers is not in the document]
dhedey: But what would happen if this arises? Would sync essentially fail / get stuck in a weird state?
Might it be better to throw and crash the node than get stuck in a weird half-live state?

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreJavaImpl.java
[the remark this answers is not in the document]
LukasGasior1: As discussed _some time ago_ :wink:  I went with your proposal and extracted the logic to `recordHighQcOrVertexChange`

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreJavaImpl.java
```diff
@@ -434,8 +508,23 @@ public Option<ImmutableList<VertexWithHash>> getVertices(HashCode vertexHash, in
     return Option.present(builder.build());
   }
 
+  private void trackCurrentStateSize(WrappedByteArray serializedVertexStoreState) {
+    this.currentSerializedSizeBytes = serializedVertexStoreState.size();
+    metrics.bft().vertexStore().byteSize().set(this.currentSerializedSizeBytes);
+  }
+
+  private WrappedByteArray serializeState(VertexStoreState state) {
```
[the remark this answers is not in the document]
LukasGasior1: renamed

core/src/main/java/com/radixdlt/rev2/REv2StateComputer.java
LukasGasior1: done

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreJavaImpl.java
```diff
@@ -95,47 +96,64 @@ public final class VertexStoreJavaImpl implements VertexStore {
...
-    final var vertexStore =
-        new VertexStoreJavaImpl(
-            ledger, hasher, vertexStoreState.getRoot(), vertexStoreState.getHighQC());
-
-    for (var vertexWithHash : vertexStoreState.getVertices()) {
-      vertexStore.vertices.put(vertexWithHash.hash(), vertexWithHash);
-      vertexStore.vertexChildren.put(vertexWithHash.hash(), new HashSet<>());
-      var siblings = vertexStore.vertexChildren.get(vertexWithHash.vertex().getParentVertexId());
-      siblings.add(vertexWithHash.hash());
+  private void resetToState(VertexStoreState state, WrappedByteArray serializedState) {
```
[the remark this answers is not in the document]
LukasGasior1: done

core/src/main/java/com/radixdlt/ledger/StateComputerLedger.java
```diff
@@ -415,7 +425,7 @@ private void commit(LedgerExtension ledgerExtension, @Nullable VertexStoreState
     // synchronization theoretically needed here).
 
     this.metrics.ledger().stateVersion().set(this.latestProof.resultantStateVersion());
-    if (vertexStore == null) {
+    if (serializedVertexStoreState.isEmpty()) {
       this.metrics.ledger().syncTransactionsProcessed().inc(extensionTransactionCount);
```
dhedey, line 429: Ahh as I imagine you've noticed - this looks to be the wrong way around

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreState.java
```diff
@@ -89,25 +89,18 @@
...
 @Immutable
 public final class VertexStoreState {
   private static final Logger logger = LogManager.getLogger();
 
   private final VertexWithHash root;
   private final HighQC highQC;
-  // TODO: collapse the following two
-  private final ImmutableList<VertexWithHash> vertices;
-  private final ImmutableMap<HashCode, VertexWithHash> idToVertex;
+  private final ImmutableSet<VertexWithHash> vertices;
```
dhedey, line 98: Perhaps we should adjust the `hashCode()` implementation of `VertexWithHash` as it's currently quite slow. Perhaps its hashcode and equal implementation should just use the `vertexHash`, if we assume it's trustworthy?

```java
  @Override
  public int hashCode() {
    return Objects.hash(this.vertexHash);
  }

  @Override
  public boolean equals(Object o) {
    if (o instanceof final VertexWithHash that) {
      return Objects.equals(this.vertexHash, that.vertexHash);
    }
    return false;
  }
```

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreState.java
```diff
@@ -89,25 +89,18 @@
...
 
   private VertexStoreState(
-      HighQC highQC,
-      VertexWithHash root,
-      ImmutableMap<HashCode, VertexWithHash> idToVertex,
-      ImmutableList<VertexWithHash> vertices) {
+      HighQC highQC, VertexWithHash root, ImmutableSet<VertexWithHash> vertices) {
```
dhedey, line 101: It's not immediately clear to me without looking at the rest of the code if the `vertices` also contains the root? Perhaps we could tweak its name or add a comment?

EDIT: Looks like it's explicitly `nonRootVertices`, so perhaps we should call it that?

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreJavaImpl.java
```diff
@@ -207,44 +216,71 @@ public InsertQcResult insertQc(QuorumCertificate qc) {
       committedUpdate = Option.empty();
     }
 
+    metrics.bft().vertexStore().vertexCount().set(vertices.size());
```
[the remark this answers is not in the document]
dhedey: This still needs to be removed, as it's handled in `recordHighQcOrVertexChange`

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreJavaImpl.java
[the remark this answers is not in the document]
dhedey: Can we discuss this today?

I don't understand how node runners are meant to debug this / why we don't throw a fatal error here with clear instructions about what node runners are intended to do?

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreJavaImpl.java
```diff
@@ -302,91 +335,97 @@ public InsertVertexChainResult insertVertexChain(VertexChain vertexChain) {
...
       }
 
-      insertVertex(v).onPresent(bftInsertUpdates::add);
+      final var insertRes = insertVertexInternal(v);
+      if (insertRes.isSuccess()) {
+        bftInsertUpdates.add(insertRes.unwrap());
+      } else {
+        switch (insertRes.unwrapError()) {
+          case ALREADY_PRESENT, PREPARE_FAILED -> {
+            // No-op, continue iterating the vertices
```
dhedey, line 344: Why is `PREPARE_FAILED` a no-op rather than a fatal error? Do we expect prepare to ever fail here?

* Is this vertex chain untrusted, so preparation failures are expected? In which case, it seems weird we handle this by continuing to iterate the vertices?
* Or has quorum signed this? In which case, this would be a divergence / (possible) byzantine quorum exception?

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreJavaImpl.java
```diff
@@ -408,6 +447,32 @@ actually allows to still be able to get a path (the issue was more likely when v
...
+  }
+
+  private VertexStoreState getState() {
+    // TODO: store list dynamically rather than recomputing
+    ImmutableSet.Builder<VertexWithHash> verticesBuilder = ImmutableSet.builder();
+    getChildrenVerticesList(this.rootVertex, verticesBuilder);
+    return VertexStoreState.create(this.highQC(), this.rootVertex, verticesBuilder.build(), hasher);
+  }
+
+  private void getChildrenVerticesList(
```
dhedey, line 467: `addChildrenRecursive` would be a much clearer name IMO :)

core/src/main/java/com/radixdlt/consensus/epoch/EpochsConsensusModule.java
```diff
@@ -457,15 +458,17 @@ private VertexStoreFactory vertexStoreFactory(
...
-      Hasher hasher) {
+      Hasher hasher,
+      Serialization serialization,
+      Metrics metrics,
+      VertexStoreConfig vertexStoreConfig) {
     return vertexStoreState ->
         new VertexStoreAdapter(
-            VertexStoreJavaImpl.create(vertexStoreState, ledger, hasher),
+            new VertexStoreJavaImpl(
+                ledger, hasher, serialization, metrics, vertexStoreConfig, vertexStoreState),
```
dhedey, line 469: Probably best to call this `initialVertexStoreState` here, and in the `BftConfiguration` class.

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreJavaImpl.java
```diff
@@ -95,47 +97,64 @@ public final class VertexStoreJavaImpl implements VertexStore {
...
 
   // These should never be null
   private VertexWithHash rootVertex;
   private HighQC highQC;
-
-  private VertexStoreJavaImpl(
-      Ledger ledger, Hasher hasher, VertexWithHash rootVertex, HighQC highQC) {
+  private int currentSerializedSizeBytes;
+
+  public VertexStoreJavaImpl(
```
dhedey, line 113: This is all much nicer / easier to follow now. The consolidation of logic in `resetToState` and `recordHighQcOrVertexChange`; and use of explicit error codes makes this much easier to understand and reason about.

dhedey (approved):
Approving subject to comments being addressed. Nice work!

core/src/main/java/com/radixdlt/ledger/StateComputerLedger.java
dhedey: (Fixed since my comment)

core/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
```diff
@@ -464,8 +465,8 @@ private void rebuildAndSyncQC(SyncState syncState) {
     // TODO: check if there are any vertices which haven't been local sync processed yet
     if (requiresLedgerSync(syncState)) {
       syncState.fetched.sort(Comparator.comparing(v -> v.vertex().getRound()));
-      ImmutableList<VertexWithHash> nonRootVertices =
-          syncState.fetched.stream().skip(1).collect(ImmutableList.toImmutableList());
+      ImmutableSet<VertexWithHash> nonRootVertices =
```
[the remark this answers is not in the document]
LukasGasior1: It doesn't look like there was a bug, but using a set just makes more sense.

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreAdapter.java
```diff
@@ -92,40 +92,56 @@ public final class VertexStoreAdapter {
...
     final var result = vertexStore.insertQc(qc);
-    if (result instanceof VertexStore.InsertQcResult.Inserted inserted) {
-      dispatchPostQcInsertionEvents(inserted);
+    switch (result) {
+      case VertexStore.InsertQcResult.Inserted inserted -> this.highQCUpdateDispatcher.dispatch(
+          new BFTHighQCUpdate(
+              inserted.newHighQc(),
+              inserted.committedUpdate().map(VertexStore.CommittedUpdate::committedVertices),
+              inserted.serializedVertexStoreState()));
+      default -> {} // no-op
```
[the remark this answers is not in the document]
LukasGasior1: yup :+1: it's actually two cases:
```
case VertexStore.InsertQcResult.Ignored ignored -> {} // No-op
case VertexStore.InsertQcResult.VertexIsMissing ignored -> {} // No-op
```

core/src/test-core/java/com/radixdlt/harness/predicates/NodePredicate.java
```diff
@@ -207,6 +208,14 @@ public static Predicate<Injector> bftAtOrOverRound(Round round) {
     return i -> i.getInstance(PacemakerState.class).highQC().getHighestRound().gte(round);
   }
 
+  public static Predicate<Injector> atOrOverRound(Round round) {
```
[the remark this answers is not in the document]
LukasGasior1: I _think_ I might have added that to support tests that use epochs, where `i.getInstance(PacemakerState.class)`-based `bftAtOrOverRound` doesn't work. But my test doesn't really use epochs anyway, so I can just use the existing method. Perhaps we should rename it to something like: `nonEpochedBftAtOrOverRound`.

core/src/main/java/com/radixdlt/consensus/vertexstore/VertexStoreState.java
LukasGasior1: yup, sounds good :+1: I believe it is trustworthy

dhedey (approved):

PR #862, opened 2024-03-06
Add projected_enactment_at_start_of_epoch to pending protocol update in health API

LukasGasior1:
Adds (all optional):
- `projected_fulfillment_at_start_of_epoch` and `projected_fulfillment_timestamp` for each threshold of a pending protocol update
- `projected_enactment_at_start_of_epoch` and `projected_enactment_timestamp` to a pending protocol update, which are the values of the threshold which is projected to fulfill earliest

Notes:
- if the threshold is not met then its values are null
- if projected epoch is out of min/max bounds then the result is also null


core/src/main/java/com/radixdlt/monitoring/LedgerSummary.java
```diff
@@ -0,0 +1,76 @@
...
+ * permissions under this License.
+ */
+
+package com.radixdlt.monitoring;
+
+import com.radixdlt.statecomputer.ProtocolState;
+import com.radixdlt.statecomputer.commit.ActiveValidatorInfo;
+import java.util.Set;
+
+public record LedgerSummary(
```
jakrawcz-rdx, line 71: This does not "summarize ledger" (a broad concept), but it looks very specific to the protocol update and its tracking/forecasting, and hence deserves longer name :p
(I mean: let's keep this nice, concise "LedgerSummary" for a more appropriate occasion)
"ProtocolUpdateTrackingContext"?

core/src/test/java/com/radixdlt/api/system/HealthHandlerTest.java
```diff
@@ -64,17 +64,97 @@
...
+  // Test assertion parameters
+  private static final long EXPECTED_PROJECTED_ENACTMENT_EPOCH = 116;
+  private static final long EXPECTED_PROJECTED_ENACTMENT_TIMESTAMP =
+      EPOCH_EFFECTIVE_START_MS + (EPOCH_TARGET_DURATION_MS * 16);
+
   @Inject private HealthHandler sut;
 
+  public HealthHandlerTest() {
+    super(
+        new AbstractModule() {
```
jakrawcz-rdx, line 115: The `Modules.override()` and the mocks here look rather cumbersome for testing :(
At the same time, I agree that the "forecast" calculations are worth testing.

How about you extract the calculations out from `*Handler`, into a specialized `*Forecaster` class, and cover it with a unit test instead?

core/src/main/java/com/radixdlt/monitoring/LedgerSummary.java
dhedey: Agreed with `LedgerSummary` not being a great name. Maybe `CurrentEpochParameters`, as I see it's used in the `MetricInstaller`?

core/src/main/java/com/radixdlt/api/system/routes/HealthHandler.java
```diff
@@ -122,24 +130,85 @@ public HealthResponse handleRequest() {
...
+      LedgerSummary ledgerSummary) {
+    final var currentEpoch = ledgerSummary.currentEpoch();
+    final var epochStartTimeCalculator =
+        new EpochStartTimeCalculator(
+            ledgerSummary.consensusManagerConfigEpochTargetDurationMs(),
+            currentEpoch,
+            ledgerSummary.consenusManagerStateEpochEffectiveStartMs());
+
+    return switch (pendingProtocolUpdate.protocolUpdateTrigger().enactmentCondition()) {
+      case EnactAtStartOfEpochIfValidatorsReady condition -> {
```
dhedey, line 150: To make it easier to verify line-by-line, I'd rather this were called `atStartOfEpochCondition`

core/src/main/java/com/radixdlt/api/system/routes/HealthHandler.java
```diff
@@ -122,24 +130,85 @@ public HealthResponse handleRequest() {
...
+                              thresholdState.last().consecutiveStartedEpochsOfSupport().toLong();
+                          if (consecutiveStartedEpochsOfSupport <= 0) {
+                            return Option.<Long>empty();
+                          } else {
+                            final var requiredConsecutiveCompletedEpochsOfSupport =
+                                thresholdState
+                                    .first()
+                                    .requiredConsecutiveCompletedEpochsOfSupport()
+                                    .toLong();
+                            final var projectedFulfillmentEpoch =
```
dhedey, line 169: To make it easier to verify line-by-line, I'd prefer this to be called `projectedFulfillmentAtStartOfEpoch`

dhedey (approved):
Really nice, thanks 👍. Will save us second-guessing ourselves when live.

core/src/main/java/com/radixdlt/monitoring/LedgerSummary.java
LukasGasior1: I might just go with `InMemorySystemInfoState`

PR #864, opened 2024-03-11
Refactor: Group a few structs together into a `CommitBundleBuilder`

jakrawcz-rdx:
## Summary

I spotted a refactoring opportunity (a fairly old TODO) while working on https://radixdlt.atlassian.net/browse/LZ-22.
It is useful to address it at this moment, since I am about to add another piece of "data inserted on commit".

## Details

As in the PR title: I only move a few repeated structs into a single re-used struct.
During the development here, I tried extending the scope a bit (e.g. pulling the `proposer_timestamp` into the "tracked state" for convenience), but the complexity suddenly exploded... and it started touching on some conceptual questions (e.g. "is a `proposer_timestamp` a part of transaction execution result?").

## Testing

This is a pure refactoring, so all existing tests just need to pass.

### DIFF
--- a/core-rust/state-manager/src/jni/test_state_reader.rs
+++ b/core-rust/state-manager/src/jni/test_state_reader.rs
@@ -271,7 +271,7 @@ extern "system" fn Java_com_radixdlt_testutil_TestStateReader_leastStaleStateHas
     jni_sbor_coded_call(&env, request_payload, |_: ()| -> u64 {
         let database = JNINodeRustEnvironment::get_database(&env, j_rust_global_context);
         let least_stale_state_version = database
-            .access_direct() // the `get_stale_tree_parts_iter()` is inside a trait requiring writeability
+            .lock() // the `get_stale_tree_parts_iter()` is inside a trait requiring writeability
             .get_stale_tree_parts_iter()
             .next()
             .map(|(state_version, _)| state_version)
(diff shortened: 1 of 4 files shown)


core-rust/state-manager/src/jni/test_state_reader.rs
```diff
@@ -271,7 +271,7 @@ extern "system" fn Java_com_radixdlt_testutil_TestStateReader_leastStaleStateHas
     jni_sbor_coded_call(&env, request_payload, |_: ()| -> u64 {
         let database = JNINodeRustEnvironment::get_database(&env, j_rust_global_context);
         let least_stale_state_version = database
-            .access_direct() // the `get_stale_tree_parts_iter()` is inside a trait requiring writeability
+            .lock() // the `get_stale_tree_parts_iter()` is inside a trait requiring writeability
```
jakrawcz-rdx, line 274: _(note from author)_

The `StateHashTreeGcTest` that used this test-util became flakey (at least on my machine). I suspect it was caused by some unfortunate timings after I moved some code around (the test-util could not measure the exactly right version difference while commit kept pushing `state_version` forward).

LukasGasior1: :+1: I also noticed it became flakey

LukasGasior1 (approved):
Nice! :+1:

PR #867, opened 2024-03-12
LZ-22: Store historical values on commit

jakrawcz-rdx:
## Summary

Addresses https://radixdlt.atlassian.net/browse/LZ-22.

## Details

The changes are easier than expected:
- The new substate JMT leaf keys are collected during execution (but associated to DB substate keys only, to save memory).
- The `JMT key -> Substate key` association is turned into `JMT key -> Substate value` (using `DatabaseUpdates`) during commit, and persisted to DB.

## Testing

Similar to https://github.com/radixdlt/babylon-node/pull/863: this is hard to test without any other piece of history-supported feature implemented. But another PR which does historical reads is coming soon, and it will perfectly test the entirety.

The feature is disabled by default Node flags.

### DIFF
--- a/core-rust/Cargo.lock
+++ b/core-rust/Cargo.lock
@@ -209,7 +209,7 @@ dependencies = [
 [[package]]
 name = "blueprint-schema-init"
 version = "1.2.0-dev"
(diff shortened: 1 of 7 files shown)


dhedey (approved):
Nice :+1:

PR #871, opened 2024-03-14
Add DB checkpoint System API

LukasGasior1:
## Summary

This adds a System API that allows to take a persistent snapshot (checkpoint) of a database, without the need to shut the node down. This can come handy in a few scenarios, e.g.:
- debugging (with both read and write on top of a snapshot!)
- taking backups
- well, I guess that's it really :smile: 

## Details

This feature uses RocksDB's build in checkpoint feature (https://rocksdb.org/blog/2015/11/10/use-checkpoints-for-efficient-snapshots.html).
I've tested it on my local mainnet full node and it takes almost no time (maybe a second or so) to create a checkpoint.
It's really only meant to be used on the same filesystem, so that the sst files can be hardlinked without copying.
I haven't tested it in a scenario where copying would be necessary, but I imagine this would take a considerable amount of time and the request would eventually time out.

## Testing

I've tested it manually on a my local mainnet node.
Included is an automated test that does a basic sanity check: that the checkpoint directory is created and contains some sst files, when the API is called.


core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -715,7 +748,8 @@ pub struct StateManagerDatabase<R> {
 
 impl ActualStateManagerDatabase {
     pub fn new(
-        root: PathBuf,
+        root_path: PathBuf,
+        checkpoints_path: PathBuf,
```
jakrawcz-rdx, line 752: Are there any cons of simply accepting a full `target_checkpoint_path` in each `/db-checkpoint` request?

Regarding pros that I see:
- higher flexibility for the caller
- less code impact (e.g. if someone uses `new_read_only()` constructor and does not care about snapshots, let's not bother them with the new `checkpoints_path: PathBuf` param)

LukasGasior1: Yeah I considered that too. I went with a configurable folder for two reasons:
- IMO it's slightly easier to use
- and it's slightly more secure / less error prone (e.g. it can prevent unthoughtful calls to create a snapshot on an external drive)

both with the emphasis on "slightly".
Is it worth the price of reduced flexibility? I can be convinced that it isn't :man_shrugging:

jakrawcz-rdx: I see these convenience aspects, but I assume that a target user here is an above-average admin (hey - someone who DOES backups, right? :D), so they will very likely integrate via some script and they will also test their paths.
So I'd still personally pick the "less impact on the unrelated code" way.

LukasGasior1: or we could keep a single configured directory, but have it managed by the API handler itself, and provided to state manager in each call, wdyt?

jakrawcz-rdx: nice, best of both worlds!

LukasGasior1: _(done)_

core/src/main/java/com/radixdlt/api/system/routes/DbCheckpointHandler.java
```diff
@@ -0,0 +1,119 @@
...
+  DbCheckpointHandler(
+      RustDbCheckpoints rustCheckpoints,
+      RuntimeProperties runtimeProperties,
+      @NodeStorageLocation String nodeStorageLocation) {
+    this.enabled = runtimeProperties.get("api.system.enable_db_checkpoint", false);
+
+    this.checkpointsPath =
+        runtimeProperties.get(
+            "db.checkpoints_path",
+            Path.of(nodeStorageLocation, "state_manager_checkpoints").toString());
```
jakrawcz-rdx, line 94: _(aesthetics-only)_ looks like our usual approach is "handle the system props in the Guice modules + pass `*Config` objects to the logic".

core/src/main/java/com/radixdlt/api/system/system-api-schema.yaml
```diff
@@ -134,6 +134,36 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+  "/db-checkpoint":
+    post:
+      summary: Create a persistent snapshot (checkpoint) of the state manager database
```
jakrawcz-rdx, line 139: I think we need to keep `summary` short, down to "Three Capitalized Words", like the other ones:
image
(tools often want to render it as a list of one-liners)

"Create DB Checkpoint"?

jakrawcz-rdx (approved):
nice, only minor whinings left 👍

core/src/main/java/com/radixdlt/api/system/routes/DbCheckpointHandler.java
LukasGasior1: good point, done

jakrawcz-rdx (approved):

PR #873, opened 2024-03-18
LZ-24: Add GC for historical state

jakrawcz-rdx:
## Summary

Addresses https://radixdlt.atlassian.net/browse/LZ-24 in a rather trivial way - totally as designed.

## Testing

Added an int-test which asserts on the history table's size.


core-rust/state-manager/src/jni/test_state_reader.rs
```diff
@@ -280,6 +281,22 @@ extern "system" fn Java_com_radixdlt_testutil_TestStateReader_leastStaleStateHas
...
+    env: JNIEnv,
+    _class: JClass,
+    j_rust_global_context: JObject,
+    request_payload: jbyteArray,
+) -> jbyteArray {
+    jni_sbor_coded_call(&env, request_payload, |_: ()| -> u64 {
+        let database = JNINodeRustEnvironment::get_database(&env, j_rust_global_context);
+        let count = database
+            .lock()
+            .count_entries("associated_state_hash_tree_values");
```
dhedey, line 295: Minor: Should probably use `&AssociatedStateHashTreeValuesCf::NAME` here

dhedey (approved):

core-rust/state-manager/src/jni/test_state_reader.rs
dhedey: There's also an argument we need to do some renames from state_hash_tree > state_tree, in line with the changes in Scrypto, but I'll leave that as optional :)

jakrawcz-rdx: hmm, that would require making those CF consts public, together with the entire `TypedCf` infra. Currently they are only known inside the `rocks_db.rs`.

This is only a test-util, which needs to reach into some DB internals.
Its genuinely hard to decide what is less ugly in this case:
A) make these internal things public
B) make the test-util aware of some name used by the internals

To me B) has less impact on production code, but let me know if you strongly prefer A).

jakrawcz-rdx: (with... limited enthusiasm, we settled for B)

PR #875, opened 2024-03-19
Add missed env vars for state tree GC's properties

jakrawcz-rdx:
## Summary

The entire scope is in the PR title.
This was simply missed earlier, while apparently useful for ephemeral networks' configuration (as reported by @les-sheppard).

### DIFF
--- a/docker/config/default.config.envsubst
+++ b/docker/config/default.config.envsubst
@@ -9,6 +9,9 @@ network.host_ip=$RADIXDLT_HOST_IP_ADDRESS
 network.p2p.seed_nodes=$RADIXDLT_NETWORK_SEEDS_REMOTE
 network.p2p.use_proxy_protocol=$RADIXDLT_NETWORK_USE_PROXY_PROTOCOL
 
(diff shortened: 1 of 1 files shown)


dhedey (approved):

docker/config/default.config.envsubst
```diff
@@ -9,6 +9,9 @@ network.host_ip=$RADIXDLT_HOST_IP_ADDRESS
 network.p2p.seed_nodes=$RADIXDLT_NETWORK_SEEDS_REMOTE
 network.p2p.use_proxy_protocol=$RADIXDLT_NETWORK_USE_PROXY_PROTOCOL
 
+state_hash_tree.gc.interval_sec=$RADIXDLT_STATE_TREE_GC_INTERVAL_SEC
```
dhedey, line 12: Should this be `state_tree` now on the LHS?

jakrawcz-rdx: Yes - there will be a PR which pulls the new scrypto dep (with the rename there) and applies the rename to existing symbols in Node, including this one.

PR #877, opened 2024-03-20
**DO NOT MERGE - TESTING ONLY Node 613 update gradle radix version

les-sheppard:
Testing the passing of environment variable to docker build.


.github/workflows/docker.yml
```diff
@@ -100,10 +100,41 @@ jobs:
...
+        with:
+          fetch-depth: 0
+      - id: setup_version_properties
+        run: |
+          echo "VERSION_BRANCH=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_ENV
+          echo "VERSION_BUILD=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
+          echo "VERSION_COMMIT=${GITHUB_SHA}" >> $GITHUB_ENV
+          echo "VERSION_LAST_TAG=$(git tag --sort=-creatordate | head -n 1)" >> $GITHUB_ENV
+          if [[ $(git describe --exact-match --tag 2>&1) != *"no tag exactly matches"* ]]; then 
+            echo "VERSION_TAG=$(git describe --exact-match --tag)" >> $GITHUB_ENV
```
dhedey, line 117: As an opportunity to improve the behaviour -- this caused us quite frustrating behaviour where an old tag name was used; if we had multiple tags against the same commit.

Is there any way to read it from the triggering tag / ref instead?

e.g. If the job is caused by the creation of a release, we use that tag name, if not, it's empty?

build.gradle
```diff
@@ -121,17 +102,15 @@ def radixVersion() {
     // Ensure version is safe for gradle's purposes
     // (we do further sanitizing for debian's purposes in debSafeVersionName())
     version = version.replaceAll('/', '~')
+
+    print "radixdlt-version:$version"
```
dhedey, line 106: Out of interest, what is this print statement for?

build.gradle
```diff
@@ -80,39 +79,21 @@ define default value for the rust binary build type
...
-        if (details.branchName == null) {
-            version = "detached-head-${details.gitHash}"
+        if (versionBranchEnvVar == null) {
+            version = "detached-head-$versionTagEnvVar"
         } else {
-            version = "${details.branchName}-${details.gitHash}"
+            version = "$versionBranchEnvVar-$versionBuildEnvVar"
         }
         // Prepend SNAPSHOT so the tag name is valid regardless of the version
         version = "SNAPSHOT-${version}"
```
dhedey, line 99: If `version` is empty (e.g. locally) we probably want to not use `SNAPSHOT`. Perhaps `SNAPSHOT-development` or something might be okay?

core/build.gradle
```diff
@@ -167,16 +165,34 @@ task runAllIntegrationTests(type: Test) {
...
     mustRunAfter(processResources)
-    // Path to the file to be written
-    file = new File("$buildDir/resources/main", 'version.properties')
+    doLast {
+        File resourceFolder = new File("$buildDir/resources/main")
+        if (!resourceFolder.exists()) {
+            resourceFolder.mkdirs()
+        }
+
+        def versionFile = new File("$resourceFolder", 'version.properties').text = """
```
dhedey, line 176: Nice! :)

build.gradle
```diff
@@ -67,7 +67,6 @@ plugins {
     id 'nebula.ospackage' version '8.4.1'
     id 'org.sonarqube' version '3.3'
     id 'com.adarshr.test-logger' version '2.1.0'
-    id 'com.palantir.git-version' version '0.12.3'
     id 'com.moowork.node' version '1.3.1'
     id 'net.nemerosa.versioning' version '3.0.0'
```
dhedey, line 71: Can we also remove `net.nemerosa.versioning` like in the other file?

.github/workflows/docker.yml
dhedey: We'd then need to tweak `calculateVersionString` in `ApplicationVersion.java` to return `details.get("tag")` if it exists, instead of using this `cleanTag` stuff, and possibly update `ApplicationVersionTest.java`.

.github/actions/get-version/action.yml
```diff
@@ -18,10 +18,7 @@ runs:
       run: |
         IS_RELEASE="${{ inputs.is_release }}"
         echo "IS_RELEASE $IS_RELEASE"
-        if [[ "$IS_RELEASE" == "false" ]] ;then
-          export CI_VERSION=true
-        fi
-        version=$(./gradlew radixCiVersion | grep radixdlt-version | cut -d: -f2)
+        version=$(./gradlew getRadixVersion | grep radixdlt-version | cut -d: -f2)
```
dhedey, line 21: I wonder if we should just make it `printRadixVersion` and get rid of the `radixdlt-version:`prefix?

dhedey: I imagine we need to pass environment variables to this script? Similar to `setup_version_properties`?

dhedey: But also - I wonder if we actually need this action at all any more?  Happy to keep it or remove it, depending on what's most convenient.

.github/actions/gradle-task/action.yml
```diff
@@ -15,9 +15,8 @@ runs:
       run: |
         if [[ ${{ inputs.is_release }} == "false"  ]];then
           echo "Workflow triggered by push to a branch"
-          export CI_VERSION=true
         fi
-        version=$(./gradlew radixCiVersion | grep radixdlt-version | cut -d: -f2)
+        version=$(./gradlew getRadixVersion | grep radixdlt-version | cut -d: -f2)
```
dhedey, line 19: Ditto regarding I imagine that we need to pass environment variables to this script? Similar to setup_version_properties?

dhedey (commented):
Thanks Les. A mix of a few comments. I might well be wrong on some of these!

build.gradle
dhedey: I'm not actually sure what the impact might be locally if we end up building the same version multiple times? Possibly not a big deal, or it might break some build things... I guess we'll find out.

.github/workflows/docker.yml
les-sheppard: Yep, I can look at this. I initially just wanted to get a like-for-like solution but happy to make improvements :)

build.gradle
les-sheppard: That's used in [this](https://github.com/radixdlt/babylon-node/pull/877/files/632afb4484e0e810162e5e2aceb7ceeaba410b58#diff-728762f38670ff7bf72d3b2e830bd93839f109400cb2506413596d2aa3d04b93) action. But you suggested we might not need the action at all. So I'll look into it to see what it's used for.

build.gradle
les-sheppard: Oops, missed that one.

.github/actions/get-version/action.yml
les-sheppard: Will investigate the action. I just went for a dumb like-for-like replacement and didn't look too deeply into what the actions did/

.github/workflows/docker.yml
les-sheppard: I've updated the github actions to use the tag that triggered a run, rather than query github for the latest tag, etc.
I've removed the `cleanTag` stuff from `calculateVersionString`.

build.gradle
les-sheppard: That was for debug - it's gone now.

build.gradle
les-sheppard: Updated to use development, if the `versionBranchEnvVar` is empty

.github/actions/gradle-task/action.yml
les-sheppard: As above, replaced `getRadixVersion` with the version from the `setup-version-properties` action

core/build.gradle
les-sheppard: Thanks :)

.github/actions/get-version/action.yml
les-sheppard: I kept the action, but removed `printRadixVersion` from the gradle file and get the version from the `setup-version-properties` action

.github/actions/setup-version-properties/action.yaml
```diff
@@ -0,0 +1,42 @@
...
+    value: ${{ steps.setup_version_properties.outputs.VERSION_LAST_TAG }}
+runs:
+  using: "composite"
+  steps:
+    - id: setup_version_properties
+      run: |
+        if [[  $GITHUB_REF == *"refs/tags/"* ]]; then 
+          VERSION_TAG=$(echo $GITHUB_REF | cut -d'/' -f 3)
+          echo "VERSION_TAG=$VERSION_TAG" >> $GITHUB_OUTPUT
+          echo "VERSION_BRANCH=$VERSION_TAG" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
dhedey: Perhaps we first set `VERSION_COMMIT`, `VERSION_BUILD`, `VERSION_LAST_TAG`, and then the `if/else` sets either `VERSION_TAG` & `VERSION_DISPLAY` or `VERSION_BRANCH` & `VERSION_DISPLAY`?

dhedey (approved):
Really nice! 👍. I imagine it's taken a lot of careful iteration to get to this point!

It's looking so simple and easy to follow now.

One or two small questions (particularly related to the native build), and a few unimportant neatenings, but happy to approve pending clarification on the native build.

.github/actions/setup-version-properties/action.yaml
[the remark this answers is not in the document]
les-sheppard: Updated as suggested

core/src/main/java/com/radixdlt/monitoring/ApplicationVersion.java
```diff
@@ -117,8 +117,8 @@ private static ApplicationVersion loadFromResource(String resourceName) {
 
   @VisibleForTesting
   static String calculateVersionString(Map<String, String> details) {
-    if (isCleanTag(details)) {
-      return lastTag(details);
+    if (isTagVersion(details)) {
```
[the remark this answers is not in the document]
les-sheppard: Made the suggested updates and added comments to reference where the version.properties file is populated.

.github/workflows/docker.yml
```diff
@@ -100,10 +100,29 @@ jobs:
       - id: setup_tags
         run: echo "tag=$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
 
+  setup_version_properties:
```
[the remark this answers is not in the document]
les-sheppard: Good spot - just added it and building the debian package to ensure it gets the version information

.github/actions/setup-version-properties/action.yaml
```diff
@@ -0,0 +1,42 @@
...
+  VERSION_LAST_TAG:
+    description:
+    value: ${{ steps.setup_version_properties.outputs.VERSION_LAST_TAG }}
+runs:
+  using: "composite"
+  steps:
+    - id: setup_version_properties
+      run: |
+        if [[  $GITHUB_REF == *"refs/tags/"* ]]; then 
+          VERSION_TAG=$(echo $GITHUB_REF | cut -d'/' -f 3)
```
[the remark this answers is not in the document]
les-sheppard: Done :)

dhedey (commented):
Awesome! Thanks Les.

The last thing though is that it should be branched off of and merged into `release/anemone`, and from there into `main` and then `develop` - according to the [branching strategy doc](https://github.com/radixdlt/babylon-node/blob/main/docs/branching-strategy.md).

This ensures that we can build any hotfixes we need to 👍

PR #882, opened 2024-03-25
Support `at_ledger_state` in /entity/info endpoint

jakrawcz-rdx:
## Summary

This adds the history support (see https://radixdlt.atlassian.net/browse/NODE-614) in the first example endpoint.
Many other similar (or simpler) PRs, for other endpoints, will follow.

## Details

The infra is re-used (sometimes copied, e.g. for independence of OpenAPI spec) from Core API.

## Testing

A test case was added to the existing tests of this endpoint.

PR #885, opened 2024-03-26
Support `at_ledger_state` in /entity/iterator endpoint

jakrawcz-rdx:
⚠️ This merges into https://github.com/radixdlt/babylon-node/pull/882

## Summary

The next PR in the "state history support for Engine State API" epic (see https://radixdlt.atlassian.net/browse/NODE-614).
Many PRs for remaining endpoints will follow.

## Details

The Entity iteration is a bit special, since it reads from the Node's own index (i.e. does not even touch the JMT-backed SubstateDatabase). See the rustdocs and code comments which explain how the "historical view" is achieved there.

## Testing

Added a parameterized HTTP-level test for a couple of cases.


core-rust/state-manager/src/store/historical_state.rs
```diff
@@ -259,6 +259,19 @@ impl<'s, R: ReadableRocks> VersionScopedSubstateDatabase<'s, StateManagerDatabas
     }
 }
 
+/// Validates the requested historical state version and returns it (if present), or returns the
+/// current state version otherwise.
+///
+/// Implementation note:
+/// We re-use the validation performed by the [`VersionScopedSubstateDatabase::new()`] (which means
+/// throwing away a constructed instance).
+pub fn resolve_effective_state_version<R: ReadableRocks>(
```
LukasGasior1, line 268: I'm not a big fan of these methods (this one and `VersionScopedSubstateDatabase::new`) taking an optional state version. This seems more of a handler concern. Wdyt about moving it to a helper method somewhere in `node-http-apis` (and only have two separate methods in `store` e.g. `new_latest()` and `new_at_state_version(non_optional_state_version)`). Not a big deal though :)

LukasGasior1 (approved):

core-rust/state-manager/src/store/historical_state.rs
jakrawcz-rdx: so this particular method gets replaced by https://github.com/radixdlt/babylon-node/pull/889/files#diff-c85ce3e1172449a43cccc5b24780b578fc136f1afc88826fcff465ac5b475a56R235 - does it look "less ugly"?

I agree that e.g. routing to current vs historical `store` construction is the handler's responsibility, but it is kinda related to determining whether it is an available past version, etc (which reaches out to DB). So it is just convenient to do near the DB...

PR #888, opened 2024-03-26
Cleanups around error-surfacing in Engine State API

jakrawcz-rdx:
⚠️ This merges into https://github.com/radixdlt/babylon-node/pull/887

## Summary

One of the little refactors/cleanups before continuing with https://radixdlt.atlassian.net/browse/NODE-614.
(another one will follow-up with some "versioned DB infra" enhancements)

## Details

I only made some errors more structured (e.g. what state version is the earliest available for history), and easier to surface (e.g. more automated "this Node flag is not enabled" error).
I also updated the apidocs to indicate the possible HTTP 409 from every endpoint which supports history (the "Conflicting Request" is the closest thing I found on the internet [for a request misaligned with server's config]).

## Testing

Some asserts about errors in existing tests were improved.


LukasGasior1 (approved):
LGTM!

PR #889, opened 2024-03-27
Refactoring: Rely only on version-scoped database in relevant endpoints

jakrawcz-rdx:
⚠️ This merges into https://github.com/radixdlt/babylon-node/pull/888

## Summary

After a couple of initial PRs in the "state history in Engine State API" epic (see https://radixdlt.atlassian.net/browse/NODE-614), Łukasz has suggested that it would be clearer and less error-prone to handle all database operations through a version-scoped wrapper.

## Details

This refactoring actually offered extra, unexpected simplifications, e.g. in the implementation of `/entity/iterator` (no need to manually implement the "version filter").

## Testing

This is a pure refactoring, the regression tests just need to pass.

### DIFF
--- a/core-rust/state-manager/src/store/historical_state.rs
+++ b/core-rust/state-manager/src/store/historical_state.rs
@@ -62,50 +62,72 @@
  * permissions under this License.
  */
 
+use std::ops::Deref;
 use substate_store_impls::state_tree::entity_tier::EntityTier;
 
 use crate::engine_prelude::*;
 use crate::store::traits::*;
-use crate::{ReadableRocks, StateManagerDatabase, StateVersion};
+use crate::traits::indices::{CreationId, EntityBlueprintId, ReNodeListingIndex};
+use crate::{LedgerHeader, LedgerProof, ReadableRocks, StateManagerDatabase, StateVersion};
 
 /// An implementation of a [`SubstateDatabase`] viewed at a specific [`StateVersion`].
 ///
 /// This database is backed by:
 /// - a [`ReadableTreeStore`] - a versioned source of ReNodes / Partitions / Substates metadata,
 /// - and a [`LeafSubstateValueStore`] - a store of Substate values' associated with their leafs.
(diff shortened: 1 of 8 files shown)


core-rust/state-manager/src/store/historical_state.rs
```diff
@@ -237,54 +263,48 @@ impl<'s, R: ReadableRocks> VersionScopedSubstateDatabase<'s, StateManagerDatabas
...
                 current_version, ..
             } => *current_version,
-            VersionScopedSubstateDatabase::Historical(database) => database.at_state_version(),
+            VersionScopedDatabase::Historical(database) => database.at_state_version,
         }
     }
 
-    fn current(database: &'s StateManagerDatabase<R>, current_version: StateVersion) -> Self {
+    /// Returns the header of the proof proving the scoped point on the ledger.
+    pub fn proving_ledger_header(&self) -> LedgerHeader {
```
LukasGasior1, line 274: A bit of naming nitpicking: this is really the _latest_ ~proof~ header, as seen by the current scope, so maybe just: `latest_ledger_header`? One might argue that this can be a bit ambiguous when reading the code (because one has to be aware of the fact that the scoped db is being used), but the same argument could apply to `get_substate` - yet we haven't introduced a separate `get_scoped_substate` (and I'm not suggesting that we do).

LukasGasior1 (approved):
Just some naming nitpicking, LGTM!

core-rust/state-manager/src/store/historical_state.rs
LukasGasior1: ah, nvm, this can also be a future proof (from the perspective of the scoped db)

LukasGasior1: `proving_ledger_header` is fine then :+1:

LukasGasior1: For the record, as discussed offline, it might turn out that this will no longer be needed in its current form.

PR #891, opened 2024-03-29
Refactoring: Update the historical state metadata from the JMT GC

jakrawcz-rdx:
## Summary

Addresses https://radixdlt.atlassian.net/browse/NODE-615.

This refactoring is not a "definitive improvement"; I think it choses another compromise:
- **Before this PR:** The historical state reader has to _read_ from the "internal" `StaleStateTreePartsCf` normally maintained by by JMT logic and JMT GC.
- **After this PR:** The JMT GC has to _write_ to the "internal" `VersionedStateTreeAssociatedValuesStatus` normally maintained by the history state infra.

Either way, some conceptual boundary is violated.

(I am not sure what would be the 100% elegant solution; e.g. I know that the above dilemma could be avoided if the historical state reader relied on `let earliest_available_version = current_version - config.history_length`, but this gets cumbersome since `config.history_length` can be reconfigured between boot-ups. So I think it's fine to choose one of the practical "compromises" above.)

## Testing

All regression tests pass fine.
I've also manually run the tests at the tip of Engine State API's branch (where state history infra is used extensively).


core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -1842,15 +1828,21 @@ impl<R: WriteableRocks> StateTreeGcStore for StateManagerDatabase<R> {
...
+    fn delete_stale_tree_parts_up_to_version(&self, state_version: StateVersion) {
         let db_context = self.open_rw_context();
-        let stale_tree_parts_cf = db_context.cf(StaleStateTreePartsCf);
-        for state_version in state_versions {
-            stale_tree_parts_cf.delete(state_version);
-        }
+        db_context
+            .cf(StaleStateTreePartsCf)
+            .delete_range(&StateVersion::pre_genesis(), &state_version);
+        let updated_status = StateTreeAssociatedValuesStatusV1 {
```
dhedey, line 1836: It's a little weird to me that we combine the deletion of the stale tree notifications with the updating of the status.

In fact, I think this is invalid and would cause a panic, because we update `VersionedStateTreeAssociatedValuesStatus` too late after we've already deleted some substate values.

The updating of the status should be tied to the presence of the substate values - so it would need to happen in the `batch_delete_node` method - we should call a `invalidate_historical_substate_value_availability_up_to(max_version_deleted)` or something - and the `invalidate_historical_substate_value_availability_up_to` should load the current status, and take max of current status, and `invalid_up_to + 1`.

dhedey (changes requested):
I do prefer this, but we need to be a little more careful - this would currently cause panics because we delete from the index before we update the indexes' validity.

core-rust/state-manager/src/store/rocks_db.rs
jakrawcz-rdx: OMG, nice catch - yeah, that would be a panic every time someone accesses historical state during JMT GC...

I've re-written this part, with more rustdocs and comments - PTAL.

core-rust/state-manager/src/store/jmt_gc.rs
```diff
@@ -153,29 +163,38 @@ impl StateTreeGc {
...
         info!("Flushing the last buffer ({} deletes)", deleted_nodes.len());
-        database.batch_delete_node(deleted_nodes.iter());
-        database.batch_delete_stale_tree_part(deleted_state_versions.iter());
+        delete_nodes_and_update_state_history(database, &deleted_nodes, to_state_version);
     }
 }
 
+/// A helper method for flushing the GC's partial DB update batches.
+///
+/// Note: the order of the operations is important for non-trivial reasons - see comments.
```
dhedey, line 181: This is fine, although a part of me would rather we did this in one `rw_context`, I'm not sure we have a good reason not to, and that way we don't have to worry about ordering so much? :)

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -1828,6 +1814,35 @@ impl<R: WriteableRocks> StateTreeGcStore for StateManagerDatabase<R> {
...
+            });
+        let Some(current_available_from) = current_available_from else {
+            // The state history feature is simply not enabled.
+            return;
+        };
+        if available_from <= current_available_from {
+            // The state history feature was enabled after this state version.
+            return;
+        }
+        let updated_status =
```
dhedey, line 1836: Minor note - but this code would be slightly safer with respect to possible changes (e.g. new versions of status, or addition of other properties to the status) if we take:
```rust
        let mut status = extension_data_cf
            .get(&ExtensionsDataKey::StateTreeAssociatedValuesStatus)
            .map(|bytes| {
                scrypto_decode::<VersionedStateTreeAssociatedValuesStatus>(&bytes)
                    .unwrap()
                    .into_latest()
            });
// ...
        status.historical_substate_values_available_from = available_from;
        extension_data_cf.put(
            &ExtensionsDataKey::StateTreeAssociatedValuesStatus,
            &scrypto_encode(&VersionedStateTreeAssociatedValuesStatus::from(status)).unwrap(),
        );
    
```

dhedey (approved):
Nice. A couple of minor optional further suggestions

core-rust/state-manager/src/store/jmt_gc.rs
jakrawcz-rdx: yeah, but from previous comment:

> It's a little weird to me that we combine the deletion of the stale tree notifications with the updating of the status.

(which is also valid remark)

Doing this under one `rw_context` would combine 3 semi-related items into one `StateManagerDatabase`-level call. This is because we never exposed any notion of "transaction" (or "== DB write batch") outside `StateManagerDatabase`. And it looks like "the right" solution here would require exactly that non-trivial piece of infra.

Of the 2 imperfect alternatives, I prefer to stay with the current state in this PR - just for the sake of conventions that I see around:
- the methods in DB traits are quite atomic / single-responsibility;
- other long-running DB processes (e.g. backfills) also use the right ordering of DB operations (to be resilient to Node restarts)

(ofc my opinion above assumes that the current impl does avoid that previous pitfall - if we find any problem with it, it will be clear signal that "yeah we better put that under one `do_these_3_things()` DB call")

core-rust/state-manager/src/store/rocks_db.rs
jakrawcz-rdx: good idea, done 👍

PR #900, opened 2024-04-05
fix: Add gradle lock files

dhedey:
## Summary

https://radixdlt.atlassian.net/browse/NODE-624

## Testing

Will check the tests still pass.


cli-tools/dependencies-tree.txt
```diff
@@ -0,0 +1,1439 @@
+
```
LukasGasior1, line 1: what's the use for these files?

LukasGasior1 (approved):
Better late than never :smile: LGTM

cli-tools/dependencies-tree.txt
dhedey: Mostly as a way to track down _why_ a dependency was included in the lock file -- as the default gradle lock files don't include this information.

PR #901, opened 2024-04-08
Refactor: some cleanups after mempool reevalution tweaks

jakrawcz-rdx:
## Summary

This PR addresses some postponed pure-refactor comments from https://github.com/radixdlt/babylon-node/pull/681.
It is supposed to conclude https://radixdlt.atlassian.net/browse/NODE-617 for now.

## Details

- Removed some now-noop delete calls.
- Removed `Arc<>`s around `MempoolData`.
- Extracted some boilerplate into `indexing.rs`.
- Made `AtSpecificState` more... specific.

## Testing

Just the regression tests need to pass.

### DIFF
--- a/core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
+++ b/core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
@@ -527,19 +527,19 @@ impl PendingTransactionResultCache {
             return record.clone();
         }
 
-        let new = PendingTransactionRecord::new(intent_hash, invalid_from_epoch, attempt);
+        let new_record = PendingTransactionRecord::new(intent_hash, invalid_from_epoch, attempt);
 
         // NB - removed is the item kicked out of the LRU cache if it's at capacity
         let removed = self
             .pending_transaction_records
-            .push(notarized_transaction_hash, new.clone());
+            .push(notarized_transaction_hash, new_record.clone());
 
         self.handled_added(intent_hash, notarized_transaction_hash);
-        if let Some((p, r)) = removed {
-            self.handled_removed(p, r);
+        if let Some((removed_notarized_transaction_hash, removed_record)) = removed {
+            self.handled_removed(removed_notarized_transaction_hash, removed_record);
         }
 
-        new
+        new_record
     }
 
     pub fn track_committed_transactions(
--- a/core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
+++ b/core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
@@ -3,7 +3,7 @@ use node_common::locks::RwLock;
 
 use crate::{
     transaction::{CheckMetadata, StaticValidation},
-    CommittedUserTransactionIdentifiers, MempoolAddRejection, StateVersion,
+    CommittedUserTransactionIdentifiers, MempoolAddRejection, StateVersion, TransactionTreeHash,
 };
 
 use crate::priority_mempool::PriorityMempool;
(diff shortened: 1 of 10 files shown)


core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
```diff
@@ -510,10 +512,10 @@ impl PendingTransactionResultCache {
         let mut write_mempool = self.mempool.write();
         if attempt.rejection.is_some() {
             write_mempool.remove_by_payload_hash(&notarized_transaction_hash);
-        } else if let Some(state_version) = attempt.against_state.specific_version() {
+        } else if let AtState::Specific(specific_state) = &attempt.against_state {
             write_mempool.update_transaction_executed_state_version(
                 &notarized_transaction_hash,
-                state_version,
+                specific_state.committed_version(),
```
dhedey, line 518: Optional thought - on reflection, I think this would be clearer as `mempool.observe_pending_execution_result(&notarized_transaction_hash, &attempt)` - and then the logic for "what should I do with this knowledge" moves into the mempool 👍

dhedey, line 518: Actually, I've just noticed that this code in `state_computer` `prepare` should be removed too, as it's now handled by `track_transaction_result` a few lines below:
```
        self.mempool_manager
            .remove_rejected(&pending_rejected_transactions);
```

...and we could then remove  `mempool_manager.remove_rejected` and `mempool.remove_by_payload_hash`

core-rust/state-manager/src/mempool/priority_mempool.rs
```diff
@@ -165,89 +149,72 @@ impl HasNotarizedTransactionHash for MempoolTransaction {
...
 
-/// A wrapper around an [`Arc<MempoolData>`] which implements ordering traits for the proposal priority.
-/// If a > b then a is prioritized over b.
+/// A transaction's proposal priority.
+///
+/// The greatest element (i.e. the one with the *highest* `tip_percentage`, and then the *oldest*
+/// `added_at`) marks the "best" transaction (i.e. the one to be executed first).
 #[derive(Clone, Eq, PartialEq)]
-pub struct MempoolDataProposalPriorityOrdering(pub Arc<MempoolData>);
+struct ProposalPriority {
```
dhedey, line 172: Nice name! 👍

core-rust/state-manager/src/mempool/priority_mempool.rs
```diff
@@ -430,79 +396,53 @@ impl PriorityMempool {
...
-        };
-
-        if !self
-            .executed_state_version_index
-            .remove(&MempoolDataStateVersion(state_version, data))
-        {
-            panic!("Mempool executed state version index out of sync on remove");
-        }
+        self.proposal_priority_index
+            .remove(ProposalPriority::new(&data), *notarized_transaction_hash);
```
dhedey, line 415: We should add an `expect(..)` to all of these to panic if for some reason the indices get out of line.

EDIT: Ahh I see this is inline on the `remove` method. All good! Perhaps `remove_expected_entry` might be clearer, but `remove` is fine too.

dhedey (approved):
Really nice! One or two suggestions you can implement if you agree

core-rust/state-manager/src/mempool/priority_mempool.rs
jakrawcz-rdx: image
... and this has to be accompanied by `insert_unique()` 👍

dhedey (approved):
Really nice :D

core-rust/state-manager/src/mempool/priority_mempool.rs
dhedey: Beautiful name suggestions :D

PR #903, opened 2024-04-10
Return synthetic header summary for a historical transaction preview

jakrawcz-rdx:
## Summary

Addresses https://radixdlt.atlassian.net/browse/NODE-626 **for the transaction preview**.
Another PR will follow, for Engine State API (and will re-use 95% of the infra introduced here).

## Details

Exactly as in the issue description.

## Testing

Added a test assert about the exact state version returned in the header summary.


LukasGasior1 (approved):
Nice!

PR #907, opened 2024-04-16
Execute Bottlenose state update

jakrawcz-rdx:
⚠️ This merges to https://github.com/radixdlt/babylon-node/pull/898.

## Summary

Addresses https://radixdlt.atlassian.net/browse/NODE-619.

## Details

Pretty much follows the DoD:
- Defines Bottlenose in `protocol_config`,
- Executes Bottlenose in `testnet_protocol_config` (at Epoch 4),
- [required adjustment] Support for new entities being created by flash transactions,
- [extra] Minor refactor of the protocol update definitions infra,

## Testing

Added tests asserting that:
- `AccountLocker` is not available before Bottlenose and is available after Bottlenose,
- The flash transactions executed by Bottlenose are correctly streamed by Core API.


core-rust/state-manager/src/protocol/protocol_updates/definitions/anemone_definition.rs
```diff
@@ -36,37 +28,7 @@ impl ProtocolUpdateDefinition for AnemoneProtocolUpdateDefinition {
         Box::new(BatchedUpdater::new(
             new_protocol_version.clone(),
             Self::state_computer_config(network_definition),
-            AnemoneBatchGenerator,
+            ScryptoEntriesBatchGenerator::new(&ANEMONE_ENTRIES),
```
dhedey, line 31: I appreciate the refactoring - but I'm not so keen on this `ProtocolUpdateEntry` formalism in the engine, and was going to refactor it next week to be more flexible / aligned with the batching flexibility we used to have in the node, and that we might need in a pinch in certain expedited protocol update scenarios.

Basically what I'm saying is I'm happy to keep this refactor for now, but it'll maybe need to be reverted in a week or two :D

core-rust/state-manager/src/transaction/executable_logic.rs
```diff
@@ -188,32 +188,22 @@ impl<'a, S: SubstateDatabase> TransactionLogic<S> for ConfiguredExecutable<'a> {
...
+                let system_structure =
+                    SystemStructure::resolve(store, &state_updates, &application_events);
+                let new_node_ids = collect_new_node_ids(&state_updates);
+                let state_update_summary =
+                    StateUpdateSummary::new(store, new_node_ids, &state_updates);
 
                 let commit_result = CommitResult {
                     state_updates,
-                    state_update_summary: Default::default(),
+                    state_update_summary,
```
dhedey, line 200: Thanks for fixing this!

I notice that we have actually done this properly now... I just want to triple check that the `state_update_summary` isn't used in the `LedgerTransactionReceipt` / for any of the hashes?

I think that's right, but just wanted to check 👍

core/src/test/java/com/radixdlt/rev2/protocol/AnemoneProtocolUpdateTest.java
```diff
@@ -205,96 +194,37 @@ public void test_get_current_time_second_precision() {
 
   @Test
   public void core_api_streams_anemone_flash_transactions() throws Exception {
-    final var coreApiPort = FreePortFinder.findFreeLocalPort();
-    try (var test = createTest(createCoreApiModule(coreApiPort))) {
+    final var coreApiHelper = new ProtocolUpdateTestUtils.CoreApiHelper();
+    try (var test = createTest(coreApiHelper.module())) {
```
dhedey, line 198: Arguably this might be easier as a Core API integration test, but this is fine too - thanks for pulling out the Core API helper to make this at least a little neater!

core/src/test/java/com/radixdlt/rev2/protocol/BottlenoseProtocolUpdateTest.java
```diff
@@ -0,0 +1,227 @@
...
+        new TransactionCallPreviewRequest()
+            .network(Network.INTEGRATIONTESTNET.getLogicalName())
+            .target(
+                new BlueprintFunctionTargetIdentifier()
+                    .packageAddress(
+                        addressing.encode(ScryptoConstants.ACCOUNT_LOCKER_PACKAGE_ADDRESS))
+                    .blueprintName("AccountLocker")
+                    .functionName("instantiate_simple")
+                    .type(TargetIdentifierType.FUNCTION))
+            .addArgumentsItem("4d0101"); // hex-encoded SBOR `true` (for `allow_recover` parameter)
```
dhedey, line 150: :D nice SBOR skillz! Side note, the TransactionCallPreviewRequest API could really do with an overhaul... I should raise a ticket for it at some point.

core-rust/state-manager/src/protocol/protocol_updates/definitions/bottlenose_definition.rs
```diff
@@ -0,0 +1,44 @@
...
+        "bottlenose-locker-package",
+        ProtocolUpdateEntry::LockerPackage,
+    ),
+    (
+        "bottlenose-account-try-deposit-or-refund",
+        ProtocolUpdateEntry::AccountTryDepositOrRefundBehaviorChanges,
+    ),
+    (
+        "bottlenose-protocol-params-to-state",
+        ProtocolUpdateEntry::ProtocolParamsToState,
```
dhedey, line 24: :D Can I put you in charge of regularly updating this branch; and also of double-checking that all the bottlenose updates are pulled through before we do the actual release?

core-rust/state-manager/src/transaction/executable_logic.rs
jakrawcz-rdx: double-checked now - it becomes a part of `LocalTransactionExecution` only (and is served from API), but does not go into any hashes 👍

core-rust/state-manager/src/protocol/protocol_updates/definitions/anemone_definition.rs
jakrawcz-rdx: I thought that `ProtocolUpdateEntry` (or rather, its surrounding infra) nicely groups the PU parts (that previously used to be "a bunch of `fn`s in neighbouring files". But if you want to refactor further, it is fine - as long as the "PU parts" (whatever representation you end up with) are easy to reference from Node.

(one part of that refactor _[for your consideration]_ could be to "actually name the transactions in Engine repo", which would simplify the Node's declarative part that you noticed at https://github.com/radixdlt/babylon-node/pull/907#discussion_r1568074902)

core-rust/state-manager/src/protocol/protocol_updates/definitions/bottlenose_definition.rs
jakrawcz-rdx: yup, I will take care of this 👍

core/src/test/java/com/radixdlt/rev2/protocol/BottlenoseProtocolUpdateTest.java
dhedey: > Is there an easy java-side Sbor.encode(true) that I could use here?
I think you be wanting `new ScryptoSbor(new CodecMap()).encode_payload(true)` - but I'm happy to leave it as `"4d0101"`.

core/src/test/java/com/radixdlt/rev2/protocol/AnemoneProtocolUpdateTest.java
dhedey: Totally agree :D. Good use of composition over inheritance!

PR #911, opened 2024-04-19
Refactor: Moving genesis scenarios from `GenesisData` to `StateManagerConfig`

jakrawcz-rdx:
## Summary

This is the first step towards https://radixdlt.atlassian.net/browse/NODE-618.
A pure refactoring - the entire scope is in the title.
The next PR will simply add `after_protocol_update_scenarios: Map<ProtocolUpdateName, List<ScenarioName>>`.

## Testing

A refactoring - just the existing tests must pass.


jakrawcz-rdx:
Closing this - as discussed with Łukasz, it is a bit cumbersome/risky to mutate the GenesisData's schema at this point. Maybe it will have to happen anyway (albeit in a different way) when we refactor "genesis == proto update".

But for now, the "proto updates' scenarios" will use a dedicated mechanism (not shared with genesis' scenarios), in another PR.

PR #912, opened 2024-04-19
Render `ScryptoInstant` with only a second-precision

jakrawcz-rdx:
## Summary

Addresses the https://radixdlt.atlassian.net/browse/NODE-632 for Core API.

## Testing

Adjusted the detailed unit test.


dhedey (approved):
👍

PR #914, opened 2024-04-19
Add a note about `core-rust` dependency of the rundocker script

klntsky:
## Summary

I was trying to run the `rundocker.sh` script and noticed this implicit dependency. I think it makes sense to make it explicit.


dhedey (approved):
Thanks [klntsky](https://github.com/klntsky) 👍

PR #915, opened 2024-04-22
Minor Engine State API tweaks

jakrawcz-rdx:
## Summary

Addresses most points from https://radixdlt.atlassian.net/browse/NODE-633 and from comments on https://github.com/radixdlt/babylon-node/pull/908.

## Details

Addressed:
- Merge latest develop into the branch.
- Configure Engine State API for testnet-node
- Change port to 3336.
- Change the API name to Engine State API - Babylon (Anemone), API logo to https://cdn.document360.io/50e78792-5410-4ac9-aa43-4612b4d33953/Images/Documentation/radix-logo-banner.png
- Revert as much of Cargo.lock as possible
- Propagate Instant rendering approach from Core API

Out of scope:
- Generate typescript API client code
  - Needs some discussion: the current TS client kinda acts as if Core API was the only API
- Set version to Beta + Explaining the implications in API docs
  - This will be a separate PR, for a different audience

## Testing

Mostly refactorings - the existing tests must pass.


core-rust/engine-state-api-server/src/engine_state_api/conversions/common.rs
```diff
@@ -39,7 +39,7 @@ pub fn to_api_ledger_header_summary(
             transaction_tree_hash: to_api_transaction_tree_hash(&hashes.transaction_root),
             receipt_tree_hash: to_api_receipt_tree_hash(&hashes.receipt_root),
         }),
-        proposer_timestamp: Box::new(to_api_instant_from_safe_timestamp(
+        proposer_timestamp: Box::new(to_api_consensus_instant(
```
dhedey, line 42: Can we tweak this to say `_from_timestamp_ms` to allow us to be able to verify each line for correctness?

core-rust/engine-state-api-server/src/engine_state_api/conversions/numerics.rs
```diff
@@ -116,39 +118,45 @@ pub fn to_api_i64_as_string(input: i64) -> String {
...
-            message: "Timestamp invalid when converted to DateTime".to_string(),
-        })?;
-
-    Ok(models::Instant {
-        unix_timestamp_ms: timestamp_millis,
-        date_time,
+) -> Result<models::ConsensusInstant, MappingError> {
+    let clamped_timestamp_millis =
+        timestamp_millis.clamp(MIN_API_TIMESTAMP_MS, MAX_API_TIMESTAMP_MS);
+    let date_time = NaiveDateTime::from_timestamp_millis(clamped_timestamp_millis)
```
dhedey, line 143: Considering we're starting again with the Engine State API schema and can fix our past mistakes, I'd rather we made `date_time` optional here, and didn't clamp the timestamp (so basically the same as `to_api_scrypto_instant`, except operating on `ms` instead of seconds).

core-rust/Cargo.lock
dhedey, line 1: Thanks for reverting these changes. Hopefully it wasn't too painful!

docker/network-size-1.yml
dhedey, line 1: Nice! Thanks for spotting/fixing these

dhedey (approved):
Looks great, thanks. A couple of quick comments

docker/network-size-1.yml
jakrawcz-rdx: there is 1x `TODO(resolve-during-review)` in `network-size-5.yml`, can you advise on it?

docker/network-size-5.yml
```diff
@@ -84,7 +88,7 @@ services:
       RADIXDLT_NODE_KEY: ${RADIXDLT_VALIDATOR_4_PRIVKEY}
     networks:
       - five_nodes
-    ports:
+    ports: # TODO(resolve-during-review): this `core4` looks weird compared to the rest; intentional?
```
dhedey, line 91: 99.9% not intentional. If you could fix it up as you do this, that would be grand :+1:

docker/network-size-1.yml
dhedey: Done:
> 99.9% not intentional. If you could fix it up as you do this, that would be grand :+1:

dhedey (approved):

PR #916, opened 2024-04-22
Update to Java 21

siy:


dhedey:
Closing as we're deciding not to update to Java 21 due to churn for node runners and migration to rust.

PR #917, opened 2024-04-22
Describe Engine State API's "beta" status in the apidocs

jakrawcz-rdx:
## Summary

Addresses item 4. of https://radixdlt.atlassian.net/browse/NODE-633, as in the title.

## Testing

N/A, it is only docs change.

### DIFF
--- a/core-rust/engine-state-api-server/engine-state-api-schema.yaml
+++ b/core-rust/engine-state-api-server/engine-state-api-schema.yaml
@@ -1,8 +1,7 @@
 openapi: 3.1.0
 
 info:
-  # Should be the release version, and the release display name
-  version: 'v0.0.1'
+  version: 'v0.1-beta'
   title: 'Engine State API - Babylon (Anemone)'
   license:
     name: The Radix License, Version 1.0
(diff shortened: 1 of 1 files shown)


core-rust/engine-state-api-server/engine-state-api-schema.yaml
```diff
@@ -11,6 +10,12 @@ info:
     url: https://cdn.document360.io/50e78792-5410-4ac9-aa43-4612b4d33953/Images/Documentation/radix-logo-banner.png
     href: https://docs.radixdlt.com
   description: |
+    *Beta Disclaimer*
+    Please note that this API is currently in its beta phase. It is subject to breaking changes
+    without any prior notice. It should not be used for business-critical integrations before the
+    `version` indicated above becomes stable, which is expected in Q4 of 2024 - please pay attention
+    to the Radix Node's future release notes.
+
```
dhedey, line 18: Perhaps the below? I've weakened the advisory a little. And removed "without prior notice" because that implies it's not been mentioned in the release notes.

What do you think?

```suggestion
    **This API is currently in Beta**
    This API interface may experience breaking changes as part of node releases. Such changes will be clearly mentioned in the [babylon-node release notes](https://github.com/radixdlt/babylon-node/releases).
    We advise against using this API for business-critical integrations before the `version` indicated above becomes stable, which is expected in Q4 of 2024.
    
```

dhedey (approved):

core-rust/engine-state-api-server/engine-state-api-schema.yaml
jakrawcz-rdx: very nice, thanks 👍 

let me only fix the "API interface" (which would unroll to non-compiling "application programming interface interface")

PR #918, opened 2024-04-22
Incorrect version in gradle task action

les-sheppard:


.github/actions/gradle-task/action.yml
```diff
@@ -19,7 +19,7 @@ runs:
         if [[ ${{ inputs.is_release }} == "false"  ]];then
           echo "Workflow triggered by push to a branch"
         fi
-        version=${{ steps.setup_version_properties.outputs.VERSION_DISPLAY }}
+        version=${{ steps.setup_version_properties.outputs.VERSION_TAG }}
         echo "Version: $version"
-        ./gradlew ${{ inputs.gradle_command }}
+        VERSION_TAG=$version ./gradlew ${{ inputs.gradle_command }}
```
dhedey, line 24: Out of interest, why don't we do this like our other jobs?

I imagine this set-up will be more safe, in case the job uses some other env too?
```
    env:
      VERSION_BRANCH: ${{ needs.setup_version_properties.outputs.VERSION_BRANCH }}
      VERSION_COMMIT: ${{ needs.setup_version_properties.outputs.VERSION_COMMIT }}
      VERSION_DISPLAY: ${{ needs.setup_version_properties.outputs.VERSION_DISPLAY }}
      VERSION_BUILD: ${{ needs.setup_version_properties.outputs.VERSION_BUILD }}
      VERSION_TAG: ${{ needs.setup_version_properties.outputs.VERSION_TAG }}
      VERSION_LAST_TAG: ${{ needs.setup_version_properties.outputs.VERSION_LAST_TAG }}
```

.github/actions/gradle-task/action.yml
```diff
@@ -19,7 +19,7 @@ runs:
         if [[ ${{ inputs.is_release }} == "false"  ]];then
           echo "Workflow triggered by push to a branch"
         fi
-        version=${{ steps.setup_version_properties.outputs.VERSION_DISPLAY }}
+        version=${{ steps.setup_version_properties.outputs.VERSION_TAG }}
```
dhedey, line 22: Looks like we might also need to update `actions/get-version` as well?

dhedey: Although to be honest, `get-version` is a dodgy abstraction now that we have the defualt way of doing it... 

Perhaps we should:
* Remove our use of `steps.get_version.outputs.radixdlt_version` to just be `VERSION_TAG` if that's possible? 
* Change `get-version` to `display-version` as I think it's just for display? And have it just print the tags:
```
VERSION_BRANCH
VERSION_COMMIT
VERSION_DISPLAY
VERSION_BUILD
VERSION_TAG
VERSION_LAST_TAG

DOCKER_TAGS = VERSION_TAG
```

dhedey (approved):

.github/actions/gradle-task/action.yml
les-sheppard: haha - I have no idea why I didn't do that originally. Probably being a bit slow towards the end of the day ;)

.github/actions/gradle-task/action.yml
les-sheppard: I've had another look at `get_version`. It doesn't look as though the output is used at all in one place: https://github.com/radixdlt/babylon-node/blob/release/anemone-v1.1.3/.github/workflows/add-artifacts-to-release.yml#L136 

The `Build core-rust` step just above it, uses `{{ github.event.release.tag_name }}` to name the zip file.

Since this workflow is only called for releases, then I think it would be safe to remove both calls to `get_version` and replace the use of `steps.get_version.outputs.radixdlt_version` in that workflow with `{{ github.event.release.tag_name }}` 

What do you think? Or would you rather use VERSION_TAG?

dhedey: I think I'd slightly rather use `VERSION_TAG` just so everything goes through one place 👍

But happy for us to get rid of the `get_version` action completely if it helps.

les-sheppard: Latest changes:

- Call the `setup_version_properties` action in the `add-artifacts-to-release worflow`
- Use the VERSION_TAG output from `setup_version_properties` 
- Removed usages of `get-version`
- Removed `get-version` action

PR #920, opened 2024-04-23
Pin a different version of openjdk17 deps in docker for ARM.

jakrawcz-rdx:
## Summary

Addresses a build failure spotted at https://github.com/radixdlt/babylon-node/actions/runs/8797900181/job/24143785471#step:19:527.
We recently had to bump a version, but it is not available for ARM (see https://packages.debian.org/bookworm/openjdk-17-jre-headless).

## Testing

N/A, a build problem only.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -210,6 +210,7 @@ jobs:
       cache_tag_suffix: arm64
       build-args: |
         WGET_VERSION=1.21.3-1+b1
+        OPENJDK_17_VERSION=17.0.10+7-1~deb12u1
         VERSION_BRANCH=${{ needs.setup_version_properties.outputs.VERSION_BRANCH }}
         VERSION_COMMIT=${{ needs.setup_version_properties.outputs.VERSION_COMMIT }}
         VERSION_DISPLAY=${{ needs.setup_version_properties.outputs.VERSION_DISPLAY }}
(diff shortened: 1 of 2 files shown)


.github/workflows/docker.yml
```diff
@@ -210,6 +210,7 @@ jobs:
       cache_tag_suffix: arm64
       build-args: |
         WGET_VERSION=1.21.3-1+b1
+        OPENJDK_17_VERSION=17.0.10+7-1~deb12u1
```
dhedey, line 213: Can we add comments explaining why we're using a different version?

Dockerfile
```diff
@@ -21,6 +21,7 @@ ENV DEBIAN_FRONTEND noninteractive
 CMD ["/bin/bash"]
 
 ARG WGET_VERSION="1.21.3-1+b2"
+ARG OPENJDK_17_VERSION="17.0.11+9-1~deb12u1"
```
dhedey, line 24: My concern is:
* This DockerFile is probably broken building locally on ARM without knowing argument magic
* We don't have comments / it's not clear that we might need to update both the ARM version when this is updated

Perhaps we should instead use `TARGETPLATFORM` and then encode the two versions inside the docker file, for better locality, and to ensure they work without magic build parameters? e.g. https://nielscautaerts.xyz/making-dockerfiles-architecture-independent.html

For this release, maybe we can just pull in `17.0`, and build something better on a separate PR?

dhedey (approved):
Approving to get this release out, we can revisit later this week on a separate branch.

Dockerfile
jakrawcz-rdx: let's do it in https://radixdlt.atlassian.net/browse/NODE-635

PR #922, opened 2024-04-23
Merge `release/anemone-v1.1.3` into `develop`

dhedey:
> [!IMPORTANT]
>
> * Please read our [Contributing Guidelines](https://github.com/radixdlt/babylon-node/blob/main/CONTRIBUTING.md) before opening a PR.
> * Before creating your PR, please ensure you have used the _correct base branch_ as per the [branching strategy](https://github.com/radixdlt/babylon-node/blob/main/docs/branching-strategy.md), both for branching from, and in the PR UI above.
>   * For most code changes, this is `develop`.
>   * For stand-alone docs changes, this is `main`.
>   * For workflow changes, this is the oldest supported `release/*` branch. 
> * Please remove these sections as you fill out your PR.
> 

## Summary

> [!TIP]
> 
> Start with the context of your PR. Why are you making this change? What does it address? Link back to an issue if relevant.
> 
> Then summarise the changes that were made. Bullet points are fine.

## Details

> [!TIP]
> 
> This section is optional. Go into more detail about the changes that were made, or the thinking behind the changes.

## Testing

> [!TIP]
> 
> Explain what testing / verification is done, including manual testing or automated testing.


core-rust/state-manager/src/store/historical_state.rs
```diff
@@ -77,48 +79,74 @@ use crate::{
 /// This database is backed by:
 /// - a [`ReadableTreeStore`] - a versioned source of ReNodes / Partitions / Substates metadata,
 /// - and a [`LeafSubstateValueStore`] - a store of Substate values' associated with their leafs.
-pub struct StateTreeBasedSubstateDatabase<'t, T> {
-    tree_store: &'t T,
+pub struct StateTreeBasedSubstateDatabase<'s, DS> {
+    base_store: DS,
     at_state_version: StateVersion,
+
+    // Note: "Why do we even need to capture `'t`?"
```
dhedey, line 86: Well, this is because of the fact that e.g. the `SubstateTier`'s `into_iter_substate_summaries_from` method exposes an iterator with a lifetime of the underlying _store_ rather than the entity tier (in other words, it can outlive the tier).

However, the underlying store comes from `self.base_store.deref()` and we need someone to capture the constraint that the reference created lives longer than the `&self` parameter in order to satisfy the `SubstateDatabase` trait definition for `list_entries_from`.

But `self.base_store.deref()` could either resolve to some random lifetime if `base_store` is a reference, or could resolve to `&self.base_store`, so could be bounded by `&self`.

We need some way of capturing that "the lifetime of `self.base_store.deref()`" outlives _ANY_ reference to self. This is actually extra information -- and this phantom lifetime is the best way of doing that.

dhedey: Perhaps it'd be clearer to actually make it a `PhantomData<&'s S>` and have  `'s, S: 's, DS: Deref<Target = S>`.

Or just `PhantomData<&'s ()>`. I think `PhantomData<&'s DS>` is a little misleading - because it's not pretending to encompass a reference to `DS`, but actually to some underlying store.

core-rust/state-manager/src/store/historical_state.rs
```diff
@@ -299,16 +335,139 @@ impl<'s, R: ReadableRocks> SubstateDatabase
...
+    fn batch_get_ancestry<'a>(
+        &self,
+        node_ids: impl IntoIterator<Item = &'a NodeId>,
+    ) -> Vec<Option<SubstateNodeAncestryRecord>> {
+        // Unfortunately, there is no easy way to filter out the "future" here
+        self.underlying().batch_get_ancestry(node_ids)
+    }
+}
+
+impl<'s, R: ReadableRocks + 's, DS: Deref<Target = StateManagerDatabase<R>>> ReNodeListingIndex
```
dhedey, line 363: `EntityListingIndex`*

core-rust/state-manager/src/store/traits.rs
```diff
@@ -647,6 +654,115 @@ pub mod extensions {
     }
 }
 
+pub mod indices {
+    use super::*;
+    use std::ops::Range;
+
+    pub trait ReNodeListingIndex {
```
dhedey, line 661: EntityListingIndex*

core-rust/state-manager/src/store/traits.rs
```diff
@@ -647,6 +654,115 @@ pub mod extensions {
...
+        }
+    }
+
+    define_single_versioned! {
+        #[derive(Debug, Clone, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
+        pub enum VersionedObjectBlueprintName => ObjectBlueprintName = ObjectBlueprintNameV1
+    }
+
+    /// An Object's ID and its blueprint name.
+    /// This is a "technical" structure stored in one of the ReNode-listing indices.
```
dhedey, line 754: Entity*

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -764,6 +800,10 @@ impl ActualStateManagerDatabase {
         state_manager_database.restore_december_2023_lost_substates(network);
         state_manager_database.ensure_historical_substate_values();
 
+        if state_manager_database.config.enable_entity_listing_indices {
```
dhedey, line 803: It would be more consistent to use an `ensure_` style and have this config check internal to the method? To either set up or tear down the index.

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -1996,6 +2059,104 @@ impl<R: WriteableRocks> StateManagerDatabase<R> {
...
+            .next()
+            .expect("next version");
+
+        let mut receipts_iter = db_context
+            .cf(TransactionReceiptsCf)
+            .iterate_from(&catchup_from_version, Direction::Forward)
+            .peekable();
+
+        while let Some((state_version, receipt)) = receipts_iter.next() {
+            self.batch_update_entity_listing_indices(
```
dhedey, line 2141: It's better to use the substate store than the receipt store for this IMO.

It's both faster, and doesn't require full receipt history to do.

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -1996,6 +2059,104 @@ impl<R: WriteableRocks> StateManagerDatabase<R> {
...
+                &db_context,
+                state_version,
+                &receipt.state_changes.substate_level_changes,
+            );
+            if state_version.number() % TXN_FLUSH_INTERVAL == 0 || receipts_iter.peek().is_none() {
+                info!(
+                    "ReNode listing indices updated to {}; flushing...",
```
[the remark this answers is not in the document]
dhedey: That's also a _lot!_ of info logs catching up on mainnet which has 75 million state versions right now.

Perhaps better to have info logs every 1 million transactions rather than every 10000? And then if you want logs every flush, perhaps make them `debug!` logs?

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -1996,6 +2059,104 @@ impl<R: WriteableRocks> StateManagerDatabase<R> {
...
+                    state_version
+                );
+                db_context.cf(ExtensionsDataCf).put(
+                    &ExtensionsDataKey::EntityListingIndicesLastProcessedStateVersion,
+                    &state_version.to_be_bytes().to_vec(),
+                );
+                db_context.flush();
+            }
+        }
+        info!("ReNode listing indices are caught up.");
```
dhedey, line 2158: Entity*

core-rust/state-manager/src/store/rocks_db.rs
dhedey: Not worth doing now, but in future, maybe worth raising as a separate ticket to change how we rebuild this index?

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -2154,6 +2315,57 @@ impl<R: ReadableRocks> IterableAccountChangeIndex for StateManagerDatabase<R> {
     }
 }
 
+impl<R: ReadableRocks> ReNodeListingIndex for StateManagerDatabase<R> {
```
dhedey, line 2318: EntityListingIndex*

core-rust/state-manager/src/store/rocks_db.rs
dhedey: Ohh actually I see that we _have_ to use the transaction receipts, due to the way the index includes the state version and index in transaction bits! Okay, ignore this :)

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -1996,6 +2059,104 @@ impl<R: WriteableRocks> StateManagerDatabase<R> {
...
+        let db_context = self.open_rw_context();
+        let catchup_from_version = db_context
+            .cf(ExtensionsDataCf)
+            .get(&ExtensionsDataKey::EntityListingIndicesLastProcessedStateVersion)
+            .map(StateVersion::from_be_bytes)
+            .unwrap_or(StateVersion::pre_genesis())
+            .next()
+            .expect("next version");
+
+        let mut receipts_iter = db_context
```
dhedey, line 2135: Hmm I thought `catchup_from_version` is actually `previous_indexed_up_to_version`?
So shouldn't we start iterating from `previous_indexed_up_to_version + 1` inclusive?

dhedey: Although I imagine setting these is idempotent, so we won't have duplicated entries? So it might be safe?

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -1996,6 +2059,104 @@ impl<R: WriteableRocks> StateManagerDatabase<R> {
...
+                node_id,
+                &TYPE_INFO_FIELD_PARTITION,
+                &TypeInfoField::TypeInfo.into(),
+            );
+            let Some(type_info_creation) = type_info_creation else {
+                continue;
+            };
+            let SubstateChangeAction::Create { new } = type_info_creation else {
+                panic!(
+                    "type info substate should be immutable: {:?}",
```
dhedey, line 2080: This seems quite scary. e.g. this won't be true if the blueprint version is updated eventually.

But the entity id and blueprint name should be fixed by the system, so I think it's fine to just ignore updates.

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -1996,6 +2059,104 @@ impl<R: WriteableRocks> StateManagerDatabase<R> {
...
+                    type_info_creation
+                );
+            };
+            let type_info = scrypto_decode::<TypeInfoSubstate>(new).expect("decode type info");
+
+            let entity_type = node_id.entity_type().expect("type of upserted ReNode");
```
dhedey, line 2086: entity*

core-rust/state-manager/src/store/rocks_db.rs
dhedey: We're also missing the teardown logic for this index.

dhedey (commented):
Various comments on the index building logic. I don't think any of this is a release blocker, although we'd ideally fix the logging.

Perhaps we can raise a ticket to tidy up the rest?

core-rust/state-manager/src/store/rocks_db.rs
jakrawcz-rdx: I went through the same journey when implementing that :p

core-rust/state-manager/src/store/rocks_db.rs
jakrawcz-rdx: a tiny `.next()` (in line 2132 above) is the `+ 1` you are looking for :D

FOR ONCE I avoided the "off by one" 🎉 (I think 🤔)

dhedey: Ahh I thought that was a `.next()` to return the first value from an iterator! 
In retrospect, perhaps `next` is a little too overloaded, but it's minor anyway?

In terms of this code, tiny suggestion - maybe separating out `let catchup_from_version  = previously_updated_to_version.next();` might make it clearer / more explicit for noobs like me?

jakrawcz-rdx: this I can do: https://github.com/radixdlt/babylon-node/pull/926/commits/e8b08ccc5ff4acb8877f003d295c97070598743f 👍

jakrawcz-rdx (approved):

PR #923, opened 2024-04-24
Feature/bump okhttp version

siy:
The existing version of okhttp (4.11.0) brings transitive dependencies which have known vulnerabilities. Although they don't affect our code directly, it still makes sense to bump the version. No other dependencies require updates.


jakrawcz-rdx (approved):
Very nice and isolated version bump 👍

PR #924, opened 2024-04-24
Bump Scrypto dep to `bottlenose-d02dac77`

jakrawcz-rdx:
## Summary

This just keeps the scrypto dep fresh for https://radixdlt.atlassian.net/browse/NODE-619.

## Details

Apart from some obvious Node adjustments, this PR contains:
- updated apidoc version to `1.2.0 (Bottlenose)` (this is the reason for massive line count after regenerating types)
- some new fields returned within `BootLoaderModuleFieldSystemBootValue`
- a slight refactor of `CostingParameters` handling (see `executable_logic.rs`)
- one instance of `TODO(resolve during review)`, related to "no fees" config

## Testing

Some test asserts were updated, since a rejected transaction is now signalled by a differently-named error (see `SuccessButFeeLoanNotRepaid`).


core-rust/core-api-server/src/core_api/conversions/substates/type_info_module.rs
```diff
@@ -10,8 +10,8 @@ pub fn to_api_system_boot_substate(
     substate: &SystemBoot,
 ) -> Result<models::Substate, MappingError> {
     let value = match substate {
-        SystemBoot::V1 { costing_parameters } => models::BootLoaderModuleFieldSystemBootValue::new(
-            to_api_system_costing_parameters(context, costing_parameters)?,
+        SystemBoot::V1(system_parameters) => models::BootLoaderModuleFieldSystemBootValue::new(
```
dhedey, line 13: The system boot and vm boot aren't really `type_info_module` substates, perhaps they should be put in `boot_loader_module.rs`?

core-rust/state-manager/src/protocol/protocol_updates/protocol_update_definition.rs
```diff
@@ -89,7 +89,7 @@ impl ProtocolStateComputerConfig {
         ProtocolStateComputerConfig {
             network,
             validation_config: ValidationConfig::default(network_id),
-            costing_parameters: CostingParameters::default(),
+            costing_parameters: CostingParameters::babylon_genesis(),
```
dhedey, line 92: For consistency, I suggest we call this method `babylon_genesis` :)

core-rust/state-manager/src/protocol/protocol_updates/protocol_update_definition.rs
```diff
@@ -114,6 +114,7 @@ impl ProtocolStateComputerConfig {
     ) -> ExecutionConfigurator {
         let mut costing_parameters = self.costing_parameters;
         if no_fees {
+            // TODO(resolve during review): replace with `SystemOverrides.disable_costing`?
```
dhedey, line 117: Yes, definitely :). There's actually a lot of stuff we can remove / simplify with these system overrides changes.

core-rust/state-manager/src/transaction/executable_logic.rs
```diff
@@ -64,43 +63,65 @@ impl ExecutionConfigurator {
     ) -> Self {
         Self {
             scrypto_vm: ScryptoVm::<DefaultWasmEngine>::default(),
-            costing_parameters,
             execution_configs: HashMap::from([
                 (
                     ConfigType::Genesis,
                     ExecutionConfig::for_genesis_transaction(network.clone())
+                        .with_costing_parameters(costing_parameters)
```
dhedey, line 70: Do we ever need to override costing parameters? Can we just get rid of this? (And instead use the new `disable_costing` if we need to?)

core-rust/state-manager/src/transaction/executable_logic.rs
```diff
@@ -64,43 +63,65 @@ impl ExecutionConfigurator {
...
                 (
                     ConfigType::OtherSystem,
-                    ExecutionConfig {
-                        max_number_of_events: 1_000_000,
-                        ..ExecutionConfig::for_system_transaction(network.clone())
-                            .with_kernel_trace(engine_trace)
-                    },
+                    ExecutionConfig::for_system_transaction(network.clone())
+                        .with_costing_parameters(costing_parameters)
+                        .with_limit_parameters(LimitParameters {
```
dhedey, line 77: `for_system_transaction` turns off the limit module and costing module now, so I think we can remove this override.

core-rust/state-manager/src/transaction/executable_logic.rs
```diff
@@ -259,3 +275,47 @@ fn collect_new_node_ids(state_updates: &StateUpdates) -> IndexSet<NodeId> {
...
+            execution_trace,
+            system_overrides,
+        } = self;
+        ExecutionConfig {
+            enable_kernel_trace,
+            enable_cost_breakdown,
+            execution_trace,
+            system_overrides: Some(SystemOverrides {
+                costing_parameters: Some(costing_parameters),
+                ..system_overrides.expect("all ExecutionConfig's constructors set this field")
```
dhedey, line 299: Perhaps we should just use `unwrap_or_default()` here? Although arguably upstream we could just get rid of `Some(SystemOverrides)` and force SystemOverrides to be provided

core-rust/state-manager/src/transaction/round_update_transaction.rs
```diff
@@ -152,6 +152,11 @@ impl PreparedRoundUpdateTransactionV1 {
                 },
                 pre_allocated_addresses: vec![],
             },
+            // Note: technically the round update is a "system transaction" (which would justify
```
dhedey, line 155: Hmm it should be true I think... any existing limits should never be hit and can effectively be 
 reasoned as being infinite.
Round changes can be quite large if they're an epoch change... So I think we should set this as True and potentially get rid of the `ConfigType::OtherSystem` type to simplify things.

Paging @talekhinezh for advice/input too.

dhedey (approved):
Looks good.

A couple of suggestions for changes to simplify our configuration in the node (which were added in to work around restrictions on the engine side) and align better with the engine now with its better abstractions.

core-rust/state-manager/src/protocol/protocol_updates/protocol_update_definition.rs
jakrawcz-rdx: N/A anymore
(the `costing_parameters` are gone from here, and the remaining impl is all about `default`s)

core-rust/state-manager/src/transaction/round_update_transaction.rs
jakrawcz-rdx: After the other changes (i.e. from other review comments) here, my `// Note:` is no longer true (i.e. we do not customize `LimitParameters` now).
I also consider the "round update" a system transaction (and I was surprised we wanted to e.g. track limits for it before - but maybe someone had a reason?).

> get rid of the `ConfigType::OtherSystem` type to simplify things

do you suggest to treat it as `::Regular`? (plz note the same would have to happen for `FlashV1`)
Logic-wise, it seems that would be totally equivalent; but maybe conceptually we want to have this special label anyway? (I don't have very strong opinion)

jakrawcz-rdx: BTW: `ConfigType::Pending` is equivalent to `ConfigType::Regular` too, at this point. I can accept getting rid of it.

jakrawcz-rdx: Conclusion:
- let's get rid of `ConfigType::OtherSystem`, because `system: true` does all this anyways.
- Engine can (and probably should?) get rid of `ExecutionConfig::for_system_transaction` & co - FYI @talekhinezh 
- the `ConfigType::Pending` stays, because we use it to differentiate some log warn threshold.

jakrawcz-rdx:
All the comments got addressed (and some discussed on Slack), lemme merge.

PR #925, opened 2024-04-24
Feature/node 622 update netty

siy:
## Summary

This PR provides update to netty dependency to the latest version in order to avoid known vulnerabilities. It also cleanups use of different versions of netty in different modules.


siy:
> I appreciate the dedup of netty versions 👍
> 
> AFAIS in the original report (https://security.snyk.io/vuln/SNYK-JAVA-IONETTY-1584064), the suggestion is to bump to >=4.1.68.Final. Maybe we should do just this instead of 109? (in the name of "minimal required change to deps")

I'm using Netty in one of my pet projects and ought to update it frequently because virtually every version until .108 has one or more vulnerabilities.

jakrawcz-rdx (approved):
ok, I guess we can trust your reasoning (and Netty's reputation) 👍

PR #930, opened 2024-04-30
Remove `ValidationConfig` from protocol configuration

jakrawcz-rdx:
## Summary

A pre-refactor for https://radixdlt.atlassian.net/browse/NODE-618.

## Details

In the spirit of YAGNI, and after some discussion, this PR removes a considerable complexity of "Support for changing `ValidationConfig` as part of protocol update". It should help to achieve the core "run scenarios as part of protocol update" easier.

## Testing

The support mentioned above was not used at all, so we just need the regression tests to pass.


jakrawcz-rdx:
Note: it looks like after refactors of https://github.com/radixdlt/babylon-node/pull/936, re-introducing a feature like this one (whatever actually: updating the validation config or any other config, or re-initializing any Rust service) as part of proto-update will be even easier.

LukasGasior1 (approved):

PR #939, opened 2024-05-06
Feature/node 427 event naming suffixes

siy:
## Summary

This PR introduces separation of events into local and remote categories. 

## Details

Introduced three marker interfaces - CoreEvent (base) and two extensions - LocalEvent and RemoteEvent. All event classes implement either LocalEvent or RemoteEvent. Some (rare) events implement both interfaces (if they can be generated locally or could be received from other nodes).


core/src/main/java/com/radixdlt/consensus/event/CoreEvent.java
```diff
@@ -0,0 +1,67 @@
...
+ * You are solely responsible for; testing, validating and evaluation of all operation
+ * logic, functionality, security and appropriateness of using the Work for any commercial
+ * or non-commercial purpose and for any reproduction or redistribution by You of the
+ * Work. You assume all risks associated with Your use of the Work and the exercise of
+ * permissions under this License.
+ */
+
+package com.radixdlt.consensus.event;
+
+public interface CoreEvent {}
```
jakrawcz-rdx, line 67: a marker interface is a perfect place to put a javadoc explaining its purpose

core/src/main/java/com/radixdlt/consensus/event/RemoteEvent.java
```diff
@@ -62,41 +62,7 @@
...
-  public T getEvent() {
-    return event;
-  }
-
-  @Override
-  public String toString() {
-    return String.format("%s[%s->%s]", getClass().getSimpleName(), this.origin, this.event);
-  }
-}
+// TODO: SIY: rename to RemoteEvent
```
jakrawcz-rdx, line 67: _(looks addressed)_

core/src/main/java/com/radixdlt/environment/rx/IncomingEvent.java
```diff
@@ -0,0 +1,103 @@
...
+
+import com.radixdlt.consensus.event.RemoteEvent;
+import java.util.Objects;
+
+/**
+ * A helper class which contains remote event and the origin node.
+ *
+ * @param <T> the event class
+ */
+public final class IncomingEvent<N, T extends RemoteEvent> implements RemoteEvent {
```
jakrawcz-rdx, line 75: looks like a good candidate for `record`?

core/src/main/java/com/radixdlt/environment/rx/ObservableProvider.java
```diff
@@ -81,7 +82,7 @@ public final class ObservableProvider<T> implements Provider<Observable<T>> {
     this.t = null;
   }
 
-  ObservableProvider(TypeLiteral<T> t) {
+  ObservableProvider(TypeLiteral<T> t) { // TODO: SIY: unused
```
jakrawcz-rdx, line 85: if so, it's an excellent opportunity to remove it

jakrawcz-rdx (commented):
very nice drive-by cleanups 👍 (e.g. `equals()`)

core/src/main/java/com/radixdlt/consensus/bft/RoundUpdate.java
```diff
@@ -65,75 +65,30 @@
...
-    if (o == null || getClass() != o.getClass()) {
-      return false;
-    }
-    final RoundUpdate that = (RoundUpdate) o;
-    return Objects.equals(currentRound, that.currentRound)
-        && Objects.equals(highQC, that.highQC)
-        && Objects.equals(leader, that.leader)
-        && Objects.equals(nextLeader, that.nextLeader);
-  }
+    record roundUpdate(
```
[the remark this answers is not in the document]
siy: Records have a public constructor, but here we need validation of the input parameters. So, there are two choices: add validation into the constructor, or use static factory method and plain constructor. Throwing exceptions in the constructor, usually recognized as a bad practice, so the only viable option is the static factory method. Switching to interface + private record naturally addresses these issues and enforces "code against interface" policy.

core/src/main/java/com/radixdlt/consensus/event/CoreEvent.java
siy: Good point. Done.

core/src/main/java/com/radixdlt/consensus/event/RemoteEvent.java
siy: Yep. Removed.

core/src/main/java/com/radixdlt/environment/rx/IncomingEvent.java
siy: Reworked it into interface and private record.

core/src/main/java/com/radixdlt/environment/rx/ObservableProvider.java
siy: Good point. Removed.

core/src/main/java/com/radixdlt/consensus/bft/RoundUpdate.java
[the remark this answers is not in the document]
jakrawcz-rdx: > Throwing exceptions in the constructor, usually recognized as a bad practice

I tried to read more about this being bad practice and I've only found vague suggestions that it results in a half-initialized object, and TBH I still don't see how bad it really is.
(I do agree that leaking a reference to `this` before constructor returns is bad, but I don't see relationship to throwing exceptions... Or am I looking in wrong direction, and constructor-exceptions are a bad practice for other reasons?)

It would be a pity to avoid constructor-exceptions, since they are so convenient with `record`s + their compact constructors :(
TBH from conceptual PoV, preventing successful initialization of a plainly-invalid object seems like a good idea, and the constructor is the most surefire place to achieve it. It also produces least boilerplate I think?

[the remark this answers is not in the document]
siy: Well, while in general throwing exceptions in constructor is acceptable, I see this as a violation of the single responsibility principle: constructor validates parameters AND initializes the object. In this regard, the factory method is a better choice as it performs validation and passes valid parameters to constructors.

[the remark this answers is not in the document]
siy: Another consideration is my general dislike to `new` in the code. Factory methods look cleaner to me, but this is a personal preference.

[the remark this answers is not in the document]
jakrawcz-rdx: I also favor factory methods, for their flexibility.

I love that e.g. `Set.of(..)` can internally do:
```
if 0 args, return new EmptySet(); // impl optimized for being empty
if 1 arg, return new SingletonSet(arg); // impl optimized for a single elem
if <10 args, return new TreeSet(args); // whatever other optimizations
if >10 args, return new HashSet(args); // ...
```
_(please excuse the silly example with Set, where this approach is probably overkill)_

And the above makes lot of sense when the constructed object is polymorphic, or maybe at least can be customized in non-trivial ways.

However, when everyone expects that the constructed event is just a POJO, and a very rigid one, and there is dozens different event types, then I think it's better to optimize for "simpler syntax, less boilerplate, direct usage of language's constructs" than for "factory method flexibility".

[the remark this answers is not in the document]
siy: I think that factory methods are a better approach from the point of view of consistency (some classes are not records, and it might be hard/inconvenient to make them records) and separation of concerns: the constructor should assemble the object, but validation is a separate task and it better be done in the factory method.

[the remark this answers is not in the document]
siy: It's more about consistency than about flexibility. There are some classes which have more than one static factory method. Also, in some cases, factory methods are used to hide some implementation details and keep defaults in one place. For example, ScheduledLocalTimeout. Now, only convention prevents direct instantiation via constructor, but I'd prefer this expressed directly in the code rather than expect that nobody will do this.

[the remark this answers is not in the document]
LukasGasior1: I'd personally just go for a top level `record RoundUpdate(...) implements LocalEvent` and ditch that non-null validation. That inner record seems a bit weird/unnecessary to me.
If we really want the validation then I'd go for an additional compact constructor, e.g.:
```
    public RoundUpdate {
        Objects.requireNonNull(...);
        ...
    }
```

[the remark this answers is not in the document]
siy: OK, looks like common consensus is to make all such classes records (where possible, there are some exceptions). Please let me know if there are any objections.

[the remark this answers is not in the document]
jakrawcz-rdx: sounds cool, I expect it will actually remove a lot of boilerplate 👍 

Regarding the "`requireNonNull(...)` inside compact constructors" - I typically value the fail-fast behavior, so I'd recommend to keep it (for the price of "whoops actually some boilerplate again!").
BUT if veterans @LukasGasior1 and @siy confirm that this codebase never had problems with cryptic NPEs from getters, then we can drop this clearly-overkill non-null validation too.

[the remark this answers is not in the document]
siy: At the best of my knowledge, there were never any NPE's or validation exceptions.

core/src/main/java/com/radixdlt/environment/ScheduledEventDispatcher.java
```diff
@@ -64,11 +64,13 @@
...
 
+import com.radixdlt.consensus.event.CoreEvent;
+
 /**
  * Dispatches an event after some given time.
  *
  * @param <T> the event class
  */
-public interface ScheduledEventDispatcher<T> {
+public interface ScheduledEventDispatcher<T extends CoreEvent> {
```
LukasGasior1, line 74: LocalEvent

core/src/main/java/com/radixdlt/environment/rx/IncomingEvent.java
```diff
@@ -0,0 +1,86 @@
...
+
+import com.radixdlt.consensus.event.RemoteEvent;
+import java.util.Objects;
+
+/**
+ * A helper class which contains remote event and the origin node.
+ *
+ * @param <T> the event class
+ */
+public interface IncomingEvent<N, T extends RemoteEvent> extends RemoteEvent {
```
LukasGasior1, line 75: does it need to implement `RemoteEvent` itself? :thinking:

core/src/main/java/com/radixdlt/environment/rx/ObservableProvider.java
```diff
@@ -66,33 +66,21 @@
...
 import com.google.inject.Inject;
 import com.google.inject.Provider;
-import com.google.inject.TypeLiteral;
+import com.radixdlt.consensus.event.CoreEvent;
 import io.reactivex.rxjava3.core.Observable;
 import java.util.Objects;
 
 /** Helper class to hook up observables to the environment */
-public final class ObservableProvider<T> implements Provider<Observable<T>> {
+public final class ObservableProvider<T extends CoreEvent> implements Provider<Observable<T>> {
```
LukasGasior1, line 74: LocalEvent

core/src/main/java/com/radixdlt/environment/rx/RxEnvironment.java
```diff
@@ -103,60 +106,61 @@ public RxEnvironment(
             .collect(Collectors.toMap(RxRemoteDispatcher::eventClass, d -> d));
   }
 
-  private <T> Optional<Subject<T>> getSubject(TypeLiteral<T> t) {
+  private <T extends CoreEvent> Optional<Subject<T>> getSubject(TypeLiteral<T> t) {
```
LukasGasior1, line 109: LocalEvent

core/src/main/java/com/radixdlt/environment/rx/RxEnvironment.java
```diff
@@ -103,60 +106,61 @@ public RxEnvironment(
...
     @SuppressWarnings("unchecked")
     Subject<T> eventDispatcher = (Subject<T>) typeLiteralSubjects.get(t);
 
     return Optional.ofNullable(eventDispatcher);
   }
 
-  private <T> Optional<Subject<T>> getSubject(Class<T> eventClass) {
+  private <T extends CoreEvent> Optional<Subject<T>> getSubject(Class<T> eventClass) {
```
LukasGasior1, line 116: LocalEvent

core/src/main/java/com/radixdlt/environment/rx/RxEnvironment.java
```diff
@@ -103,60 +106,61 @@ public RxEnvironment(
...
     @SuppressWarnings("unchecked")
     Subject<T> eventDispatcher = (Subject<T>) subjects.get(eventClass);
 
     return Optional.ofNullable(eventDispatcher);
   }
 
   @Override
-  public <T> EventDispatcher<T> getDispatcher(Class<T> eventClass) {
+  public <T extends CoreEvent> EventDispatcher<T> getDispatcher(Class<T> eventClass) {
```
LukasGasior1, line 124: LocalEvent

core/src/main/java/com/radixdlt/environment/rx/RxEnvironment.java
```diff
@@ -103,60 +106,61 @@ public RxEnvironment(
...
 
-  public <T> Observable<T> getObservable(Class<T> eventClass) {
+  public <T extends CoreEvent> Observable<T> getObservable(Class<T> eventClass) {
     return getSubject(eventClass)
         .orElseThrow(
             () -> new IllegalStateException(eventClass + " not registered as observable."));
   }
 
-  public <T> Observable<T> getObservable(TypeLiteral<T> t) {
+  public <T extends CoreEvent> Observable<T> getObservable(TypeLiteral<T> t) {
```
LukasGasior1, line 163: LocalEvent

core/src/main/java/com/radixdlt/ledger/LedgerExtension.java
```diff
@@ -88,7 +89,8 @@
  *       "TransactionRun" and "CommittedTransactionWithProof".
  * </ul>
  */
-public record LedgerExtension(List<RawLedgerTransaction> transactions, LedgerProof proof) {
+public record LedgerExtension(List<RawLedgerTransaction> transactions, LedgerProof proof)
```
LukasGasior1, line 92: it's a bit weird that we're using this record as an event on its own :thinking: 
what do you think about adding a wrapper for it with a better name? e.g. `record VerifiedLedgerSyncResponse(ledgerExtension)`.
Possibly in a separate pr

core/src/main/java/com/radixdlt/ledger/LedgerUpdate.java
```diff
@@ -77,4 +79,5 @@ public record LedgerUpdate(
     Option<EpochChange> epochChange,
     ProtocolState resultantProtocolState,
     // Only used in tests
-    List<RawLedgerTransaction> committedNonProtocolUpdateTransactions) {}
+    List<RawLedgerTransaction> committedNonProtocolUpdateTransactions)
+    implements LocalEvent, RemoteEvent {}
```
LukasGasior1, line 83: this is a LocalEvent

core/src/main/java/com/radixdlt/p2p/discovery/PeerDiscovery.java
```diff
@@ -90,7 +91,7 @@
  * "discovered" from the config (bootstrapDiscovery) and more peers are requested from the peers
  * we're already connected to.
  */
-public final class PeerDiscovery {
+public final class PeerDiscovery implements RemoteEvent {
```
LukasGasior1, line 94: this isn't an event

core/src/test-core/java/com/radixdlt/environment/deterministic/MultiNodeDeterministicRunner.java
```diff
@@ -104,7 +105,7 @@ public Injector getNode(int index) {
     return nodes.get(index);
   }
 
-  public <T> void dispatchToAll(Key<EventDispatcher<T>> dispatcherKey, T t) {
+  public <T extends LocalEvent> void dispatchToAll(Key<EventDispatcher<T>> dispatcherKey, T t) {
```
LukasGasior1, line 108: looks like we can remove this method; it's no longer used

core/src/test-core/java/com/radixdlt/environment/deterministic/network/ControlledDispatcher.java
```diff
@@ -102,15 +105,16 @@ private static long addTimeNoOverflow(long a, long b) {
   }
 
   @Override
-  public <T> EventDispatcher<T> getDispatcher(Class<T> eventClass) {
+  public <T extends CoreEvent> EventDispatcher<T> getDispatcher(Class<T> eventClass) {
```
LukasGasior1, line 108: LocalEvent

core/src/test-core/java/com/radixdlt/harness/deterministic/NodeEvents.java
```diff
@@ -100,11 +101,12 @@ public NodeEvents(Map<Class<?>, Set<NodeEventProcessor<?>>> processors) {
     this.processors = Objects.requireNonNull(processors);
   }
 
-  public <T> EventProcessor<T> processor(NodeId node, Class<T> eventClass) {
+  public <T extends CoreEvent> EventProcessor<T> processor(NodeId node, Class<T> eventClass) {
     return t -> processors.get(eventClass).forEach(c -> c.process(node, t));
   }
 
-  public <T> EventProcessorOnDispatch<T> processorOnDispatch(NodeId node, Class<T> eventClass) {
+  public <T extends CoreEvent> EventProcessorOnDispatch<T> processorOnDispatch(
```
LukasGasior1, line 108: looks like this whole class is no longer used and can be removed together with (also unused) `SafetyCheckerModule`

core/src/test-core/java/com/radixdlt/targeted/mempool/MempoolFillerUpdate.java
```diff
@@ -64,83 +64,36 @@
...
     this.parallelTransactions = parallelTransactions;
     this.sendToSelf = sendToSelf;
-    this.completableFuture = completableFuture;
   }
 
   public static MempoolFillerUpdate enable(int parallelTransactions, boolean sendToSelf) {
-    return new MempoolFillerUpdate(parallelTransactions, sendToSelf, null);
-  }
-
-  public static MempoolFillerUpdate enable(
```
LukasGasior1, line 89: Mempool filler seems no longer used, can we remove it? together with `MempoolFillerStarter` and related messages, etc?

LukasGasior1 (commented):
Nice, it should be much easier to follow now with those marker interfaces :+1: Added some comments - a few classes are marked incorrectly

core/src/main/java/com/radixdlt/consensus/bft/RoundUpdate.java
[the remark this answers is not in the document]
LukasGasior1: hah yeah I think we've got a team of pretty well established never-nullers :smile:

core/src/test-core/java/com/radixdlt/targeted/mempool/MempoolFillerUpdate.java
siy: Done

core/src/test-core/java/com/radixdlt/environment/deterministic/network/ControlledDispatcher.java
siy: Unfortunately, this change clashes with this piece of code ([Dispatchers.java](https://github.com/radixdlt/babylon-node/blob/4e79c00a70cd7c4ed06b77c587c40d6ee34e5e77/core/src/main/java/com/radixdlt/environment/Dispatchers.java#L162)):
```
    @Override
    public RemoteEventDispatcher<NodeId, T> get() {
      var remoteDispatcher = environmentProvider.get().getRemoteDispatcher(messageType);
      var localDispatcher = environmentProvider.get().getDispatcher(messageType);
```
messageType has RemoteEvent bound, while .getDisplatcher() requires LocalEvent. I have two variants: 
- Make CoreEvent as a type bound for .getDispatcher() return value
- Set type bound to RemoteEvent & LocalEvent and make all RemoteEvent implementations to implement LocalEvent too.
The latter variant looks like it defies the purpose of the type separation, so I decided to lower the type bound to the common CoreEvent type.

core/src/main/java/com/radixdlt/p2p/discovery/PeerDiscovery.java
siy: Fixed

core/src/main/java/com/radixdlt/consensus/event/LocalEvent.java
```diff
@@ -0,0 +1,68 @@
...
+ * You are solely responsible for; testing, validating and evaluation of all operation
+ * logic, functionality, security and appropriateness of using the Work for any commercial
+ * or non-commercial purpose and for any reproduction or redistribution by You of the
+ * Work. You assume all risks associated with Your use of the Work and the exercise of
+ * permissions under this License.
+ */
+
+package com.radixdlt.consensus.event;
+
+/** Marker interface for events which are generated locally. */
```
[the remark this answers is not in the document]
siy: Done

core/src/main/java/com/radixdlt/consensus/event/RemoteEvent.java
```diff
@@ -62,41 +62,10 @@
...
-package com.radixdlt.environment.rx;
-
-import java.util.Objects;
+package com.radixdlt.consensus.event;
 
 /**
- * A helper class which contains remote event and the origin node.
- *
- * @param <T> the event class
+ * Marker interface for events which are received from other nodes. Note that some events may be
```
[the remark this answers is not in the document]
siy: Done

core/src/test-core/java/com/radixdlt/environment/deterministic/MultiNodeDeterministicRunner.java
siy: Done

core/src/main/java/com/radixdlt/consensus/bft/Round.java
```diff
@@ -64,8 +64,10 @@
 
 package com.radixdlt.consensus.bft;
 
+import com.radixdlt.consensus.event.LocalEvent;
+
 /** Represents a BFT round/round used by the Pacemaker of a BFT instance */
-public final class Round implements Comparable<Round> {
+public final class Round implements Comparable<Round>, LocalEvent {
```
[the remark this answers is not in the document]
siy: It is bound to the event processor in one of the test modules. I've removed that part and removed Round from events hierarchy.

core/src/main/java/com/radixdlt/consensus/bft/ProposalRejected.java
```diff
@@ -64,5 +64,8 @@
 
 package com.radixdlt.consensus.bft;
 
+import com.radixdlt.consensus.event.LocalEvent;
+import com.radixdlt.consensus.event.RemoteEvent;
+
 /** An event indicating that a proposal has been rejected */
-public record ProposalRejected(Round round) {}
+public record ProposalRejected(Round round) implements RemoteEvent, LocalEvent {}
```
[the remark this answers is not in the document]
siy: Fixed

core/src/main/java/com/radixdlt/consensus/ConsensusByzantineEvent.java
```diff
@@ -66,9 +66,11 @@
 
 import com.google.common.hash.HashCode;
 import com.radixdlt.consensus.bft.BFTValidatorId;
+import com.radixdlt.consensus.event.LocalEvent;
+import com.radixdlt.consensus.event.RemoteEvent;
 import com.radixdlt.p2p.NodeId;
 
-public sealed interface ConsensusByzantineEvent {
+public sealed interface ConsensusByzantineEvent extends RemoteEvent, LocalEvent {
```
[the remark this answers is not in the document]
siy: Fixed

jakrawcz-rdx (approved):
The boilerplate reduction is massive!

I will assume that Łukasz correctly pointed out what is Local and what is Remote, and if so - then the PR looks ready to merge to me 👍

core/src/test-core/java/com/radixdlt/environment/deterministic/network/ControlledDispatcher.java
LukasGasior1: Hmh, if we're dispatching a RemoteEvent to ourselves that means it must also be a LocalEvent. Wdyt about doing a runtime check and a cast or throw if this isn't a local event.

core/src/main/java/com/radixdlt/environment/Environment.java
```diff
@@ -65,18 +65,23 @@
...
+import com.radixdlt.consensus.event.RemoteEvent;
 import com.radixdlt.p2p.NodeId;
 
 /**
  * The environment where events get dispatched to. The implementing environment is then responsible
  * for getting dispatched events to their processors.
  */
 public interface Environment {
-  <T> EventDispatcher<T> getDispatcher(Class<T> eventClass);
+  <T extends CoreEvent> EventDispatcher<T> getDispatcher(Class<T> eventClass);
```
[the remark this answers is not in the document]
siy: There is a place where it is used for remote events as well. See [other comment](https://github.com/radixdlt/babylon-node/pull/939#discussion_r1595573766) for details.

core/src/main/java/com/radixdlt/environment/EventDispatcher.java
```diff
@@ -64,11 +64,13 @@
...
 
+import com.radixdlt.consensus.event.CoreEvent;
+
 /**
  * A dispatcher of an event to an environment
  *
  * @param <T> the event class
  */
-public interface EventDispatcher<T> {
+public interface EventDispatcher<T extends CoreEvent> {
```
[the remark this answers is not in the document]
siy: Unfortunately, this is tightly coupled with the getDispatcher() return type. See comment below.

core/src/main/java/com/radixdlt/environment/rx/IncomingEvent.java
siy: Unfortunately yes,(RxModuleRunnerImpl.java:146-150):
```
    public <N, T extends RemoteEvent> Builder add(
        Flowable<IncomingEvent<N, T>> o, RemoteEventProcessor<N, T> p) {
      subscriptionsBuilder.add(new Subscription<>(o.toObservable(), p::process));
      return this;
    }
```

core/src/main/java/com/radixdlt/environment/EventProducer.java
```diff
@@ -64,9 +64,10 @@
 
 package com.radixdlt.environment;
 
+import com.radixdlt.consensus.event.CoreEvent;
 import java.util.function.Supplier;
 
-public final class EventProducer<T> implements EventProcessor<T> {
+public final class EventProducer<T extends CoreEvent> implements EventProcessor<T> {
```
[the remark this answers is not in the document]
siy: Same issue

core/src/main/java/com/radixdlt/environment/NoOpEnvironment.java
```diff
@@ -65,22 +65,28 @@
...
 
 import com.google.inject.TypeLiteral;
+import com.radixdlt.consensus.event.CoreEvent;
+import com.radixdlt.consensus.event.LocalEvent;
+import com.radixdlt.consensus.event.RemoteEvent;
 import com.radixdlt.p2p.NodeId;
 
 public final class NoOpEnvironment implements Environment {
-  public <T> EventDispatcher<T> getDispatcher(Class<T> eventClass) {
+  public <T extends CoreEvent> EventDispatcher<T> getDispatcher(Class<T> eventClass) {
```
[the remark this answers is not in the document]
siy: Unfortunately, same issue.

core/src/main/java/com/radixdlt/environment/ScheduledEventDispatcher.java
siy: Same issue as well.

core/src/main/java/com/radixdlt/environment/rx/ObservableProvider.java
siy: Same issue.

core/src/main/java/com/radixdlt/environment/rx/RxEnvironment.java
siy: Same issue too.

core/src/main/java/com/radixdlt/environment/rx/RxEnvironment.java
siy: Same issue.

core/src/main/java/com/radixdlt/environment/rx/RxEnvironment.java
siy: Same issue.

core/src/main/java/com/radixdlt/environment/rx/RxEnvironment.java
```diff
@@ -103,60 +106,61 @@ public RxEnvironment(
...
 
-    @SuppressWarnings("unchecked")
-    final RemoteEventDispatcher<NodeId, T> dispatcher =
-        (RemoteEventDispatcher<NodeId, T>) remoteDispatchers.get(messageType).dispatcher();
-    return dispatcher;
+    return (RemoteEventDispatcher<NodeId, T>) remoteDispatchers.get(messageType).dispatcher();
   }
 
-  public <T> Observable<T> getObservable(Class<T> eventClass) {
+  public <T extends CoreEvent> Observable<T> getObservable(Class<T> eventClass) {
```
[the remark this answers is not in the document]
siy: Same issue.

core/src/main/java/com/radixdlt/environment/rx/RxEnvironment.java
siy: Same issue.

core/src/main/java/com/radixdlt/ledger/LedgerExtension.java
siy: I think it is worth to add to separate ticket.

core/src/main/java/com/radixdlt/ledger/LedgerUpdate.java
siy: Fixed

core/src/test-core/java/com/radixdlt/environment/deterministic/network/ControlledDispatcher.java
siy: This is a good idea, but I think that such rework is sensibly outside the scope of the ticket, which is not meant to change behavior. I think it makes sense to perform this change (as well as cleanup of remaining CoreEvent and fixing LedgerExtension issue) in the separate ticket/PR.

PR #943, opened 2024-05-10
Add protocol update scenario execution to testnets and localnet

LukasGasior1:


jakrawcz-rdx (approved):
Thanks for remembering about it 👍

PR #944, opened 2024-05-10
Update Bottlenose from earlier release branches

jakrawcz-rdx:
## Summary

This addresses a couple minor items of the checklist https://radixdlt.atlassian.net/browse/NODE-640.

## Details

- pull `main` and `anemone-1.1.3`
  - note: technically we could also pull `develop`, but it has only 1 newer refactor PR which merged today, so I prefer to consider it "past the Bottlenose cut-off point"
- Update [babylon-node/docs/branching-strategy.md at main · radixdlt/babylon-node](https://github.com/radixdlt/babylon-node/blob/main/docs/branching-strategy.md) on the latest supported branch.
- Update the version in testnet-node/docker-compose.yml to the eventually-to-be-published-version. 
- A few other items turned out to be no-op:
  - No need to bump version in `system-api-schema.yml` (no changes from previous release)
  - The version in `core-api-schema.yml` is already at 1.2.0
  - The version in `engine-state-api-schema.yml` is at 0.1-beta and can stay this way

## Testing

Only regression tests should pass.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -12,6 +12,8 @@ From the repository root, run the following - replacing 2 with the number of val
 ./docker/scripts/rundocker.sh 2
 ```
 
+This script requires [`../core-rust`](../core-rust) to be compiled.
+
 ## The Dockerfile
 
 ![The structure of the docker image](babylon-node-docker-build.png)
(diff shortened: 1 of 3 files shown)


LukasGasior1 (approved):
:+1:

PR #947, opened 2024-05-14
Feature/node 646 follow up for node 427

siy:
Follow-up changes for NODE-427 (event renaming). Cleaned up most uses of CoreEvent and removed some unused code.


core/src/main/java/com/radixdlt/environment/EventProcessorOnRunner.java
```diff
@@ -70,7 +70,8 @@
 import java.util.Optional;
 
 /**
- * An Event Processor registered to run on a runner.
+ * An Event Processor registered to run on a runner. Can be used to process both, local and remote
+ * events.
```
LukasGasior1, line 74: same here, why does it need to handle remote events too?

cli-tools/src/main/java/com/radixdlt/shell/RadixShell.java
```diff
@@ -301,16 +300,17 @@ public <T extends LocalEvent> void dispatch(T t) {
...
-      final var disposable =
-          injector
-              .getInstance(RxEnvironment.class)
-              .getObservable(eventClass)
-              .subscribe(consumer::accept);
-
-      eventConsumers.add(disposable);
-      return disposable;
-    }
+    //    public <T extends CoreEvent> Disposable onEvent(Class<T> eventClass, Consumer<T> consumer)
```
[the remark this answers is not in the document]
siy: Yep. Fixed.

core/src/main/java/com/radixdlt/consensus/sync/GetVerticesRequest.java
```diff
@@ -65,7 +65,8 @@
 package com.radixdlt.consensus.sync;
 
 import com.google.common.hash.HashCode;
+import com.radixdlt.consensus.event.LocalEvent;
 import com.radixdlt.consensus.event.RemoteEvent;
 
-/** Parameters for a local get vertices request */
-public record GetVerticesRequest(HashCode vertexId, int count) implements RemoteEvent {}
+/** Parameters for a local get vertices request. Also, could be scheduled locally. */
```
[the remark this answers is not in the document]
siy: It should not be, but some tests use it together with `ScheduledEventDispatcher` which requires `LocalEvent`. So, we have a tradeoff - add `LocalEvent` here or change the type bound for `ScheduledEventDispatcher` to `CoreEvent`. To me, the former option is less permissive, hence better. Mentioning that this is necessary only for tests should help clarify things, though.

core/src/main/java/com/radixdlt/environment/EventProcessorOnRunner.java
siy: The issue lies in the way `EventProcessorOnDispatch` is used in `RemoteDispatcherProvider` (see Dispatchers.java):
```
    @Inject private Set<EventProcessorOnDispatch<?>> onDispatchProcessors;
```
The original code makes no differentiation between local and remote events when `onDispatch` is concerned. This inevitably pulls `EventProcessor` into the same issue because of the `EventProcessorOnDispatch.getProcessor` method.

So, the comment is correct as for the current state of the things. Separation of `EventProcessorOnDispatch` for remote events requires reworking of the modules and bindings, which I'd prefer not to do now.

core/src/main/java/com/radixdlt/environment/EventProcessor.java
```diff
@@ -67,7 +67,8 @@
 import com.radixdlt.consensus.event.CoreEvent;
 
 /**
- * Processes an event. An environment is responsible for making the call into processors.
+ * Processes an event. An environment is responsible for making the call into processors. Note that
+ * this processor can be used for both types of events, local and remote.
```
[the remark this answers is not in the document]
siy: See comment below.

core/src/main/java/com/radixdlt/consensus/sync/GetVerticesRequest.java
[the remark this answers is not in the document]
LukasGasior1: I think it's just `EpochManagerTest` that's (wrongly) binding that to a scheduled dispatcher. We can just remove it from there

core/src/main/java/com/radixdlt/environment/EventProcessor.java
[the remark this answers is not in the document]
LukasGasior1: It seems that it's just `VertexRequestRateInvariant` (used in two tests) that captures a remote event (`GetVerticesRequest`) using `nodeEvents` and that requires those bindings in `DispatcherModule` / `SimulationNodeEventsModule`. What do you think about adding a separate interface for this purpose, e.g. `interface RemoteEventCapture<T: RemoteEvent> { void capture(t T);  }` and then in `DispatcherModule`:

```
  @Provides
  private RemoteEventDispatcher<NodeId, GetVerticesRequest> verticesRequestDispatcher(
      Set<RemoteEventCapture<GetVerticesRequest>> capture,
      Environment environment,
      Metrics metrics) {
    var dispatcher = environment.getRemoteDispatcher(GetVerticesRequest.class);
    return (node, request) -> {
      metrics.bft().sync().requestsSent().inc();
      dispatcher.dispatch(node, request);
      capture.forEach(e -> e.capture(request));
    };
  }
```

and in `SimulationNodeEventsModule`:

```
  @ProvidesIntoSet
  private RemoteEventCapture<GetVerticesRequest> getVerticesRequestCapture(
      @Self NodeId node, NodeEvents nodeEvents) {
    return nodeEvents.processor(node, GetVerticesRequest.class);
  }
```

?

That would allow use to have a cleaner overall design where EventDispatcher/EventProcessor are only for local events and RemoteEventDispatcher / RemoteEventProcessor are remote only.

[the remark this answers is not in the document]
LukasGasior1: Check out my commit on this branch: `feature/NODE-646-follow-up-for-NODE-427-local-events`; something like this should work

core/src/main/java/com/radixdlt/environment/EventProcessorOnRunner.java
LukasGasior1: I don't think much rework is needed, check my commit on `feature/NODE-646-follow-up-for-NODE-427-local-events`

siy: Incorporated your changes with minor tweaks.

core/src/main/java/com/radixdlt/consensus/sync/GetVerticesRequest.java
```diff
@@ -67,5 +67,5 @@
 import com.google.common.hash.HashCode;
 import com.radixdlt.consensus.event.RemoteEvent;
 
-/** Parameters for a local get vertices request */
+/** Parameters for a local get vertices request. Also, could be scheduled locally in tests. */
```
LukasGasior1, line 70: I don't think this comment is correct - we don't schedule it anywhere in tests, unless I'm missing something?
The first part of this comment seems wrong too: "Parameters for a local get vertices request".
I'd change it to: "A request to fetch up to _count_ vertices starting at _vertexId_ from a remote peer.".

core/src/main/java/com/radixdlt/environment/Dispatchers.java
```diff
@@ -161,19 +158,35 @@ private static final class RemoteDispatcherProvider<T extends RemoteEvent>
...
-          remoteDispatcher.dispatch(node, e);
-        }
-        onDispatch.forEach(p -> p.process(e));
-      };
+
+      if (LocalEvent.class.isAssignableFrom(messageType)) {
+        @SuppressWarnings({"unchecked", "rawtypes"})
+        var localMessageType = (Class<LocalEvent>) (Class) messageType;
+        var localDispatcher = environmentProvider.get().getDispatcher(localMessageType);
+        var onDispatch =
```
LukasGasior1, line 166: hmh, it might be better to move this inside `if (node.equals(self)) {` block. We shouldn't  have any "on dispatch" processors for remote events

core/src/main/java/com/radixdlt/environment/EventProcessor.java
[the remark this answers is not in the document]
LukasGasior1: this comment still needs an update after recent changes :)

core/src/main/java/com/radixdlt/environment/EventProcessorOnDispatch.java
```diff
@@ -64,33 +64,25 @@
...
 
-import com.radixdlt.consensus.event.CoreEvent;
-import java.util.Objects;
+import com.radixdlt.consensus.event.LocalEvent;
 import java.util.Optional;
 
 /**
- * A processor which is to process an event synchronously at time of dispatch
+ * A processor which is to process an event synchronously at time of dispatch. Note that this
+ * processor can be used for both types of events, local and remote.
```
LukasGasior1, line 72: this comment needs an update; this processor is for local events only

LukasGasior1 (commented):
a few minor comments about updating the comments :smile:

core/src/main/java/com/radixdlt/consensus/sync/GetVerticesRequest.java
siy: Fixed

core/src/main/java/com/radixdlt/environment/Dispatchers.java
siy: Fixed.

core/src/main/java/com/radixdlt/environment/EventProcessorOnDispatch.java
siy: Fixed too.

LukasGasior1 (approved):
Nice! Looks good

PR #948, opened 2024-05-14
Apply a builder pattern for test configs

jakrawcz-rdx:
## Summary

This addresses https://radixdlt.atlassian.net/browse/NODE-645.

## Details

The main config items now use a builder pattern, and the tests only customize the parts that they care about.
The customization of nested items is still a bit awkward, but it is a much rarer case (and TBH the new shape is good enough, especially within the testing realm).

## Testing

Only test code is affected.


jakrawcz-rdx:
Interestingly, one of the flakey tests happened here: https://github.com/radixdlt/babylon-node/actions/runs/9081934995/job/24957048528?pr=948#step:9:2015
(and the newest commit uses the new config customization powers to disable the GC)

LukasGasior1 (approved):
Noice :+1:

PR #957, opened 2024-05-16
Pass the openjdk version directly (remove the no-longer-needed docker arg)

jakrawcz-rdx:
## Summary

This partially addresses the https://radixdlt.atlassian.net/browse/NODE-635 - it only rollbacks a remaining part of the old, no longer needed hack.
The true scope of that issue (make a version arg conditional on `TARGETPLATFORM`) has no use-case at the moment.

## Testing

Only a mechanical change in dockerfile.


mstrug-rdx (approved):
Looks ok, according to Jira description.

PR #959, opened 2024-05-17
Better separate protocol state vs current protocol version

jakrawcz-rdx:
## Summary

When working on "staged instantiation" (needed for https://radixdlt.atlassian.net/browse/NODE-648) I noticed a refactoring opportunity: the "current protocol version" wasn't really updated by `TransactionTracker`, and it was a bit confusing that it is a part of `ProtocolState`. Now it is a separate piece of `ProtocolManager`'s state, clearly updated only as the last step of a protocol update execution process.

## Details

Please see one `TODO` that I added after noticing a theoretical deficiency of the today's `TransactionSeriesExecutor`. I really believe we should address it soon, if we don't want to be surprised.

## Testing

No behavior changes, only regression tests need to pass.


jakrawcz-rdx:
(the failing check is sonar complaining about `ImmutableMap.of(), ImmutableList.of()` which cannot be replaced though 🤷 )

iamyulong (approved):

PR #962, opened 2024-05-22
Architecture: Export Radix Engine

talekhinezh:
Exports Radix Engine structures. Done so that clients can more easily integrate with both Node/RadixEngine structures without needing to deal with the dependency mess.


core-rust/state-manager/Cargo.toml
```diff
@@ -9,6 +9,7 @@ node-common = { path = "../node-common" }
 sbor = { workspace = true }
 radix-transactions = { workspace = true }
 radix-transaction-scenarios = { workspace = true }
+radix-clis = { workspace = true }
```
iamyulong, line 12: Any reason for including `radix-clis`?

iamyulong (commented):
Looks like our dependency security scanner is blocking the PR.

core-rust/state-manager/Cargo.toml
talekhinezh: There are some structures there which `ledger-tools` requires

iamyulong (approved):

PR #963, opened 2024-05-24
Change how RATE_LIMIT is reported in logs

siy:
Small change, which reduces the frequency of the RATE_LIMIT message and shifts it to INFO level. Remaining messages logged at DEBUG level.

### DIFF
--- a/core/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
+++ b/core/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
@@ -427,15 +427,17 @@ public final class BFTSync implements BFTSyncer {
     var syncRequestState = bftSyncing.getOrDefault(request, new SyncRequestState(authors, round));
 
     if (syncRequestState.syncIds.isEmpty()) {
+      var author = authors.get(0);
+
       if (this.syncRequestRateLimiter.tryAcquire()) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
```diff
@@ -427,21 +428,29 @@ private void sendBFTSyncRequest(
...
+        } else if (log.isDebugEnabled()) {
+          log.debug(buildMessage(reason, round, authors, request));
+        }
       }
       this.bftSyncing.put(request, syncRequestState);
     }
     syncRequestState.syncIds.add(syncId);
   }
 
+  private String buildMessage(
```
LukasGasior1, line 446: I'd rename it to something more specific, e.g. `outboundRateLimitLogMessage`

LukasGasior1: And pass an actual author as a parameter, rather than the whole list. It's not this method's responsibility to know that `authors.get(0)` was used.

siy: Good point. Renamed.

siy: Extracted local variable as it is used in at least 3 places.

siy:
Reworked into another PR

PR #977, opened 2024-06-13
Make call-preview endpoint work without working faucet or with empty …

siy:
[Fix `call-preview` endpoint.](https://radixdlt.atlassian.net/browse/PRO-3)

Existing `call-preview` endpoint is unable to provide a preview if the faucet is not available or empty (i.e. in production environment). This PR addresses this issue and makes the endpoint work regardless of the presence of the faucet.

### DIFF
--- a/core/src/test/java/com/radixdlt/api/core/CallPreviewTest.java
+++ b/core/src/test/java/com/radixdlt/api/core/CallPreviewTest.java
@@ -64,124 +64,42 @@
 
 package com.radixdlt.api.core;
 
-import static com.radixdlt.environment.deterministic.network.MessageSelector.firstSelector;
-import static com.radixdlt.lang.Tuple.tuple;
 import static org.junit.Assert.assertEquals;
 
-import com.google.common.collect.ImmutableList;
-import com.google.inject.AbstractModule;
-import com.google.inject.Module;
-import com.google.inject.multibindings.ProvidesIntoSet;
 import com.radixdlt.addressing.Addressing;
-import com.radixdlt.api.CoreApiServer;
-import com.radixdlt.api.CoreApiServerModule;
-import com.radixdlt.api.DummySslContextFactory;
+import com.radixdlt.api.DeterministicCoreApiTestBase;
 import com.radixdlt.api.core.generated.api.TransactionApi;
-import com.radixdlt.api.core.generated.client.ApiClient;
 import com.radixdlt.api.core.generated.client.ApiException;
-import com.radixdlt.api.core.generated.models.*;
+import com.radixdlt.api.core.generated.models.BlueprintFunctionTargetIdentifier;
+import com.radixdlt.api.core.generated.models.TargetIdentifierType;
+import com.radixdlt.api.core.generated.models.TransactionCallPreviewRequest;
+import com.radixdlt.api.core.generated.models.TransactionStatus;
 import com.radixdlt.crypto.ECKeyPair;
-import com.radixdlt.environment.CoreApiServerFlags;
-import com.radixdlt.environment.StartProcessorOnRunner;
-import com.radixdlt.environment.deterministic.network.MessageMutator;
 import com.radixdlt.genesis.*;
-import com.radixdlt.harness.deterministic.DeterministicTest;
-import com.radixdlt.harness.deterministic.PhysicalNodeConfig;
-import com.radixdlt.identifiers.Address;
-import com.radixdlt.modules.FunctionalRadixNodeModule;
 import com.radixdlt.modules.StateComputerConfig;
 import com.radixdlt.networks.Network;
 import com.radixdlt.rev2.Decimal;
 import com.radixdlt.rev2.ScryptoConstants;
-import com.radixdlt.utils.FreePortFinder;
-import com.radixdlt.utils.UInt32;
-import com.radixdlt.utils.UInt64;
-import java.net.http.HttpClient;
-import org.junit.Rule;
 import org.junit.Test;
-import org.junit.rules.TemporaryFolder;
 
-public final class CallPreviewWithoutFaucetTest {
-  private final ApiClient apiClient;
-  private final Module coreServerModule;
-
-  @Rule public TemporaryFolder folder = new TemporaryFolder();
-
-  public CallPreviewWithoutFaucetTest() {
-    var coreApiPort = FreePortFinder.findFreeLocalPort();
-
-    apiClient = buildApiClient(coreApiPort);
-    coreServerModule =
-        new AbstractModule() {
-          @Override
-          protected void configure() {
-            install(
-                new CoreApiServerModule("127.0.0.1", coreApiPort, new CoreApiServerFlags(true)));
-          }
-
-          @ProvidesIntoSet
-          private StartProcessorOnRunner startCoreApi(CoreApiServer coreApiServer) {
-            return new StartProcessorOnRunner("coreApi", coreApiServer::start);
-          }
-        };
-  }
-
-  private DeterministicTest createTest() {
+public final class CallPreviewTest extends DeterministicCoreApiTestBase {
+  private StateComputerConfig prepareConfig() {
     final var validatorKey = ECKeyPair.fromSeed(new byte[] {0x02}).getPublicKey();
 
-    var genesisData =
-        new GenesisData(
-            UInt64.fromNonNegativeLong(1L),
-            0,
-            GenesisConsensusManagerConfig.testingDefaultEmpty(),
-            ImmutableList.of(
-                new GenesisDataChunk.Validators(
-                    ImmutableList.of(
-                        new GenesisValidator(
-                            validatorKey,
-                            true,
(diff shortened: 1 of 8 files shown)


core/src/test/java/com/radixdlt/rev2/nofaucet/CallPreviewWithoutFaucetTest.java
```diff
@@ -0,0 +1,185 @@
...
+                            ImmutableList.of(
+                                new GenesisStakeAllocation(
+                                    UInt32.fromNonNegativeInt(0), Decimal.ONE)))))),
+            Decimal.ofNonNegative(0L),
+            ImmutableList.of());
+
+    return DeterministicTest.builder()
+        .addPhysicalNodes(PhysicalNodeConfig.createBatch(1, true))
+        .messageSelector(firstSelector())
+        .messageMutator(MessageMutator.dropTimeouts())
```
dhedey, line 130: Why are we dropping timeouts?

core/src/test/java/com/radixdlt/rev2/nofaucet/CallPreviewWithoutFaucetTest.java
```diff
@@ -0,0 +1,185 @@
...
+import com.radixdlt.rev2.Decimal;
+import com.radixdlt.rev2.ScryptoConstants;
+import com.radixdlt.rev2.protocol.ProtocolUpdateTestUtils;
+import com.radixdlt.utils.UInt32;
+import com.radixdlt.utils.UInt64;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+public final class CallPreviewWithoutFaucetTest {
```
dhedey, line 95: Out of interest, why didn't we use a test under `/api/core`, e.g. creating a `TransactionCallPreviewTest.java` similiar to `TransactionPreviewTest.java`?

In fact, can I request that we re-write the test there?

dhedey: You can then configure the test to run different scenarios, as per `test_core_api_can_submit_and_commit_transaction_after_running_all_scenarios`

core/src/test/java/com/radixdlt/rev2/nofaucet/CallPreviewWithoutFaucetTest.java
```diff
@@ -0,0 +1,185 @@
...
+  @Rule public TemporaryFolder folder = new TemporaryFolder();
+
+  private DeterministicTest createTest(Module... extraModules) {
```
dhedey, line 98: Lots of this configuration seems irrelevant and prone to breaking - can we either use some helper method like
`GenesisBuilder.createTestGenesisWithNumValidators` ?

core/src/test/java/com/radixdlt/rev2/nofaucet/CallPreviewWithoutFaucetTest.java
```diff
@@ -0,0 +1,185 @@
...
+    final var accountLockerCall =
+        new TransactionCallPreviewRequest()
+            .network(Network.INTEGRATIONTESTNET.getLogicalName())
+            .target(
+                new BlueprintFunctionTargetIdentifier()
+                    .packageAddress(addressing.encode(ScryptoConstants.RESOURCE_PACKAGE_ADDRESS))
+                    .blueprintName("FungibleResourceManager")
+                    .functionName("create")
+                    .type(TargetIdentifierType.FUNCTION))
+            .addArgumentsItem("4d220000")
```
dhedey, line 165: (Side note, irrelevant to this PR!) Hah oh - I forgot how awful this interface is. I believe there's a ticket in the backlog to allow using manifest syntax for this in a new endpoint...

core/src/test/java/com/radixdlt/rev2/nofaucet/CallPreviewWithoutFaucetTest.java
```diff
@@ -0,0 +1,185 @@
...
+            .addArgumentsItem("4d0100")
+            .addArgumentsItem("4d0700")
+            .addArgumentsItem("4d2106220000220000220000220000220000220000")
+            .addArgumentsItem("4d2102230c2100230c2200")
+            .addArgumentsItem("4d220000");
+
+    final var coreApiHelper = new ProtocolUpdateTestUtils.CoreApiHelper();
```
dhedey, line 172: We're not in a protocol update test, so feels a little weird to use this.

core/src/test/java/com/radixdlt/rev2/nofaucet/CallPreviewWithoutFaucetTest.java
```diff
@@ -0,0 +1,185 @@
...
+    try (var test = createTest(coreApiHelper.module())) {
+      // Arrange: Start a single node network
+      test.startAllNodes();
+
+      // Act: Preview a transaction
+      final var callBeforeBottlenose =
```
dhedey, line 178: How do we know we're actually before bottlenose?

dhedey: I'd much rather we make it explicit / clear.

e.g. we can set a really long epoch in genesis so that anemone doesn't get enacted (for now) and add a comment explaining _why_ we have set the long epoch.

(And then, in the long-term, we should be able to generally reconfigure protocol update running in the test-runner, e.g. create a: `integration_test_no_scenarios_protocol_config() -> ProtocolConfig`, or use the `ProtocolUpdateContentOverrides` mechanism - or something, like we can now do on the rust side.

core/src/test/java/com/radixdlt/rev2/nofaucet/CallPreviewWithoutFaucetTest.java
dhedey: In fact, if we just use a core api test base, we can get rid of lots of this.

dhedey (changes requested):
Nice functional changes - but I'd like us to make the test less boiler-platey and make the hacks clearer with some comments 👍

core/src/test/java/com/radixdlt/rev2/nofaucet/CallPreviewWithoutFaucetTest.java
siy: These are leftovers from the test I've used as a base.

core/src/test/java/com/radixdlt/rev2/nofaucet/CallPreviewWithoutFaucetTest.java
siy: I'll move test there, but this is rather organizational change. This test is specific in the sense that it has different genesis, hence reusing existing classes gets quite inconvenient.

core/src/test/java/com/radixdlt/rev2/nofaucet/CallPreviewWithoutFaucetTest.java
siy: Yes, but this test requires the faucet to be empty, which is not the case for the majority of tests.

core/src/test/java/com/radixdlt/rev2/nofaucet/CallPreviewWithoutFaucetTest.java
siy: Actually, this is also leftover from the test I've used as a base.

core/src/test/java/com/radixdlt/rev2/nofaucet/CallPreviewWithoutFaucetTest.java
dhedey: > I'll move test there, but this is rather organizational change. This test is specific in the sense that it has different genesis, hence reusing existing classes gets quite inconvenient.

`test_core_api_can_submit_and_commit_transaction_after_running_all_scenarios` which I mentioned above uses a different genesis alongside various helpers and builders which reduces the boilerplate for such changes. It can be used as an example for such a change.

core/src/test/java/com/radixdlt/api/core/CallPreviewTest.java
```diff
@@ -0,0 +1,142 @@
...
+import com.radixdlt.api.core.generated.models.TransactionStatus;
+import com.radixdlt.crypto.ECKeyPair;
+import com.radixdlt.genesis.*;
+import com.radixdlt.modules.StateComputerConfig;
+import com.radixdlt.networks.Network;
+import com.radixdlt.rev2.Decimal;
+import com.radixdlt.rev2.ScryptoConstants;
+import org.junit.Test;
+
+public final class CallPreviewTest extends DeterministicCoreApiTestBase {
```
dhedey, line 85: Much better without all the boilerplate, thanks!

dhedey (approved):

PR #979, opened 2024-07-15
Support uninstantiated entities in a unified way across ES API endpoints

jakrawcz-rdx:
## Summary

Addresses https://radixdlt.atlassian.net/browse/NTM-72.

## Details

Previously the uninstantiated entities were only supported by `/entity/info` endpoint, via a piece of custom code (catch the "NodeId not found" case, force instantiation, load data from staged store).
After this change, this is now done for all endpoints. Doing this required moving the "force instantiation" mechanism to a preparatory step before the actual (possibly multiple) loads needed by the endpoint - hence, the business "loaders" now have a common factory responsible for that (== see the meat of this PR at `EngineStateLoaderFactory`).

## Testing

Added tests for all applicable endpoints which should see data of uninstantiated entity.

### DIFF
--- a/core-rust/engine-state-api-server/src/engine_state_api/factories.rs
+++ b/core-rust/engine-state-api-server/src/engine_state_api/factories.rs
@@ -1,0 +1,210 @@
+use crate::engine_prelude::*;
+
+use radix_engine::transaction::{
+    execute_and_commit_transaction, CommitResult, ExecutionConfig, StateUpdateSummary,
+    TransactionResult,
+};
+use radix_engine::vm::wasm::DefaultWasmEngine;
+use radix_engine::vm::{NoExtension, ScryptoVm, VmInit};
+use radix_substate_store_impls::substate_database_overlay::{
+    SubstateDatabaseOverlay, UnmergeableSubstateDatabaseOverlay,
+};
+
+use super::*;
+
+/// A factory of various "loaders" used by Engine State API.
+///
+/// Instantiating a loader is sometimes not trivial, since it may require e.g. a "staged
+/// instantiation" of an uninstantiated entity - hence most endpoint handlers should interact with
+/// this factory.
+pub struct EngineStateLoaderFactory<'s, S> {
+    database: UnmergeableSubstateDatabaseOverlay<'s, S>,
+    staged_node_ids: IndexSet<NodeId>,
+}
+
+impl<'s, S: SubstateDatabase> EngineStateLoaderFactory<'s, S> {
+    /// Creates a factory using the given database.
+    pub fn new(database: &'s S) -> Self {
+        Self {
+            database: SubstateDatabaseOverlay::new_unmergeable(database),
+            staged_node_ids: index_set_new(),
+        }
+    }
+
+    /// Performs a "staged instantiation" of the given entity, if required.
+    ///
+    /// The uninstantiated instance will be "forced" into being by executing a simple transaction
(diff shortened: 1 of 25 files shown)


core-rust/engine-state-api-server/src/engine_state_api/factories.rs
```diff
@@ -0,0 +1,210 @@
...
+        let intent = create_intent_forcing_instantiation(node_id);
+
+        // Note: here, we initialize a new `ScryptoVm` instance every time - however, this is
+        // currently a very lightweight operation. At the same time, we would NOT be able to feel
+        // the benefits of caching an instance (since we do not use any WASM here). If these
+        // assumptions change in future, then it will make most sense to turn the `Self` into a
+        // long-lived service holding a `ScryptoVm` as its dependency.
+        let scrypto_vm = ScryptoVm::<DefaultWasmEngine>::default();
+
+        let receipt = execute_and_commit_transaction(
```
iamyulong, line 132: Who came up with this instantiation through manifest solution lol!

A small side effect of this approach is an update to the existing transaction tracker. Would this affect any of engine state APIs?

core-rust/node-common/src/jni/addressing.rs
```diff
@@ -158,6 +158,24 @@ extern "system" fn Java_com_radixdlt_identifiers_Address_nativeVirtualAccountAdd
...
+#[no_mangle]
+extern "system" fn Java_com_radixdlt_identifiers_Address_nativeVirtualIdentityAddress(
+    env: JNIEnv,
+    _class: JClass,
+    request_payload: jbyteArray,
+) -> jbyteArray {
+    jni_sbor_coded_call(
+        &env,
+        request_payload,
+        |unchecked_public_key_bytes: [u8; Secp256k1PublicKey::LENGTH]| {
```
iamyulong, line 170: How about ed25519?

jakrawcz-rdx: I only added this method for test purposes (I need to obtain some identity address to test stuff with), and I arbitrarily chosen Secp256k1.
(in a Rest API-level test I don't aim to cover all cases)

core-rust/engine-state-api-server/src/engine_state_api/factories.rs
jakrawcz-rdx: This approach was decided before https://github.com/radixdlt/babylon-node/pull/978 after some discussion with Josh (see https://radixdlt.atlassian.net/browse/NODE-648?focusedCommentId=20074). Other alternatives would either require either a too-big refactor in the Engine, or a too-error-prone hardcoding of values in the Node, so here we are...

Conceptually, it is the most bulletproof way (i.e. it cannot go wrong, it returns exactly what a real transaction would do). Practically, its runtime cost is acceptable (see https://github.com/radixdlt/babylon-node/pull/978#issuecomment-2194189897).

> A small side effect of this approach is an update to the existing transaction tracker. Would this affect any of engine state APIs?

It definitely cannot affect the behavior of current endpoints.
If we ever add some kind of "batch" endpoint (e.g. read the state of TransactionTracker + read the state of non-instantiated Account in the same call), then we will have to be careful (i.e. isolate those reads; do not re-use the staged store).

iamyulong (approved):

jakrawcz-rdx:
Only the Phylum part of the build is failing, but probably because of some misconfiguration (see https://rdxworks.slack.com/archives/C02EQ2E7544/p1721394003538299), so lemme merge.

PR #982, opened 2024-07-29
Move safety store to rust

siy:
## Summary

This PR moves Address Book and Safety Store to RocksDB (i.e. at Rust side of the Node code). The migration is of the existing stores is performed automatically at startup.


core/src/test-core/java/com/radixdlt/modules/FunctionalRadixNodeModule.java
```diff
@@ -316,7 +325,30 @@ public void configure() {
 
     switch (this.safetyRecoveryConfig) {
       case MOCKED -> install(new MockedSafetyStoreModule());
-      case BERKELEY_DB -> install(new BerkeleySafetyStoreModule());
+      case REAL -> install(
```
dhedey, line 328: Shouldn't this use Rocks, not Berkeley?

cli-tools/src/main/java/com/radixdlt/shell/RadixShell.java
```diff
@@ -178,9 +178,10 @@ public Node build() throws Exception {
         dataDir = new File(Files.createTempDirectory("radix-shell-node-").toString());
       }
 
-      customProperties.build().forEach((k, v) -> properties.set(k, v));
+      customProperties.build().forEach(properties::set);
 
       properties.set("db.location", dataDir.toString());
+      properties.set("db.location.node", new File(dataDir, "node").toString());
```
dhedey, line 184: In my opinion we shouldn't be introducing explicit configuration for e.g. the `db.location.node` field - it should just be a fixed path under `db.location`. Having the layout fixed under `db.location` means we don't have to worry about people changing this when people build backup scripts.

Side note - if we're introducing a new config, it should be included in the env_subst file so it works in Docker.

core-rust/state-manager/src/jni/node_rust_environment.rs
```diff
@@ -122,13 +125,16 @@ pub struct JNINodeRustEnvironment {
     pub state_manager: StateManager,
     pub metric_registry: Arc<Registry>,
     pub running_task_tracker: UntilDropTracker,
+    pub node_store: Arc<DbLock<ActualNodeDatabase>>,
```
dhedey, line 128: Three things:
* I thought we were having _separate_ databases for Safety and Address Book (and possibly migrations)?
* If we did have one database, in my opinion we shouldn't call it the "node database" - it's a very misleading name as it doesn't include e.g. the ledger database.
* I don't see why we need a lock around them here - this seems like it will be uselessly contentioned (particularly with a combined database, where e.g. a write of the address book would block reading the safety store) -  Rocks does locking internally anyway where it needs to.

core/src/main/java/com/radixdlt/p2p/addressbook/BerkeleyAddressBookStore.java
```diff
@@ -101,7 +101,7 @@ public final class BerkeleyAddressBookStore implements AddressBookPersistence {
   public BerkeleyAddressBookStore(
       Serialization serialization,
       Metrics metrics,
-      @NodeStorageLocation String nodeStorageLocation,
+      @StateManagerStorageLocation String nodeStorageLocation,
```
dhedey, line 104: I don't understand this - We're effectively moving the Berkeley address book here? 

I'd assumed we only want this for the purposes of migrating to the new address book store? So it should use the old location?

core-rust/state-manager/src/store/address_book_components.rs
```diff
@@ -0,0 +1,40 @@
+use crate::engine_prelude::Sbor;
+
+
+/// The ID of the node stored in the Address book (Secp256k1 public key)
+#[derive(Clone, Copy, Sbor)]
+pub struct AddressBookNodeId(pub [u8; AddressBookNodeId::LENGTH]);
```
dhedey, line 6: This stuff shouldn't be in the `state-manager` crate as it's not relevant to the ledger / state manager database.

We may want to introduce new crates for `consensus` to stick the safety store, and `p2p` for the address book?

Where `jni-export` includes `consensus`, `p2p` etc.

We may then want to move some of the ledger helpers into `node-common`.

core-rust/state-manager/src/store/address_book_components.rs
```diff
@@ -0,0 +1,40 @@
...
+pub enum ConnectionStatus {
+    Connected,
+    Disconnected,
+}
+
+/// Address book entry
+#[derive(Clone, Sbor)]
+pub struct AddressBookEntry {
+    pub node_id: AddressBookNodeId,
+    pub banned_until: Option<i64>,
```
dhedey, line 38: If that's a timestamp we probably want to create an `EpochTimestamp` new-type wrapping e.g. chrono's [DateTime<Utc>](https://docs.rs/chrono/latest/chrono/struct.DateTime.html) - with this to make it SBOR encode as an i64:

```
#[sbor(
    as_type = "i64",
    as_ref = "self. timestamp()",
    from_value = "Self(DateTime::from_timestamp(value, 0))"
)]
```

core-rust/state-manager/src/store/column_families.rs
```diff
@@ -0,0 +1,471 @@
...
+    type Value = DbSubstateValue;
+
+    const DEFAULT_NAME: &'static str = "associated_state_tree_values";
+    type KeyCodec = StoredTreeNodeKeyDbCodec;
+    type ValueCodec = DirectDbCodec;
+}
+
+/// Address book and safety state store migration status. Filled once during the migration.
+pub struct MigrationStatusCf;
+impl DefaultCf for MigrationStatusCf {
```
dhedey, line 432: So (1) these new CFs shouldn't be in `state-manager` and (2) we should make sure they're versioned, so we can update the models in future! :)

core-rust/state-manager/src/store/typed_cf_api.rs
```diff
@@ -738,6 +767,41 @@ impl DbCodec<()> for UnitDbCodec {
     }
 }
 
+/// A [`DbCodec]` capable of representing [`MigrationId]`.
```
dhedey, line 770: Out of interest, why did we opt for manual codecs rather than SBOR encoding the id and migration status?

dhedey (changes requested):
Thanks Sergiy! For this review, I've just started with a high level / architectural review. Happy to do a more detailed review once the high level stuff feels correct.

In terms of the architecture, I think we need to move around the crate structure on the Rust side, and separate out the databases better -- I've captured more detailed thoughts in comments.

The code itself is looking nice though, we just want to make sure the architecture is right first.

core-rust/state-manager/src/jni/node_rust_environment.rs
siy: Perhaps this is a misunderstanding because I got the conclusion that these stores should be separate from the ledger, but otherwise it's fine to keep them together. But I see no problems separating them now.

core/src/main/java/com/radixdlt/p2p/addressbook/BerkeleyAddressBookStore.java
siy: This is to preserve the Berkeley address book location where it was before. It might be a bit misleading as the old `NodeStorageLocation` annotation is renamed to `StateManagerStorageLocation` and the new `NodeStorageLocation` annotation is now responsible for node-specific data - address book and safety state.

core-rust/state-manager/src/store/address_book_components.rs
siy: Makes sense. Will refactor these.

core-rust/state-manager/src/store/address_book_components.rs
siy: Thanks for the hint.

core-rust/state-manager/src/store/column_families.rs
siy: Makes sense. Will update.

core-rust/state-manager/src/store/typed_cf_api.rs
siy: Good point. I've created them at the beginning, when I had a very basic understanding of how things done here. Will change to SBOR encoding.

cli-tools/src/main/java/com/radixdlt/shell/RadixShell.java
siy: Reworked to use the same config variable and fixed subdirectories under it.

core/src/main/java/com/radixdlt/p2p/addressbook/BerkeleyAddressBookStore.java
siy: Restored old naming

core-rust/state-manager/src/store/typed_cf_api.rs
siy: Fixed

core-rust/state-manager/src/store/address_book_components.rs
siy: After looking deeper, I realized that implementation must make assumptions about future use and convenient representation. Also, it requires adding a new dependency on `crono` just to make the code compile, there is no use of this representation on the Rust side for now. So, I've created a type alias and added a comment how this can be changed if necessary. Actual change, better to postpone until these timestamps will be actually used on the Rust side. Someone who will be implementing this part will be free to choose the most convenient underlying type without breaking binary compatibility.

core/src/test-core/java/com/radixdlt/modules/FunctionalRadixNodeModule.java
siy: Fixed.

core/src/test/java/com/radixdlt/rev2/RustMempoolTest.java
```diff
@@ -383,6 +302,51 @@ public void test_rust_mempool_getRelayTxns() throws Exception {
     }
   }
 
+  private NodeRustEnvironment createNodeRustEnvironment() throws IOException {
```
dhedey, line 305: I wonder if it's worth moving this to some unit test helper or something? Given that we've duplicated it across a few tests now?

dhedey: With the tests, we've been following the "Good Scout" rule and leaving our Java tests in a better shape than we found them every change.

core/src/test/java/com/radixdlt/consensus/safety/RocksSafetyStateStoreTest.java
dhedey, line 1: I can't see any tests of the migration?

I'd like to see a test where we:
* Create a Berkeley store (e.g. just via a constructor) and commit various bits of data
* Boot up a node. Wait for it to boot up. This should trigger the migration of the data.
* Delete the data from the Berkeley store.
* Check the state can be read from the node (ie that it must have migrated to Rust)

dhedey: (for each of the stores).

Booting up the node can be done with a `SimulationTest` builder, setting a FunctionalRadixNodeModule set to use the same temporary folder for its database as the Berkeley store.

core-rust/p2p/src/rocks_db.rs
```diff
@@ -0,0 +1,245 @@
...
+pub type ActualMigrationDatabase = MigrationDatabase<DirectRocks>;
+
+/// A RocksDB-backed persistence layer for address book.
+pub struct AddressBookDatabase<R> {
+    /// Underlying RocksDB instance.
+    rocks: R,
+}
+
+/// A RocksDB-backed persistence layer for safety store.
+pub struct SafetyStoreDatabase<R> {
```
dhedey, line 120: I'd expect the `SafetyStore` to be in a separate consensus crate. But I'm happy for us to move it at a later date if we just want to get this ticket in.

core/src/test-core/java/com/radixdlt/modules/PrefixedNodeStorageLocationModule.java
```diff
@@ -86,7 +86,7 @@ public PrefixedNodeStorageLocationModule(String baseLocation) {
   @Provides
   @Singleton
   @NodeStorageLocation
-  private String nodeStorageLocation(@Self ECDSASecp256k1PublicKey publicKey) {
+  private String stateManagerStorageLocation(@Self ECDSASecp256k1PublicKey publicKey) {
```
dhedey, line 89: I think this is still the `nodeStorageLocation`?

core/src/main/java/com/radixdlt/consensus/safety/MigratingSafetyStore.java
```diff
@@ -0,0 +1,81 @@
...
+ * or non-commercial purpose and for any reproduction or redistribution by You of the
+ * Work. You assume all risks associated with Your use of the Work and the exercise of
+ * permissions under this License.
+ */
+
+package com.radixdlt.consensus.safety;
+
+import java.util.Optional;
+
+public class MigratingSafetyStore implements PersistentSafetyStateStore {
```
dhedey, line 69: I think this is unused / can be deleted? 👍

dhedey (commented):
Nice work!

I haven't checked the DTO code -- ideally I'd like to see some tests that the migrations are working as expected, by writing to Berkeley, migrating, then reading from Rocks and comparing.

Separately, I think it'd be better in the long run to not have a migration store, and move the migration state into the individual stores; to avoid having to explain to node runners what this migration store is :)

core-rust/state-manager/src/jni/node_rust_environment.rs
```diff
@@ -163,18 +165,51 @@ impl JNINodeRustEnvironment {
 
         let running_task_tracker = scheduler.into_task_tracker();
 
+        let address_book_db_path = Self::combine(&base_path, "address_book");
+        let safety_store_db_path = Self::combine(&base_path, "safety_store");
```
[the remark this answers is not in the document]
siy: Fixed.

core/src/main/java/com/radixdlt/RadixNodeModule.java
```diff
@@ -391,7 +391,6 @@ protected void configure() {
             ScenariosExecutionConfig.resolveForNetwork(network)));
 
     // Recovery
-    install(new BerkeleySafetyStoreModule());
```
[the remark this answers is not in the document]
siy: Yes, it is. For some reason, git thinks that it was renamed into BFTValidatorDTO.

core/src/main/java/com/radixdlt/rev2/modules/NodePersistenceModule.java
```diff
@@ -0,0 +1,180 @@
...
+  @Provides
+  @Singleton
+  RocksDbAddressBookStore rocksDbAddressBookStore(
+      Metrics metrics, NodeRustEnvironment environment) {
+    return RocksDbAddressBookStore.create(metrics, environment);
+  }
+
+  @Provides
+  @Singleton
+  RocksDbHighPriorityPeersStore rocksDbHighPriorityPeersStore(
```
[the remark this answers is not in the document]
siy: `RocksDbAddressBookStore` holds only the address book part. I wanted to keep all stores separate. On the Java side, putting it all together looked like an ad-hoc solution.

core/src/test/java/com/radixdlt/rev2/RustMempoolTest.java
siy: Reworked tests to reuse common code. Unfortunately, due to Gradle issues it appeared too time-consuming to attempt to reuse test dependencies, and I ought to create two similar (but not identical) base classes for tests which require NodeRustEnvironment.

core/src/main/java/com/radixdlt/consensus/safety/MigratingSafetyStore.java
siy: Removed file.

core-rust/p2p/src/rocks_db.rs
siy: I think further refactoring (including moving p2p JNI part) should be done at the next stage, after discussion and making decision on final structure.

core/src/main/java/com/radixdlt/rev2/modules/NodePersistenceModule.java
```diff
@@ -0,0 +1,180 @@
...
+
+  @Provides
+  @Singleton
+  PersistentSafetyStateStore persistentSafetyStateStore(
+      RocksDbMigrationStore migrationStore,
+      BerkeleySafetyStateStore berkeleySafetyStateStore,
+      RocksSafetyStateStore rocksSafetyStateStore) {
+
+    try (berkeleySafetyStateStore) {
+      if (!migrationStore.isMigrated(StoreId.SAFETY_STORE)) {
```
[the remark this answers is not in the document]
siy: Yes, it makes sense. Will rework it.

core/src/main/java/com/radixdlt/rev2/modules/NodePersistenceModule.java
```diff
@@ -0,0 +1,180 @@
...
+  AddressBookPersistence addressBookPersistence(
+      RocksDbMigrationStore migrationStore,
+      BerkeleyAddressBookStore berkeleyAddressBookStore,
+      RocksAddressBookStore addressBookStore) {
+
+    try (berkeleyAddressBookStore) {
+      if (!migrationStore.isMigrated(StoreId.ADDRESS_BOOK)) {
+        berkeleyAddressBookStore.getAllEntries().forEach(addressBookStore::upsertEntry);
+        addressBookStore.storeHighPriorityPeers(berkeleyAddressBookStore.getHighPriorityPeers());
+        migrationStore.migrationDone(StoreId.ADDRESS_BOOK);
```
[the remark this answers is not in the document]
siy: Fixed

core/src/main/java/com/radixdlt/rev2/modules/NodePersistenceModule.java
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  @Provides
+  @Singleton
+  AddressBookPersistence addressBookPersistence(
+      RocksDbMigrationStore migrationStore,
+      BerkeleyAddressBookStore berkeleyAddressBookStore,
+      RocksAddressBookStore addressBookStore) {
+
+    try (berkeleyAddressBookStore) {
```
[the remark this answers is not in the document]
siy: Reworked to use migration status inside each store.

core/src/test/java/com/radixdlt/rev2/RustMempoolTest.java
dhedey: This is what `core/src/test-core` is for :). Create a new file `NodeRustEnvironmentBuilder.java` in `core/src/test-core/java/com/radixdlt/rev2`.

We've agreed as a team to use composition over inheritance for any new test utilities, so please don't introduce base classes :)

core/src/main/java/com/radixdlt/rev2/modules/NodePersistenceModule.java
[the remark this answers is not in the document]
siy: Reworked

core/src/test/java/com/radixdlt/rev2/RustMempoolTest.java
siy: Such a location does not allow using it from `core-rust-bridge`, where 3 tests are located.

siy: Reworked to have it as a utility class, rather than a base class.

core-rust/p2p/src/column_families.rs
```diff
@@ -0,0 +1,50 @@
...
+    const DEFAULT_NAME: &'static str = "address_book";
+    type KeyCodec = AddressBookNodeIdDbCodec;
+    type ValueCodec = DirectDbCodec;
+}
+
+/// Safety store
+pub struct SafetyStoreCf;
+impl DefaultCf for SafetyStoreCf {
+    type Key = ();
+    type Value = Vec<u8>;
```
dhedey, line 35: I've just realised we're storing all these values as raw Vec<u8> instead of as upgradable SBOR. Why is that?

I'm concerned about this:
* DSON has built in support for adding new columns
* SBOR does not, so we need to add in something else to allow us to change our models over time. So far, our canonical choice has been wrapping things in a versioning enum. It's not ideal, but it's better than nothing

Just storing raw bytes without having the encoding/decoding models on the rust side also kinda defeats the purpose of moving the storage of this state into rust.

Could we tweak these to use `impl VersionedCf` instead?

core-rust/state-manager/src/jni/p2p/high_priority_peers.rs
```diff
@@ -0,0 +1,116 @@
...
+#[no_mangle]
+extern "system" fn Java_com_radixdlt_p2p_RocksDbHighPriorityPeersStore_upsertAllHighPriorityPeers(
+    env: JNIEnv,
+    _class: JClass,
+    j_rust_global_context: JObject,
+    payload: jbyteArray,
+) -> jbyteArray {
+    jni_raw_sbor_fallible_call(&env, payload, |bytes| {
+        JNINodeRustEnvironment::get_address_book_database(&env, j_rust_global_context)
+            .upsert_all_peers(&bytes);
```
dhedey, line 84: This looks wrong. I think perhaps this function just needs to be renamed `upsert_all_high_priority_peers`, or is this a bug?

core-rust/state-manager/src/jni/p2p/high_priority_peers.rs
```diff
@@ -0,0 +1,116 @@
...
+#[no_mangle]
+extern "system" fn Java_com_radixdlt_p2p_RocksDbHighPriorityPeersStore_getAllHighPriorityPeers(
+    env: JNIEnv,
+    _class: JClass,
+    j_rust_global_context: JObject,
+    payload: jbyteArray,
+) -> jbyteArray {
+    jni_sbor_coded_call(&env, payload, |_: ()| -> Option<Vec<u8>> {
+        JNINodeRustEnvironment::get_address_book_database(&env, j_rust_global_context)
+            .get_all_peers()
```
dhedey, line 98: Ditto - this looks wrong. I think perhaps this function just needs to be renamed `get_all_high_priority_peers`, or is this a bug?

core-rust/state-manager/src/store/column_families.rs
```diff
@@ -0,0 +1,428 @@
...
+    BlueprintAndCreationIndexKeyDbCodec, EpochDbCodec, HashDbCodec, NodeIdDbCodec,
+    PrefixGlobalAddressDbCodec, RawLedgerTransactionDbCodec, ScenarioSequenceNumberDbCodec,
+    StateVersionDbCodec, StoredTreeNodeKeyDbCodec, SubstateKeyDbCodec,
+    TypeAndCreationIndexKeyDbCodec,
+};
+use node_common::store::typed_cf_api::{
+    DefaultCf, DirectDbCodec,
+    PredefinedDbCodec, TypedCf, UnitDbCodec, VersionedCf,
+};
+use crate::transaction::{LedgerTransactionHash, RawLedgerTransaction};
```
dhedey, line 39: We generally prefer wildcard imports `*` on the above, so that diffs are less noisy

core-rust/state-manager/src/store/rocks_db.rs
```diff
@@ -1347,11 +752,11 @@ impl<R: WriteableRocks> ExecutedScenarioStore for StateManagerDatabase<R> {
 
 pub struct RocksDBCommittedTransactionBundleIterator<'r> {
     state_version: StateVersion,
-    txns_iter: Box<dyn Iterator<Item = (StateVersion, RawLedgerTransaction)> + 'r>,
-    ledger_receipts_iter: Box<dyn Iterator<Item = (StateVersion, LedgerTransactionReceipt)> + 'r>,
-    local_executions_iter: Box<dyn Iterator<Item = (StateVersion, LocalTransactionExecution)> + 'r>,
+    txns_iter: Box<dyn Iterator<Item=(StateVersion, RawLedgerTransaction)> + 'r>,
```
dhedey, line 755: It looks like some weird non-standard formatter ran on this file, could we revert these changes?

I'd imagine the standard formatter should revert them, but maybe the changes aren't in the standard formatter's preference.

Still, we use spaces around our `=` and cargo format automatically indents the `for` when it puts it on a new line, so lets keep that.

core/src/main/java/com/radixdlt/rev2/modules/NodePersistenceModule.java
```diff
@@ -67,36 +67,62 @@
...
-      @NodeStorageLocation String nodeStorageLocation) {
-    return new BerkeleySafetyStateStore(
-        serialization,
-        metrics,
-        nodeStorageLocation,
-        BerkeleyDbDefaults.createDefaultEnvConfigFromProperties(properties));
+      @NodeStorageLocation String nodeStorageLocation,
+      NodeRustEnvironment environment) {
+
+    var berkeleySafetyStateStore =
```
dhedey, line 109: Can we move the creation of the `berkeleySafetyStateStore` inside the `!rocksSafetyStateStore.isMigrated()` check?

That way, we can avoid wasting time initializing Berkeley on start-up (and also I imagine avoid creating all the Berkeley files on disk) once we have migrated.

core/src/main/java/com/radixdlt/rev2/modules/AddressBookModule.java
```diff
@@ -0,0 +1,113 @@
...
+public class AddressBookModule extends AbstractModule {
+  @Provides
+  @Singleton
+  AddressBookPersistence addressBookPersistence(
+      RuntimeProperties properties,
+      Serialization serialization,
+      Metrics metrics,
+      @NodeStorageLocation String nodeStorageLocation,
+      NodeRustEnvironment environment) {
+    var berkeleyAddressBookStore =
```
dhedey, line 91: Can we move the creation of the `berkeleyAddressBookStore` inside the !isMigrated() check?

That way, we can avoid wasting time initializing Berkeley on start-up (and also I imagine avoid creating all the Berkeley files on disk) once we have migrated.

core/src/main/java/com/radixdlt/rev2/modules/REv2StateManagerModule.java
```diff
@@ -314,6 +318,11 @@ EventProcessor<BFTInsertUpdate> onInsertUpdatePersistVertexStore(
       bind(new Key<MempoolInserter<RawNotarizedTransaction>>() {}).to(RustMempool.class);
       bind(MempoolReevaluator.class).to(RustMempool.class);
     }
+
+    if (installPersistence) {
```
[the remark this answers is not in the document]
dhedey: This wasn't addressed in this new review.

core/src/test-core/java/com/radixdlt/modules/FunctionalRadixNodeModule.java
```diff
@@ -123,7 +120,8 @@ static NodeStorageConfig file(File folder) {
 
   public enum SafetyRecoveryConfig {
     MOCKED,
-    BERKELEY_DB,
+    REAL,
+    REAL_WITH_ADDRESS_BOOK
```
dhedey, line 124: Address book is completely unrelated to the safety store. We shouldn't add this as an option.

core/src/main/java/com/radixdlt/rev2/modules/REv2StateManagerModule.java
```diff
@@ -188,7 +191,8 @@ public static REv2StateManagerModule createForTesting(
         ledgerSyncLimitsConfig,
         protocolConfig,
         noFees,
-        scenariosExecutionConfig);
+        scenariosExecutionConfig,
+        false);
```
dhedey, line 195: Why does a testing state manager module not install persistence? This is surprising to me?

core/src/main/java/com/radixdlt/rev2/modules/REv2StateManagerModule.java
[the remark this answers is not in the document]
dhedey: Can we remove this flag and configure this some other way which is less intrusive?

core/src/test/java/com/radixdlt/p2p/test/P2PTestNetworkRunner.java
```diff
@@ -188,6 +194,44 @@ private static Injector createInjector(
       throws ParseException {
     final var properties = RuntimeProperties.fromCommandLineArgs(new String[] {});
     return Guice.createInjector(
+        new AbstractModule() {
+          @Override
+          protected void configure() {
+            bind(AddressBookPersistence.class).to(BerkeleyAddressBookStore.class);
```
dhedey, line 200: Why are we using Berkeley here instead of Rocks / the default?

Did Rocks break? This feels like a scary change to me :).

Either this test should be changed to use FunctionalNodeModule if it's an integration test; or if that's too big a change, at least at a comment here to explain we're hacking in the deprecated address Berkeley stores for now because it's too disruptive to add a node rust environment here.

core/src/test/java/com/radixdlt/p2p/addressbook/AddressBookMigrationTest.java
```diff
@@ -0,0 +1,164 @@
...
+import java.util.Set;
+import org.apache.commons.cli.ParseException;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+public class AddressBookMigrationTest {
+  @Rule public TemporaryFolder folder = new TemporaryFolder();
+
+  private Module createCommonModule() {
```
dhedey, line 95: We've had a rule to avoid creating Guice injectors for new tests, outside of `FunctionalNodeModule` - because they're very flakey to changes.

Why can't we just create the `BerkeleyAddressBookStore` in a private method in the test code without DI?

dhedey: In fact, if we change the `AddressBookModule` to have a static `ensureMigrated(rocksAddressBookStore, <config>)` method, then we can just test that method here without creating any test injectors at all

core/src/test/java/com/radixdlt/p2p/addressbook/AddressBookMigrationTest.java
```diff
@@ -0,0 +1,164 @@
...
+
+  @Test
+  public void address_book_migrates_successfully() {
+    var injector =
+        Guice.createInjector(createCommonModule(), new AddressBookModule(), createBerkeleyModule());
+    var addresses = Set.of(newAddressBookEntry(1), newAddressBookEntry(2), newAddressBookEntry(3));
+
+    // Create some entries in the Berkeley address book
+    try (var berkeleyAddressBook = injector.getInstance(BerkeleyAddressBookStore.class)) {
+      addresses.forEach(berkeleyAddressBook::upsertEntry);
```
dhedey, line 157: Nice test - but I notice it's not complete? Can we also add some high priority peers?

core/src/test/java/com/radixdlt/consensus/safety/SafetyStoreMigrationTest.java
```diff
@@ -0,0 +1,170 @@
...
+import java.util.Map;
+import org.apache.commons.cli.ParseException;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+public class SafetyStoreMigrationTest {
+  @Rule public TemporaryFolder folder = new TemporaryFolder();
+
+  private Module createCommonModule() {
```
dhedey, line 95: Similar to the other test - can we avoid adding new guice modules in tests? They create a maintenance burden

core/src/test/java/com/radixdlt/consensus/safety/SafetyStoreMigrationTest.java
```diff
@@ -0,0 +1,170 @@
...
+      }
+    };
+  }
+
+  @Test
+  public void address_book_migrates_successfully() {
+    var injector =
+        Guice.createInjector(
+            createCommonModule(),
+            new NodePersistenceModule(AddressBookPolicy.DISABLED),
```
dhedey, line 154: The address book persistence state should be irrelevant to this test - something feels off here

dhedey (commented):
Thanks Sergiy - it's looking a lot better, we're nearly there! Most of my comments can be left to a follow-up review, but the following should be done before merge - as they are core parts of the migration logic / end-state:
* Changing the Cfs to be VersionedCfs over SBOR state (sorry I missed this before in such a large review)
* Ensuring we don't instantiate Berkeley stores during bootstrapping unless we need to because we're not migrated.
* Ensuring the address book migration test is complete (I think we're missing high priority peers at the moment)

core-rust-bridge/src/test/java/com/radixdlt/helper/NodeRustEnvironmentBuilder.java
```diff
@@ -0,0 +1,109 @@
...
+import com.radixdlt.protocol.ProtocolConfig;
+import com.radixdlt.rev2.NetworkDefinition;
+import com.radixdlt.transaction.LedgerSyncLimitsConfig;
+
+public final class NodeRustEnvironmentBuilder {
+  private NodeRustEnvironmentBuilder() {
+    throw new IllegalStateException("Can't construct");
+  }
+
+  public static NodeRustEnvironment createNodeRustEnvironment(String dbPath) {
```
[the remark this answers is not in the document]
siy: The Builder pattern makes sense when there are no (or very few) mandatory parameters. In this case, there is only one (and mandatory) parameter. The Factory Method better suits in such a use case.

core-rust/node-common/src/java/utils.rs
```diff
@@ -115,6 +115,21 @@ pub fn jni_sbor_coded_fallible_call<Args: ScryptoDecode, Response: ScryptoEncode
     .unwrap_or_else(std::ptr::null_mut)
 }
 
+/// A convenience method for the case when input should remain in the form of raw byte array
+/// (i.e. no SBOR decoding is needed). Output is still properly encoded with SBOR.
+pub fn jni_raw_sbor_fallible_call<Response: ScryptoEncode>(
```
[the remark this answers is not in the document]
siy: This is an SBOR encoded structure, and it is automatically SBOR-encode on the Java side. On the Rust side it has no use now and there is no sense decoding it because DB contains SBOR encoded struct anyway. There were two reasonable solutions 
- manually encode SBOR on the Java side, wrap it into array and when it comes back - manually unwrap and decode
- get direct access to a byte array before it is decoded on the Rust side and pass it to the relevant underlying method

The second approach looked much simpler to me.

core-rust-bridge/src/test/java/com/radixdlt/helper/NodeRustEnvironmentBuilder.java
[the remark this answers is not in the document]
siy: Renamed class to avoid confusion.

core-rust/p2p/src/column_families.rs
siy: The current implementation does not send serialized `DSON` on the Rust side and does not rely on `DSON` capabilities. Instead, it sends SBOR-encoded `SafetyStateDTO`. This DTO and all other DTO's referred by `SafetyStateDTO` are a prototype for Rust-side implementation once the rest of the code will be moved to Rust. Currently, we don't need to decode this struct, so there is no need to implement all these DTO's on the Rust side. Part of the structs already present on the Rust side, but require some changes (like adding annotations). Other part is missing, but there are no relevant modules/infrastructure for the correct placement for these structs. Putting them all together looked impractical too.

core-rust/state-manager/Cargo.toml
```diff
@@ -27,7 +29,7 @@ blake2 = { workspace = true }
 hex = { workspace = true }
 rand = { workspace = true }
 
-rocksdb = { version = "=0.21.0" }
+#rocksdb = { version = "=0.21.0" }
```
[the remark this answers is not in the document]
siy: Removed

core-rust/state-manager/src/store/column_families.rs
siy: Fixed

core-rust/state-manager/src/store/rocks_db.rs
siy: Fixed.

core/src/main/java/com/radixdlt/rev2/modules/NodePersistenceModule.java
siy: Done

core/src/main/java/com/radixdlt/rev2/modules/AddressBookModule.java
siy: Done

core/src/test-core/java/com/radixdlt/modules/FunctionalRadixNodeModule.java
siy: Reworked.

core/src/main/java/com/radixdlt/rev2/modules/REv2StateManagerModule.java
siy: It was a surprise to me too, but it appeared that `FunctionalRadixNodeModule` unconditionally installs mocked `AddressBookPersistence` and this clashes with the persistence configured here.

core/src/test/java/com/radixdlt/p2p/test/P2PTestNetworkRunner.java
siy: I've tried to replace the implementation here, but got a very strange error:
```
com.google.inject.CreationException: Unable to create injector, see the following errors:

1) An exception was caught and reported. Message: called `Result::unwrap()` on an `Err` value: Error { message: "IO error: lock hold by current process, acquire time 1724758428 acquiring thread 6169243648: /var/folders/kq/xrwzyf4j3yq7pqnx_zkx9tym0000gn/T/junit5819788118170431666/state_manager/LOCK: No locks available" }
  at [unknown source]

2) [Guice/MissingConstructor]: No injectable constructor for type NodeRustEnvironment.

class NodeRustEnvironment does not have a @Inject annotated constructor or a no-arg constructor.

Requested by:
1  : NodeRustEnvironment.class(NodeRustEnvironment.java:75)
     at P2PTestNetworkRunner$1.configure(P2PTestNetworkRunner.java:198)
```
I'm going to add a comment and leave this part as is.

core/src/test/java/com/radixdlt/p2p/addressbook/AddressBookMigrationTest.java
siy: Reworked

core/src/test/java/com/radixdlt/p2p/addressbook/AddressBookMigrationTest.java
siy: Fixed

core/src/test/java/com/radixdlt/consensus/safety/SafetyStoreMigrationTest.java
siy: Reworked

core/src/test/java/com/radixdlt/consensus/safety/SafetyStoreMigrationTest.java
siy: Reworked.

core-rust/state-manager/src/store/rocks_db.rs
siy: Reformatted files with Cargo fmt.

core-rust/p2p/src/column_families.rs
dhedey: Well we can’t update these models in future unless they’re versioned.

So we’ll need to do a second migration unless we get this right now.

core-rust/state-manager/src/jni/p2p/high_priority_peers.rs
siy: Fixed

core-rust/state-manager/src/jni/p2p/high_priority_peers.rs
siy: Fixed

core-rust/p2p/src/column_families.rs
siy: Reworked to use versioned column families.

core-rust/node-common/src/java/utils.rs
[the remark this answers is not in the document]
siy: No longer an issue, and the corresponding helper is removed.

core/src/test/java/com/radixdlt/p2p/test/P2PTestNetworkRunner.java
```diff
@@ -188,6 +194,48 @@ private static Injector createInjector(
...
+            bind(PersistentSafetyStateStore.class).to(BerkeleySafetyStateStore.class);
+            Multibinder.newSetBinder(binder(), NodeAutoCloseable.class)
+                .addBinding()
+                .to(BerkeleySafetyStateStore.class);
+          }
+
+          @Provides
+          @Singleton
+          BerkeleySafetyStateStore safetyStateStore(
```
dhedey, line 213: Minor - do we need a `BerkeleySafetyStateStore` here?

dhedey (approved):
Thanks Sergiy :+1:

Happy for us to merge this, as any remaining comments are not related to how we persist things.

We can pick up any outstanding comments in a later PR which re-orgs the crates on the rust side.

core/src/test/java/com/radixdlt/p2p/test/P2PTestNetworkRunner.java
siy: Reworked module.

core-rust/p2p/src/safety_store_components.rs
```diff
@@ -0,0 +1,203 @@
...
+
+#[derive(Debug, Clone, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
+pub struct HighQC {
+    highest_quorum_certificate: QuorumCertificate,
+    highest_committed_quorum_certificate: QuorumCertificate,
+    highest_timeout_certificate: Option<TimeoutCertificate>,
+}
+
+#[derive(Debug, Clone, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
+pub struct LedgerHeader {
```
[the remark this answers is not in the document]
siy: Added comment

core-rust/p2p/src/components.rs
```diff
@@ -0,0 +1,36 @@
+use crate::engine_prelude::{ScryptoCategorize, ScryptoDecode, ScryptoEncode};
+
+/// The ID of the node stored in the Address book (Secp256k1 public key)
+#[derive(
+    Debug, Clone, ScryptoCategorize, ScryptoEncode, ScryptoDecode, PartialOrd, Ord, PartialEq, Eq,
+)]
+#[sbor(transparent)]
+pub struct RawPublicKey(pub [u8; RawPublicKey::LENGTH]);
```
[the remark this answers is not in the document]
siy: Renamed to `NodeSecp256k1PublicKey`. This clearly references the source where this data comes from and should later help with merging/wrapping existing type/etc.

core-rust/p2p/src/components.rs
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    pub fn as_bytes(&self) -> &[u8; Self::LENGTH] {
+        &self.0
+    }
+}
+
+/// The Secp256K1 signature
+#[derive(Debug, Clone, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
+pub struct Signature(pub [u8; Signature::LENGTH]);
```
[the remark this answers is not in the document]
siy: Renamed to `NodeSignature` for the same reasons ad for `NodeSecp256k1PublicKey`.

core-rust/state-manager/src/jni/node_rust_environment.rs
```diff
@@ -163,18 +164,48 @@ impl JNINodeRustEnvironment {
...
+            },
+            ..config
+        };
+
+        (base_path, config)
+    }
+
+    fn combine(base: &String, ext: &str) -> PathBuf {
+        let mut path = base.clone();
+        path.push(MAIN_SEPARATOR);
```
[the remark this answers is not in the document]
siy: Reworked. Realized that we can build `PathBuf` from the collection of strings.

PR #988, opened 2024-08-06
Abort early when deferred costs and loan are covered

iamyulong:
To include https://github.com/radixdlt/radixdlt-scrypto/pull/1867

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,7 +51,7 @@ ENV VERSION_LAST_TAG=$VERSION_LAST_TAG
 RUN apt-get update \
   && apt-get install -y --no-install-recommends \
     docker.io=20.10.24+dfsg1-1+b3 \
-    libssl-dev=3.0.11-1~deb12u2 \
+    libssl-dev=3.0.13-1~deb12u1 \
     pkg-config=1.8.1-1 \
     unzip=6.0-28 \
     wget=${WGET_VERSION} \
@@ -104,7 +104,7 @@ COPY --from=java-build-stage /radixdlt/core/build/distributions /
 # LAYER: library-build-stage-base
 # Creates the base image for building the rust library
 # =================================================================================================
-FROM debian:12.1-slim as library-build-stage-base
+FROM debian:12.1-slim AS library-build-stage-base
 WORKDIR /app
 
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -125,15 +125,14 @@ WORKDIR /app
 RUN apt-get update \
   && apt-get -y --no-install-recommends install \
     ca-certificates \
-    build-essential=12.9 \
```
dhedey, line 128: Can you fix the big comment above? Possibly we can keep the bulk of the comment as I still think it's quite useful, but we can say we are not fixing the versions at present, due to issues with the debian package manage taking down outdated versions.

Dockerfile
```diff
@@ -259,16 +258,11 @@ ARG OPENJDK_17_VERSION="17.0.11+9-1~deb12u1"
...
-    # https://security-tracker.debian.org/tracker/CVE-2023-38545
-    curl=7.88.1-10+deb12u5 \
-    gettext-base=0.21-12 \
-    daemontools=1:0.76-8.1 \
-    # https://security-tracker.debian.org/tracker/CVE-2023-4911
-    # Fixes CVE-2023-4911 can be removed when we update the base OS image to include this fix
-    # docker run -it debian:12.1-slim ldd --version
-    # This fix can be removed as long as the version printed in the above command is 2.36-9+deb12u3 or above
-    libc6=2.36-9+deb12u4 \
+    openjdk-17-jre-headless \
```
dhedey, line 261: Ditto on fixing the comment above.

Dockerfile
```diff
@@ -50,14 +50,14 @@ ENV VERSION_LAST_TAG=$VERSION_LAST_TAG
...
-    software-properties-common=0.99.30-4 \
+    docker.io \
+    libssl-dev \
+    pkg-config \
+    unzip \
+    wget \
+    software-properties-common \
   && apt-get install -y --no-install-recommends \
-    openjdk-17-jdk=${OPENJDK_17_VERSION} \
+    openjdk-17-jdk \
```
dhedey, line 60: Where is the `${OPENJDK_17_VERSION}` format specified? and the `${WGET_VERSION}`? We should probably remove those too?

In fact, I'd like to look into the reason these were parametrized? (maybe a git blame or a slack search) to check we're not creating problems here?

dhedey (commented):
I'm not sure if this is supposed to be merged, or just a prototype branch; but I think we should do the "remove fixing of debian package versions" in a separate PR.

PR #991, opened 2024-08-31
Add toolkit receipt to Core API

0xOmarA:
## Summary

Exposed the toolkit transaction receipt added in https://github.com/radixdlt/radixdlt-scrypto/pull/1893 in the Core API.

## Testing

Two simple tests were added with this PR:

* A test that the toolkit receipt is provided when the opt-in flag is set to `true`.
* A test that the toolkit receipt is not provided by default.


core-rust/Cargo.toml
```diff
@@ -23,17 +23,18 @@ resolver = "2"
...
-radix-common = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "bottlenose-af22013e", features = ["serde"] }
-radix-engine-interface = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "bottlenose-af22013e" }
-radix-engine = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "bottlenose-af22013e" }
-radix-substate-store-impls = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "bottlenose-af22013e" }
-radix-substate-store-interface = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "bottlenose-af22013e" }
-radix-substate-store-queries = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "bottlenose-af22013e" }
-radix-rust = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "bottlenose-af22013e", features = ["serde"] }
-radix-blueprint-schema-init = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "bottlenose-af22013e", features = ["serde"] }
+sbor = { git = "https://github.com/radixdlt/radixdlt-scrypto", rev = "198e691fe5ed6037c1cff9e356ca1c5b2297bde3", features = ["serde"] }
```
0xOmarA, line 26: Will need to change this to a tag once https://github.com/radixdlt/radixdlt-scrypto/pull/1893 is merged.

Dockerfile
```diff
@@ -50,7 +50,7 @@ ENV VERSION_LAST_TAG=$VERSION_LAST_TAG
 RUN apt-get update \
   && apt-get install -y --no-install-recommends \
     docker.io=20.10.24+dfsg1-1+b3 \
-    libssl-dev=3.0.13-1~deb12u1 \
+    libssl-dev=3.0.14-1~deb12u1 \
```
0xOmarA, line 53: All changes in this file were required to get the tests to pass.

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -7475,6 +7478,15 @@ components:
                 because the blueprint code relies on the total supply of the pool units to calculate their
                 redemption worth, and you've just inflated the total supply through the mint operation.
               type: boolean
+    TransactionPreviewRequestOptIns:
+      type: object
+      properties:
+        request_radix_engine_toolkit_receipt:
```
dhedey, line 7484: For consistency, we probably just want `radix_engine_toolkit_receipt`

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -7441,6 +7441,9 @@ components:
         message:
           description: An optional transaction message. Only affects the costing.
           $ref: "#/components/schemas/TransactionMessage"
+        opt_ins:
```
dhedey, line 7444: Sorry for confusion, on Core API, it looks like we call these "Options" see e.g.:

```
sbor_format_options:
          $ref: "#/components/schemas/SborFormatOptions"
```

So we should probably call this:
```
response_options:
    $ref: "#/components/schemas/TransactionPreviewResponseOptions"
```

core-rust/core-api-server/src/core_api/handlers/transaction_preview.rs
```diff
@@ -93,13 +99,33 @@ fn extract_preview_request(
...
-    // This is interpreted by the toolkit in the wallet
+    // This is interpreted by the toolkit in the wallet. This will be removed with the release of
+    // the cuttlefish protocol update.
     let encoded_receipt = to_hex(scrypto_encode(&versioned_engine_receipt).unwrap());
 
+    // Produce a toolkit transaction receipt for the transaction preview if it was requested in the
+    // request opt-ins.
+    let toolkit_receipt = if should_include_toolkit_receipt {
+        Some(
+            RuntimeToolkitTransactionReceipt::try_from(versioned_engine_receipt.clone())
```
dhedey, line 115: Very minor (and I don't expect we need to change this) - given that `RuntimeToolkitTransactionReceipt` is a temporary model, it's slightly annoying that its creation  requires an owned `versioned_engine_receipt` rather than a reference. (In fact, it could even just store references into the receipt).

dhedey (approved):
Looks good.

Slight preference to change the property name and type for consistency with the API if we can? Otherwise, looks good. 

Once this is merged, let me know and Pawel or I can pick up the Gateway integration.

core-rust/core-api-server/core-api-schema.yaml
0xOmarA: Sounds good! Making this change

core-rust/core-api-server/core-api-schema.yaml
0xOmarA: Ah, options sounds good to me! I will change the name!

core-rust/core-api-server/src/core_api/handlers/transaction_preview.rs
0xOmarA: Very true, I think that there's a world where we move towards that direction of making some of the data in the runtime toolkit receipt be references. I think that having everything be owned by the runtime receipt was easier, but I definitely agree that it is unnecessary.

PR #998, opened 2024-09-08
Build images with artifacts outside of docker

ghost:


docker/dockerfiles/app.dockerfile
```diff
@@ -0,0 +1,101 @@
...
+    RADIXDLT_CORE_API_PORT=3333 \
+    RADIXDLT_CORE_API_BIND_ADDRESS=0.0.0.0 \
+    RADIXDLT_SYSTEM_API_PORT=3334 \
+    RADIXDLT_SYSTEM_API_BIND_ADDRESS=0.0.0.0 \
+    RADIXDLT_PROMETHEUS_API_PORT=3335 \
+    RADIXDLT_PROMETHEUS_API_BIND_ADDRESS=0.0.0.0 \
+    RADIXDLT_ENGINE_STATE_API_PORT=3336 \
+    RADIXDLT_ENGINE_STATE_API_BIND_ADDRESS=0.0.0.0 \
+    RADIXDLT_NETWORK_ID=240 \
+    RADIXDLT_NODE_KEY_CREATE_IF_MISSING=false
```
dhedey, line 86: Everything up to here could be an app base image.

docker/dockerfiles/app.dockerfile
```diff
@@ -0,0 +1,101 @@
...
+
+# Copy in the application artifacts
+# The artifacts directory on the host must be populated with the required files.
+COPY artifacts/*.jar /opt/radixdlt/lib/
```
dhedey, line 90: This doesn't allow building locally (as in `docker build` and `./run-docker.sh` no longer work locally).

And it goes against the "build on same arch as run" mentality.

We historically had issues with arch and libc incompatibility doing it this way.

PR #1004, opened 2024-09-23
Use a rust layer with a full app build

ghost:


docker/dockerfiles/rust.dockerfile
```diff
@@ -37,10 +37,11 @@ RUN apt-get update \
 # We fix the version of Rust here to ensure that we can update it without having
 # issues with the caching layers containing outdated versions which aren't compatible.
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs -o rustup.sh \
-  && sh rustup.sh -y --target 1.71.1-aarch64-unknown-linux-gnu 1.71.1-x86_64-unknown-linux-gnu --default-toolchain 1.77.2
+  && sh rustup.sh -y --default-toolchain 1.77.2
 
-RUN "$HOME/.cargo/bin/cargo" install sccache --version 0.7.4
+ENV PATH="$PATH:/root/.cargo/bin"
 
-ENV CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc
```
dhedey, line 44: I imagine we still need the linkers for e.g. the ARM builds?

docker/dockerfiles/app.dockerfile
dhedey, line 1: Perhaps we can call this folder `base-images`, e.g. `docker/base-images/app.dockerfile`

docker/dockerfiles/rust.dockerfile
```diff
@@ -0,0 +1,47 @@
...
+# We fix the version of Rust here to ensure that we can update it without having
+# issues with the caching layers containing outdated versions which aren't compatible.
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs -o rustup.sh \
+  && sh rustup.sh -y --default-toolchain 1.81.0
+
+ENV PATH="$PATH:/root/.cargo/bin"
+
+COPY . .
+
+RUN cd core-rust && \
```
dhedey, line 46: Can we add a comment to explain why we're doing a build here?

```
# We do a pre-emptive build so that the builder layer contains pre-built dependencies,
# to speed up future compilation (assuming most dependencies are the same in the
# builder image and the built code)
```

docker/dockerfiles/rust.dockerfile
```diff
@@ -0,0 +1,47 @@
+# =================================================================================================
+# LAYER: library-build-stage-base
+# Creates the base image for building the rust library
+# =================================================================================================
+FROM debian:12.1-slim AS library-build-stage-base
```
dhedey, line 5: Should we call this e.g. `rust-builder` ? In here and line 2

docker/dockerfiles/java.dockerfile
```diff
@@ -0,0 +1,69 @@
...
+# For the standard build:
+# - TARGETPLATFORM - provided automatically, specified via `docker build --platform xyz`
+# - RUST_PROFILE - optional - either `debug` or `release` - defaults to release
+#
+# There are no args for the local rust builder.
+# =================================================================================================
+# LAYER: java-build-stage
+# The base for building the Java application
+# =================================================================================================
+FROM debian:12.1-slim AS java-build-stage
```
dhedey, line 13: For we call this e.g. `java-builder` ? in here and line 10

docker/dockerfiles/rust.dockerfile
```diff
@@ -0,0 +1,47 @@
...
+  && rm -rf /var/lib/apt/lists/*
+
+# We fix the version of Rust here to ensure that we can update it without having
+# issues with the caching layers containing outdated versions which aren't compatible.
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs -o rustup.sh \
+  && sh rustup.sh -y --default-toolchain 1.81.0
+
+ENV PATH="$PATH:/root/.cargo/bin"
+
+COPY . .
```
dhedey, line 44: Do we know why we need the whole of the repository, rather than just the `core-rust` folder, as in the `Dockerfile.v2`?

If we do need the whole repository here, it might be worth explaining why :)

Dockerfile.v2
```diff
@@ -0,0 +1,50 @@
...
+    cp core /artifacts
+
+FROM radixdlt/babylon-node-build-layers:v1.3.0-rust AS rust-container
+
+WORKDIR /app
+
+COPY core-rust/ /app
+
+RUN mkdir -p /artifacts && \
+    /root/.cargo/bin/cargo build --profile=release && \
```
dhedey, line 40: Given we set `ENV PATH="$PATH:/root/.cargo/bin"` in the rust dockerfile, perhaps we can just do `cargo build` here, to better align across the two builds?

Dockerfile.v2
```diff
@@ -0,0 +1,50 @@
...
+  /radixdlt/
+COPY ./gradle /radixdlt/gradle
+COPY ./common /radixdlt/common
+COPY ./core /radixdlt/core
+COPY ./core-rust-bridge /radixdlt/core-rust-bridge
+COPY ./cli-tools /radixdlt/cli-tools
+COPY ./shell /radixdlt/shell
+COPY ./keygen /radixdlt/keygen
+# Need .git for tag versions - but this can probably be removed soon
+COPY ./.git/* /radixdlt/.git/
```
dhedey, line 23: I'm fairly certain this can now be removed after Les's changes ~8 months ago :+1:

.github/workflows/docker-v2.yml
```diff
@@ -0,0 +1,228 @@
...
+      outputs:
+        tag: ${{ steps.setup_tags.outputs.tag }}
+      steps:
+        - uses: RDXWorks-actions/checkout@main
+          with:
+            fetch-depth: 0
+        - id: setup_tags
+          run: echo "tag=$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
+
+  build_deb:
```
dhedey, line 49: I appreciate it's not related to this change, but would it be easy to move this out to a separate workflow, e.g. `deb-package.yml`?

It's not related to docker, and confusing it's in here... You'll need to copy `setup_version_properties` across to it too.

.github/workflows/publish-build-layer-images.yml
```diff
@@ -0,0 +1,206 @@
...
+    uses: radixdlt/public-iac-resuable-artifacts/.github/workflows/docker-build.yml@main
+    with:
+      runs_on: ubuntu-16-cores-selfhosted
+      environment: "release"
+      image_registry: "docker.io"
+      image_organization: "radixdlt"
+      image_name: "babylon-node-build-layers"
+      tag: ${{ inputs.docker_tag }}-rust
+      context: "."
+      dockerfile: "Dockerfile"
```
dhedey, line 26: Hmm ... shouldn't this be `docker/dockerfiles/rust.dockerfile` etc? (or after rename, `docker/base-images/app.dockerfile`?)

docker/dockerfiles/java.dockerfile
dhedey: And possibly `java-builder.dockerfile` ?

.github/workflows/docker-layers.yml
```diff
@@ -0,0 +1,323 @@
+name: Docker layers
```
dhedey, line 1: I'm not sure I follow what this is supposed to do? I think `publish-build-layer-images.yml` is supposed to do this (and looks cleaner and is better named)? So perhaps this needs deleting?

.github/workflows/publish-build-layer-images.yml
```diff
@@ -0,0 +1,206 @@
...
+  workflow_dispatch:
+    inputs:
+      docker_tag:
+        description: "Docker tag to be published"
+
+permissions:
+  packages: write
+  pull-requests: write
+  id-token: write
+  contents: read
```
dhedey, line 13: Do we need all these permissions? Naively it looks like we should just need contents read and the ability to access the `secrets.DOCKERHUB_RELEASER_ROLE` ?

.github/workflows/docker-v2-release.yml
```diff
@@ -0,0 +1,277 @@
+name: Docker v2 release
+
+on:
+  pull_request: # TODO: Remove after testing the PR
+  release:
+    types: [published]
+  push:
+    branches:
+      - release\/*
```
dhedey, line 9: This is a change compared with `docker.yml` and isn't right.

We want to release to dockerhub only when this happens:
```
  release:
    types: [published]
```

We want to test the build (and possibly publish internally) when this happens:
```
  push:
    branches:
      - develop
      - main
      - release\/*
```

.github/workflows/docker-v2-release.yml
```diff
@@ -0,0 +1,277 @@
+name: Docker v2 release
```
dhedey, line 1: There seems to be a lot of duplication between this and `docker-v2.yml` - possibly when we finish other clear-up, it might be worth considering if we can merge them? (possibly using `    if: github.event_name == 'release'` to replace internal publishes with publishes to Dockerhub?)

.github/workflows/docker-v2.yml
```diff
@@ -0,0 +1,228 @@
+name: Docker v2
```
dhedey, line 1: (comments on `docker-v2-release.yml` also apply here)

.github/workflows/docker-v2-release.yml
```diff
@@ -0,0 +1,277 @@
...
+      outputs:
+        tag: ${{ steps.setup_tags.outputs.tag }}
+      steps:
+        - uses: RDXWorks-actions/checkout@main
+          with:
+            fetch-depth: 0
+        - id: setup_tags
+          run: echo "tag=$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
+
+  build_deb:
```
dhedey, line 50: I appreciate it's not related to this change, but would it be easy to move this out to a separate workflow, e.g. `deb-package.yml`?

It's not related to docker, and confusing it's in here... You'll need to copy setup_version_properties across to it too.

.github/workflows/docker-v2-release.yml
```diff
@@ -0,0 +1,277 @@
...
+        run: |
+          sudo apt-get update && sudo apt-get install -y make
+          cd core && make build-core
+      - name: Upload generated debian package
+        uses: RDXWorks-actions/upload-artifact@main
+        with:
+          name: deb4docker
+          path: "${{ github.workspace }}/docker/*.deb"
+
+  build_docker:
```
dhedey, line 108: Can we clear this up by using the build docker action like `publish-build-layer-images`?

Dockerfile.v2
```diff
@@ -0,0 +1,50 @@
...
+
+RUN SKIP_NATIVE_RUST_BUILD=TRUE ./gradlew clean build -x test -Pci=true -PrustBinaryBuildType=release
```
dhedey, line 25: Compared with the `Dockerfile`, I think we've lost these:

```
ARG VERSION_BRANCH=""
ARG VERSION_COMMIT=""
ARG VERSION_DISPLAY=""
ARG VERSION_BUILD=""
ARG VERSION_TAG=""
ARG VERSION_LAST_TAG=""

ENV VERSION_BRANCH=$VERSION_BRANCH
ENV VERSION_COMMIT=$VERSION_COMMIT
ENV VERSION_DISPLAY=$VERSION_DISPLAY
ENV VERSION_BUILD=$VERSION_BUILD
ENV VERSION_TAG=$VERSION_TAG
ENV VERSION_LAST_TAG=$VERSION_LAST_TAG
```

Which need to be passed in as arguments from github in order to get the correct versioning information baked into the jar.

dhedey (commented):
Thanks! Nice work. There are quite a few things to fix / neaten, but this looks great and like it will be a good simplification and speed-up.

Once we're tested and happy, I can re-review and then we can change the PR to get rid of the `v2` things and replace the existing workflows etc.

Dockerfile.v2
```diff
@@ -0,0 +1,63 @@
+FROM radixdlt/babylon-node-build-layers:v1.3.0-java AS java-container
+
+WORKDIR /radixdlt
+
+ARG WGET_VERSION="1.21.3-1+b2"
```
dhedey, line 5: The `WGET_VERSION` isn't needed I think?

.github/workflows/docker.yml
```diff
@@ -78,7 +73,7 @@ jobs:
         uses: RDXWorks-actions/cache/restore@main
         with:
           path: /tmp/outputs/cache/docker
-          key: babylon-node-default-${{ hashFiles('./Dockerfile') }}
+          key: babylon-node-default-${{ hashFiles('./Dockerfile.v2') }}
```
dhedey, line 76: Now that we're merging things together, can we replace `./Dockerfile` with `./Dockerfile.v2`? :)

Dockerfile.v2
```diff
@@ -0,0 +1,63 @@
...
+
+RUN SKIP_NATIVE_RUST_BUILD=TRUE ./gradlew clean build -x test -Pci=true -PrustBinaryBuildType=release
+
+RUN cd core/build/distributions && \
+    unzip -j *.zip && \
+    mkdir -p /artifacts && \
+    cp *.jar /artifacts && \
+    cp core /artifacts
+
+FROM radixdlt/babylon-node-build-layers:v1.3.0-rust AS rust-container
```
dhedey, line 46: Minor - is it possible to move these versions together to variables at the top of this file? That way it's easier to update them all and not accidentally miss one :D

dhedey (commented):
Looks great!

A couple of small comments, but otherwise I think we're very close to  getting this merged.

I suggest the process is:
* We move/rebase this branch to `release/cuttlefish`
* We then use it to build the `1.3.0` version of the base images
* We set the docker build to use the `1.3.0` docker image.
* We can then merge the PR into `release/cuttlefish`

And finally we can merge `release/cuttlefish` into `develop` 👍

dhedey (approved):

Dockerfile
```diff
@@ -54,11 +54,39 @@ RUN mkdir -p /artifacts && \
...
+FROM scratch AS java-container
+COPY --from=java-build-stage /radixdlt/core/build/distributions /
+
+# =================================================================================================
+# LAYER: library-builder-local
+# This layer is just used for local building in development via `local-cached-rust-build.yaml`.
+# Specifically - the Rust isn't built as part of the image, instead the CMD of the image is to do the build.
+# It allows us to use volumes at runtime to cache the build dependencies and artifacts.
+# =================================================================================================
+FROM radixdlt/babylon-node-build-layers:${IMAGE_VERSION}-rust AS library-builder-local
```
dhedey, line 77: Ah dang this is for the buildRustForDocker which is now only used by the buildDeb stuff.

Ideally we'd rename that to `buildRustForDeb` or something, but doesn't need to be this PR. And actually I think it makes more sense to nuke all this and potentially just run a standard cargo build on the target Debian-compatible system. Anyway, as I say, doesn't need to be this PR :)

Dockerfile
```diff
@@ -54,11 +54,39 @@ RUN mkdir -p /artifacts && \
...
+# =================================================================================================
+# LAYER: java-container
+# Exports only the java application artifacts from the java-build-stage
+# =================================================================================================
+FROM scratch AS java-container
+COPY --from=java-build-stage /radixdlt/core/build/distributions /
+
+# =================================================================================================
+# LAYER: library-builder-local
+# This layer is just used for local building in development via `local-cached-rust-build.yaml`.
```
dhedey, line 73: This comment hasn't been true for a few years - in local development, we run a native build.

I think this layer is just used for building the Rust image for building a debian package.

dhedey (approved):

PR #1008, opened 2024-10-19
feature: Genesis executes as a protocol update

dhedey:
## Summary

Reworked the protocol update triggering, enactment and scenario execution.
* Genesis is now a protocol update, and can be automatically resumed.
* Scenarios now execute in a separate batch group
* Protocol updates can now be triggered immediately after previous updates - so e.g. we don't have to wait till epoch 5 to get cuttlefish activated for testnets and local tests.

And now, post-merge-in of https://github.com/radixdlt/babylon-node/pull/1009 we have a few minor changes:
* Removes the name of genesis from `ProtocolConfig`, because it's relied on by static logic, and I don't feel like this is something we should allow people to change (this might break some of our tests/testnets, I'm liaising with Les to discuss)
* Updates the testnet configuration to run Anemone/Bottlenose back-to-back at genesis
* Changes the testing defaults to:
    * Run all testnet-configured protocol updates on launch - this means that, by default, tests are running against the latest engine. (This was traditionally too slow to activate before, until we enabled back-to-back configuration).
    * But don't run scenarios (because this would slow down the tests too much).
* Fixes various tests broken by running all protocol updates in our tests

In next PRs:
* Other Cuttlefish implementation

## Testing

* Have tested that genesis can be resumed by stopping it mid-update
* I've tested that it can sync stokenet genesis locally


core-rust/core-api-server/src/core_api/handlers/stream_transactions.rs
```diff
@@ -189,11 +184,28 @@ pub fn to_api_ledger_proof(
...
+            batch_group_index: _,
+            batch_group_descriptor: _,
             batch_index,
-        } => models::LedgerProofOrigin::ProtocolUpdateLedgerProofOrigin {
-            protocol_version_name: protocol_version_name.to_string(),
-            batch_idx: to_api_u32_as_i64(*batch_index),
-        },
+            is_end_of_update: _,
+        } => {
+            // todo!(FIX)
```
dhedey, line 193: Will fix in the next PR/s when I fix Core API to work with Cuttlefish

core/src/main/java/com/radixdlt/rev2/REv2StateComputer.java
```diff
@@ -323,25 +323,19 @@ public LedgerProofBundle commit(LedgerExtension ledgerExtension, VertexStoreStat
...
-      if (newHeader.nextProtocolVersion().isPresent()) {
-        final var nextProtocolVersion = newHeader.nextProtocolVersion().unwrap();
-        latestProofWhichInitiatedAProtocolUpdate = Option.some(newLatestProof);
-        final var finalExecutionProof =
-            this.rustProtocolUpdate.applyProtocolUpdate(nextProtocolVersion).postUpdateProof();
-        newLatestProof = finalExecutionProof;
-        latestProtocolUpdateExecutionProof = Option.some(finalExecutionProof);
-      }
-    } while (newLatestProof != lastHandledProof);
+    final var newHeader = newLatestProof.ledgerHeader();
```
dhedey, line 327: I previously created this loop in the last PR, but decided it made more sense handling this loop in rust in one place - which is now triggered by `applyKnownPendingProtocolUpdate`.

core/src/main/java/com/radixdlt/sync/LedgerSyncDtoConversions.java
```diff
@@ -114,18 +112,14 @@ public static LedgerProof syncDtoToConsensusOriginatedLedgerProof(
             ledgerProof.getOpaque(), REv2ToConsensus.signatures(ledgerProof.getSignatures())));
   }
 
+  /**
```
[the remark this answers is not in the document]
LukasGasior1: I think this is fine. It seems to only be used in MockedSyncServiceModule, so maybe let's inline this conversion there?

core-rust/state-manager/src/protocol/protocol_updates/protocol_update_progress.rs
```diff
@@ -19,31 +20,40 @@ impl ProtocolUpdateProgress {
...
                 batch_index,
-            } => ProtocolUpdateProgress::UpdateInProgress {
-                protocol_version_name: protocol_version_name.clone(),
-                last_batch_index: *batch_index as usize,
-            },
+                is_end_of_update,
+            } => {
+                if is_end_of_update {
+                    ProtocolUpdateProgress::NotUpdating
+                } else {
```
iamyulong, line 49: More of a question: during our sync, you mentioned that there would be trigger after `end_of_update`. If a node stops right after committing the last element of the last batch, how would the triggering be recovered as we set progress to `NotUpdating`?

core-rust/state-manager/src/protocol/protocol_state.rs
```diff
@@ -18,76 +19,173 @@ impl NodeProtocolUpdateExecutor {
...
+                ProtocolUpdateProgress::UpdateInProgress {
+                    protocol_version_name,
+                    last_batch_group_index,
+                    last_batch_index,
+                } => {
+                    info!("[UPDATE: {protocol_version_name}] Resuming protocol update");
+                    self.execute_protocol_update_actions(
+                        &protocol_version_name,
+                        last_batch_group_index,
+                        last_batch_index.checked_add(1).unwrap(),
```
iamyulong, line 67: What happens if the `last_batch_index` is already the last?

core-rust/state-manager/src/protocol/protocol_state.rs
```diff
@@ -18,76 +19,173 @@ impl NodeProtocolUpdateExecutor {
...
+                0
+            };
+            let batches = {
+                // In a separate block to ensure the database lock guard is dropped promptly.
+                batch_group.generate_batches(self.database.lock().deref())
+            };
+            let total_batches = batches.len();
+            let remaining_batches = batches.into_iter().enumerate().skip(start_at_batch);
+            for (batch_index, batch_generator) in remaining_batches {
+                let batch_number = batch_index + 1;
```
iamyulong, line 128: Doesn't seem right, shouldn't the `batch_number` be `start_at_batch + batch_index`?

core-rust-bridge/src/main/java/com/radixdlt/protocol/ProtocolConfig.java
```diff
@@ -74,22 +74,19 @@
...
 public record ProtocolConfig(
-    String genesisProtocolVersion,
     ImmutableList<ProtocolUpdateTrigger> protocolUpdateTriggers,
     Map<String, byte[]> rawProtocolUpdateContentOverrides) {
 
   public static final String GENESIS_PROTOCOL_VERSION_NAME = "babylon-genesis";
+  public static final String ANEMONE_PROTOCOL_VERSION_NAME = "anemone";
+  public static final String BOTTLENOSE_PROTOCOL_VERSION_NAME = "bottlenose";
+
+  public static final String LATEST_PROTOCOL_VERSION_NAME = BOTTLENOSE_PROTOCOL_VERSION_NAME;
```
LukasGasior1, line 84: this seems easy to miss when introducing a new update :thinking: wdyt about moving it to `NetworkStatusTest`, as it's only used there?

core-rust-bridge/src/main/java/com/radixdlt/statecomputer/commit/LedgerProof.java
```diff
@@ -94,7 +95,16 @@ public static LedgerProof testingGenesis(
...
             Option.empty());
 
-    return new LedgerProof(genesisLedgerHeader, new LedgerProofOrigin.Genesis(HashUtils.zero256()));
+    return new LedgerProof(
+        genesisLedgerHeader,
+        new LedgerProofOrigin.ProtocolUpdate(
+            ProtocolConfig.GENESIS_PROTOCOL_VERSION_NAME,
+            Option.some(HashUtils.zero256()),
+            UInt64.ZERO,
+            "-",
```
LukasGasior1, line 104: super minor: let's stick to an empty string here, for consistency

core-rust/state-manager/src/protocol/genesis_data_resolver.rs
```diff
@@ -62,54 +62,112 @@
  * permissions under this License.
```
LukasGasior1, line 62: lol, I find it funny that git actually categorized this as *.java -> *.rs rename + file edit :smile:

core-rust/state-manager/src/protocol/protocol_configs/mainnet_protocol_config.rs
```diff
@@ -6,7 +6,7 @@ use ProtocolUpdateEnactmentCondition::*;
 pub fn mainnet_protocol_config() -> ProtocolConfig {
     // See config_printer.rs
     ProtocolConfig::new_with_triggers(hashmap! {
-        ANEMONE_PROTOCOL_VERSION => EnactAtStartOfEpochIfValidatorsReady {
+        ProtocolVersionName::anemone() => EnactAtStartOfEpochIfValidatorsReady {
```
LukasGasior1, line 9: should we consider updating anomone and bottlenose to `EnactAtStartOfEpochUnconditionally`? :thinking:

core-rust/state-manager/src/state_manager.rs
```diff
@@ -374,24 +385,44 @@ impl StateManager {
...
     /// is in particular state and ready for protocol update. Hence, we trust the input here and
     /// unconditionally update the internally-maintained protocol version to its new value.
-    pub fn apply_protocol_update(
+    pub fn apply_known_pending_protocol_updates(&self) -> ProtocolUpdateResult {
+        let resultant_version = self
+            .protocol_update_executor
+            .resume_protocol_update_if_any();
+
+        let Some(resultant_version) = resultant_version else {
+            panic!("apply_protocol_update_and_any_following is only expected to be called if a pending protocol update is known");
```
LukasGasior1, line 406: `apply_protocol_update_and_any_following` -> `apply_known_pending_protocol_updates`

core-rust/state-manager/src/protocol/protocol_state.rs
LukasGasior1: looks right to me :thinking: the list is enumerated before `skip`

core/src/main/java/com/radixdlt/ledger/LedgerProofBundle.java
```diff
@@ -152,21 +145,19 @@ public long resultantStateVersion() {
...
       case LedgerProofOrigin.Consensus consensus -> primaryProof;
-      case LedgerProofOrigin.Genesis genesis -> primaryProof;
-      case LedgerProofOrigin.ProtocolUpdate protocolUpdate ->
-      // If our latest proof is of protocol update origin,
-      // then we're returning the closest proof that initializes a protocol update.
-      // Since protocol updates can't themselves trigger another protocol update,
-      // this is guaranteed to be a consensus (or, possibly in some testing corner case, genesis)
-      // proof.
-      latestProofWhichInitiatedAProtocolUpdate().unwrap();
+      case LedgerProofOrigin.ProtocolUpdate ignored -> {
```
LukasGasior1, line 154: could be simplified to:
```
      case LedgerProofOrigin.ProtocolUpdate ignored ->
        latestProofWhichInitiatedOneOrMoreProtocolUpdates()
          .or(latestProofWhichInitiatedAnEpochChange());
```

LukasGasior1 (approved):

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -7925,15 +7964,50 @@ components:
...
+            atomically.
+
+            NOTE: Some of these values may be placeholder values for protocol updates pre-Cuttlefish
+            on nodes which haven't resynced since Cuttlefish. In particular, the following values might be
+            invalid on such nodesL
+
+            * `config_hash` (placeholder of all zeros)
+            * `batch_group_idx` (placeholder of 0)
+            * `batch_group_name` (placeholder of "")
+            * `batch_idx` (whatever the non-grouped index was)
```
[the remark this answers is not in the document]
dhedey: I wanted to make the migration conceptually reversible for purity; but tbh it's not needed and unhelpful.

I agree with you, I'll change to 0 here as I think it's more consistent :)

core-rust/state-manager/src/protocol/protocol_updates/protocol_update_progress.rs
dhedey: That's correct.

And the answer is that such proofs will have both `ledger_header.next_protocol_version` and `is_end_of_update` set, and about 10 lines above, we have this code:
```rust
        if let Some(latest_proof_protocol_version) =
            &latest_proof.ledger_header.next_protocol_version
        {
            return ProtocolUpdateProgress::UpdateInitiatedButNothingCommitted {
                protocol_version_name: latest_proof_protocol_version.clone(),
            };
        }
```

The `next_protocol_version` is set in `SystemCommitRequestFactory::create_proof` as a result of:
```
        let end_state = series_executor.finalize_series(batch_details.to_batch_situation());
        SystemPrepareResult::from_committed_series(committed_transactions, end_state)
```
And the `series_executor.finalize_series` ultimately gets it from:
```
        let next_protocol_version = self
            .protocol_state
            .check_for_update_trigger_at_end_of_batch(
                self.epoch_change.as_ref(),
                self.state_version,
                batch_situation,
            );
```

core-rust/state-manager/src/protocol/protocol_state.rs
dhedey: Good question! :)

So `last_batch_index` is not the very last batch in the update, because otherwise we will be matching on `ProtocolUpdateProgress::NotUpdating` or `ProtocolUpdateProgress::UpdateInitiatedButNothingCommitted` (in the immediate trigger case).

So the only edge case is that `last_batch_index` is the last in its batch group, but _not_ the final batch group.

In which case, this iterator is empty when we execute the protocol update `let remaining_batches = batches.into_iter().enumerate().skip(start_at_batch);`, and we immediate go onto the next batch group. The logging is maybe a tiny bit misleading, but the execution logic is sound.

I'll add a comment here to explain this 👍

core-rust-bridge/src/main/java/com/radixdlt/protocol/ProtocolConfig.java
dhedey: Well my thinking was that this might well be useful in other tests :). Indeed, we use a similar pattern across the Scrypto code base.

I view the `NetworkStatusTest` as ensuring this constant stays up to date. Will add some comments...

core-rust/state-manager/src/protocol/protocol_configs/mainnet_protocol_config.rs
dhedey: Could do, but what benefit do you see? Slight win in performance?

PR #1011, opened 2024-10-21
feature: Enable cuttlefish

dhedey:
## Summary
* Enables cuttlefish on testnets
* The Core API won't _panic_, but some of its substate models are still incomplete.

The following will come in a follow-up PR:
* Fix `BootLoaderModuleFieldTransactionValidationConfiguration` substate
* Add some more Core API tests
* Add whether subintents were nullified to transaction receipts
* Add V2 Transaction Preview
* Add Subintent status

## Testing

Adds tests:
* `CuttlefishProtocolUpdateTest` in Java:
  * `transaction_v2_behaviour_across_cuttlefish`
  * `protocol_update_process_updates_status_summary`
* `TransactionSubmitTest` in Java:
  * `test_core_api_can_submit_and_commit_transaction_v2_after_cuttlefish`


core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -8110,6 +8187,7 @@ components:
           # NOTE: These need to match LedgerTransactionType
           Genesis: '#/components/schemas/GenesisLedgerTransaction'
           User: '#/components/schemas/UserLedgerTransaction'
+          UserV2: '#/components/schemas/UserLedgerTransactionV2'
```
PawelPawelec-RDX, line 8190: UserV2 is not added to LedgerTransactionType, it fails to generate proper models.

dhedey: Thanks for the heads up, this is fixed in the latest push 👍

dhedey: Fixed in latest push, thanks for spotting this

core-rust-bridge/src/main/java/com/radixdlt/protocol/ProtocolConfig.java
```diff
@@ -105,6 +119,43 @@ public static ProtocolConfig onlyGenesis() {
     return new ProtocolConfig(ImmutableList.of());
   }
 
+  public static ProtocolConfig launchAt(String upToProtocolVersionName) {
+    String newVersion = null;
+    List<ProtocolUpdateTrigger> protocolUpdateTriggers = new ArrayList<>();
+    for (String version : VERSION_NAMES) {
+      var lastVersion = newVersion;
+      newVersion = version;
+      if (lastVersion != null) {
```
iamyulong, line 128: Minor: for readability, it would be better to just update the newVersion after the first if clause, instead of introudcing a new varible.

core-rust-bridge/src/main/java/com/radixdlt/protocol/ProtocolConfig.java
```diff
@@ -105,6 +119,43 @@ public static ProtocolConfig onlyGenesis() {
...
+        if (newVersion.equals(upToProtocolVersionName)) {
+          protocolUpdateTriggers.add(
+              new ProtocolUpdateTrigger(newVersion, unconditionallyAtEpoch(epoch)));
+          break;
+        } else {
+          protocolUpdateTriggers.add(
+              new ProtocolUpdateTrigger(newVersion, immediatelyAfter(lastVersion)));
+        }
+      }
+    }
```
iamyulong, line 155: Slightly confused by the logic, shouldn't we have
```
Genesis
P1 -> unconditional at (epoch)
P2 -> immediately after (P1)
P3 -> immediately after (P2)
```

core-rust/core-api-server/src/core_api/handlers/stream_transactions.rs
```diff
@@ -772,3 +841,228 @@ fn to_api_direct_substate_id(
...
+    } = transaction_intent;
+    Ok(models::TransactionIntentV2 {
+        hash: to_api_transaction_intent_hash(transaction_intent_hash),
+        hash_bech32m: to_api_hash_bech32m(context, transaction_intent_hash)?,
+        transaction_header: Box::new(to_api_transaction_header_v2(transaction_header)?),
+        root_intent_core: Box::new(to_api_intent_core_v2(context, root_intent_core)?),
+        non_root_subintents: non_root_subintents
+            .0
+            .iter()
+            // TODO:CUTTLEFISH - fix the subintent hashes
```
iamyulong, line 942: Just a thought: slighlty annoyed that we need to caclualte the intent hash on-the-fly. Maybe we can store that.

core-rust/state-manager/src/jni/transaction_preparer.rs
```diff
@@ -116,6 +116,102 @@ extern "system" fn Java_com_radixdlt_transaction_TransactionPreparer_prepareInte
...
+
+#[derive(Debug, Clone, PartialEq, Eq, ScryptoSbor)]
+struct PrepareTransactionIntentV2Response {
+    raw_transaction_intent: RawTransactionIntent,
+    transaction_intent_hash: TransactionIntentHash,
+    subintent_hashes: Vec<SubintentHash>,
+}
+
+#[no_mangle]
+extern "system" fn Java_com_radixdlt_transaction_TransactionPreparer_prepareTransactionIntentV2(
```
iamyulong, line 135: For future consideration, we may want to move all the test-related JNI interfaces to a dedicated folder `jni-test`.

core-rust/state-manager/src/protocol/protocol_configs/mainnet_protocol_config.rs
```diff
@@ -4,7 +4,7 @@ use crate::protocol::*;
 use ProtocolUpdateEnactmentCondition::*;
 
 pub fn mainnet_protocol_config() -> ProtocolConfig {
-    // See config_printer.rs
+    // See config_printer.rs > print_calculated_protocol_config_code()
     ProtocolConfig::new_with_triggers(hashmap! {
```
iamyulong, line 8: We're missing cuttlefish config for mainnet?

dhedey: Yes - we will need to add it before we publish the node when we know when we want it to actually be :)

core-rust/state-manager/src/jni/transaction_preparer.rs
dhedey: Yeah I think it's a nice idea. Will see if I can find time to do it now, if not we can do it in a future PR 👍

core-rust/core-api-server/src/core_api/handlers/stream_transactions.rs
dhedey: Yes that's my plan! And hence why I didn't implement a hacky recalculation here.

core-rust-bridge/src/main/java/com/radixdlt/protocol/ProtocolConfig.java
dhedey: This takes a protocol version. It's used to have, say, Anemone / Bottlenose enact at genesis, then do some tests, then enact Cuttlefish a bit later. I've added a comment to explain :)

core-rust/core-api-server/src/core_api/handlers/stream_transactions.rs
dhedey: (fixed now)

iamyulong (approved):

PR #1012, opened 2024-10-31
Use the public IP address that is reported the most

iamyulong:
## Summary

Use the public IP address that is reported the most by the sources, to work around error:

```
An IP address of this node hasn't been configured. Make sure you set your `network.host_ip` property. An attempt was made to acquire it from an external oracle, but that also failed (services queried: [https://ifconfig.me/, https://ipv4.icanhazip.com/, https://www.trackip.net/ip, https://myexternalip.com/raw, https://ipecho.net/plain, https://ifconfig.co/ip, https://checkip.amazonaws.com/]).
```

### DIFF
--- a/core/src/main/java/com/radixdlt/p2p/hostip/NetworkQueryHostIp.java
+++ b/core/src/main/java/com/radixdlt/p2p/hostip/NetworkQueryHostIp.java
@@ -84,7 +84,8 @@ import java.util.Optional;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
-import okhttp3.*;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -171,14 +172,21 @@ final class NetworkQueryHostIp {
       }
     }
     if (successCounts.isEmpty()) {
-      log.warn("No suitable address found");
+      log.warn(
+          "Attempts to resolve this node's public IP address with external resolution services"
+              + " failed at every attempt. Perhaps this node cannot connect to the internet?");
     }
     if (successCounts.size() > 1) {
       String votes =
           successCounts.keySet().stream()
               .map(key -> key + "=" + successCounts.get(key))
               .collect(Collectors.joining(", ", "{", "}"));
-      log.warn("More than 1 IP address found: " + votes);
+      log.warn(
+          String.format(
+              "Attempts to resolve this node's public IP address with external resolution services"
+                  + " resulted in more than one distinct IP address. The node will continue by"
+                  + " using the most common: %s. The full list of resolved addresses is: %s",
+              maxResult.get(), votes));
     }
     return new VotedResult(maxResult, queryResults.build());
   }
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/radixdlt/p2p/hostip/NetworkQueryHostIp.java
```diff
@@ -148,33 +147,29 @@ public VotedResult queryNetworkHosts() {
...
-          log.debug("Found address {}", success.get());
-          return new VotedResult(success, queryResults.build());
+        if (newCount > maxCount) {
+          maxCount = newCount;
+          maxResult = result.toOptional();
         }
       }
     }
-    log.debug("No suitable address found");
-    return new VotedResult(Optional.empty(), queryResults.build());
```
dhedey, line 177: In the case where we have one vote for two different IPs (and the rest of the calls fail), we choose one at random now...

That might be the right move to be defensive and allow the node to boot up... But I think I'd like to handle this a little more clearly. 

In particular:
* If `successCounts.size() == 0` it would be nice to `log.warn` that we couldn't resolve any IP address from external providers - maybe the network is down?
* And if we have `successCounts > 1` then we log with `log.warn`, outputting all the ips and vote counts and that we have arbitrarily resolved `X`?

dhedey (approved):

core/src/main/java/com/radixdlt/p2p/hostip/NetworkQueryHostIp.java
```diff
@@ -148,33 +148,39 @@ public VotedResult queryNetworkHosts() {
...
+    if (successCounts.isEmpty()) {
+      log.warn("No suitable address found");
+    }
+    if (successCounts.size() > 1) {
+      String votes =
+          successCounts.keySet().stream()
+              .map(key -> key + "=" + successCounts.get(key))
+              .collect(Collectors.joining(", ", "{", "}"));
+      log.warn("More than 1 IP address found: " + votes);
```
dhedey, line 181: MINOR: This is not super clear to me as a node runner whether it's something to be worried about - could we instead say something like:
```java
log.warn(String.format(
    "Attempts to resolve this node's public IP address with external resolution services resulted in more than one distinct IP address. The node will continue by using the most common: %s. The full list of resolved addresses is: %s",
    maxResult.get(),
    votes,
));
```

core/src/main/java/com/radixdlt/p2p/hostip/NetworkQueryHostIp.java
```diff
@@ -148,33 +148,39 @@ public VotedResult queryNetworkHosts() {
...
+        if (newCount > maxCount) {
+          maxCount = newCount;
+          maxResult = result.toOptional();
         }
       }
     }
-    log.debug("No suitable address found");
-    return new VotedResult(Optional.empty(), queryResults.build());
+    if (successCounts.isEmpty()) {
+      log.warn("No suitable address found");
```
dhedey, line 174: MINOR: This is not super clear to me as a node runner, perhaps something like this?
```java
log.warn("Attempts to resolve this node's public IP address with external resolution services failed at every attempt. Perhaps this node cannot connect to the internet?");
```

PR #1015, opened 2024-11-06
test: Add test for duplicate submission of subintent

dhedey:
## Summary

Adds a test for duplicate submission of subintent, with an assertion commented out with a TODO:CUTTLEFISH which I'll fix in the next PR.

Also includes some other test-related refactors to commonise/unify test helpers.

PR #1016, opened 2024-11-07
Feature/immediately handle subintent nullification

dhedey:
## Summary

The main intention was to ensure subintent handling is fixed/immediate in the pending result cache and the mempool.

As part of this ticket, I worked through a few refactors:
* Encapsulated mempool related systems directly under the `MempoolManager` where it can (A) safely handle atomicity of operations and avoid deadlocks, and (B) enable us to eventually safely refactor and possibly merge the cache and mempool. That was the intended direction ~18 months ago, I'm not quite sure why it ended up not being finished - but I've pushed through with the merger anyway. I think it works nicely.
* Improved the scope of our locks to be smaller. In one particular case in `add_if_committable_internal` I expect this to have a non-trivial impact on throughput.
* Improved the pending rejection "is permanent" checks to require passing an `AtState` so we can't use the wrong thing by mistake; and commonise the code.

## Testing

* Existing tests pass
* Fixed the test added in https://github.com/radixdlt/babylon-node/pull/1015

Might add some more unit tests tomorrow.


core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
```diff
@@ -14,46 +14,78 @@ pub type ExecutionRejectionReason = RejectionReason;
...
 }
 
 impl MempoolRejectionReason {
-    pub fn is_permanent_for_payload(&self) -> bool {
+    pub fn is_from_execution(&self) -> bool {
+        match self {
+            MempoolRejectionReason::TransactionIntentAlreadyCommitted(_) => false,
+            MempoolRejectionReason::SubintentAlreadyFinalized(_) => false,
+            MempoolRejectionReason::FromExecution(_) => true,
+            MempoolRejectionReason::ValidationError(_) => false,
```
iamyulong, line 48: Per cuttlefish update, transaction validation is stateful. Is the permanent rejection logic below still correct?

core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
```diff
@@ -513,134 +549,202 @@ const NON_REJECTION_RECALCULATION_DELAY: Duration = Duration::from_secs(120);
...
-    pending_transaction_records: LruCache<NotarizedTransactionHash, PendingTransactionRecord>,
+    pending_transaction_records: LruCache<
+        NotarizedTransactionHash,
+        (
+            PendingTransactionRecord,
+            TransactionIntentHash,
+            Vec<SubintentHash>,
+        ),
+    >,
+    // INVARIANT: The `intent_lookup` and `subintent_lookup` are kept exactly in sync with
```
iamyulong, line 560: Is mempool transaction adding single-threaded? The "sync" logic doesn't look thread-safe - there is a race condition where the reverse lookup goes out-of-sync with pending transaction records.

iamyulong (approved):

core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
dhedey: Not quite - if a Cuttlefish-only transaction is evaluated against Bottlenose it will be permanently rejected at validation time, but then become acceptable in Cuttlefish. 

I think this is unlikely to cause any critical issues -- and have some ideas for tweaking this in future, as discussed in the task list document -- but it's quite a hard subject to resolve satisfactorily without a lot of work. Happy to discuss later in the week.

core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
iamyulong: You're right. The `&mut` references are obtained from `write()` on the RW Lock, not some unsafe transmutation, which prevents unsynchronized modification to the cache data structure.

PR #1017, opened 2024-11-08
Feature/implement transaction preview v2

dhedey:
## Summary

* Implements `/core/transaction/preview-v2`
* Also:
  * Updates lots of the Core API docs
  * Makes more parameters optional in the preview v1 API

## Testing

Added a few sanity tests to `TransactionPreviewV2Test.java`


core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -7880,7 +7882,223 @@ components:
...
+        logs:
+          type: boolean
+          description: | 
+            This flag controls whether the preview response will include execution logs.
+            If not provided, this defaults to `false` and no logs will be provided in the response.
+    PreviewTransactionType:
+      type: string
+      enum:
+        - Compiled
+    PreviewTransaction:
```
PawelPawelec-RDX, line 7983: as this is used only in v2 maybe we should name It PreviewTransactionV2 ?

dhedey: Well it's intended not just to be V2 transactions. e.g. it could include PreviewTransactionV3

iamyulong (approved):
LGTM

PR #1018, opened 2024-11-08
Feature/mesh api block

lrubasze:
## Summary

- Implement `block` endpoints
- Fixed balances returned
- Some cleanup
- Configured end index for mesh-cli-test to assure test will finish within reasonable time


.github/workflows/ci.yml
```diff
@@ -251,11 +251,10 @@ jobs:
           restore-keys: ${{ runner.os }}-gradle
       - name: Install mesh-cli
         run: curl -sSfL https://raw.githubusercontent.com/coinbase/mesh-cli/master/scripts/install.sh | sh -s
-      - name: Generate keygen
-        run: ./gradlew keygen:run --args='-k ../core/node.ks -p ""'
-      - name: Build Node
-        run: ./gradlew build
       - name: Run Node in the background
+        env:
```
lrubasze, line 255: This also fixes the issue observed earlier
```
Command Failed: unable to initialize asserter for fetcher: /network/status not attempting retry: /network/status response is invalid: current block timestamp 0 is invalid: timestamp is before 01/01/2000
```
It seems like a node was not processing any transactions.

core-rust/mesh-api-server/src/mesh_api/conversions/block.rs
```diff
@@ -13,8 +13,49 @@ pub(crate) fn to_mesh_api_block_identifier_from_state_version(
...
+) -> Result<Option<StateVersion>, ExtractionError> {
+    let state_version = if let Some(index) = block_identifier.index {
+        Some(StateVersion::of(index as u64))
+    } else if let Some(hash) = &block_identifier.hash {
+        let index = hash
+            .parse::<i64>()
+            .map_err(|_| ExtractionError::InvalidInteger {
+                message: "Error converting hash to integer".to_string(),
+            })?;
+        Some(StateVersion::of(index as u64))
```
dhedey, line 33: MINOR: If both are provided, we should probably check they agree, like in `extract_state_version_from_mesh_api_block_identifier`

core-rust/mesh-api-server/src/mesh_api/conversions/hashes.rs
```diff
@@ -1,5 +1,11 @@
 use crate::prelude::*;
 
-pub fn to_api_transaction_tree_hash(transaction_tree_hash: &TransactionTreeHash) -> String {
-    to_hex(transaction_tree_hash)
+pub fn to_api_hash_bech32m<T: IsTransactionHash>(
```
dhedey, line 3: Possibly `to_api_transaction_hash_bech32m` might be better?

core-rust/mesh-api-server/src/mesh_api/handlers/network_options.rs
```diff
@@ -20,23 +20,17 @@ pub(crate) async fn handle_network_options(
...
-            // see comment https://github.com/radixdlt/babylon-node/pull/1013#discussion_r1830848173
-            operation_types: vec![
-                "Withdraw".to_string(),
-                "Deposit".to_string(),
-                "Mint".to_string(),
-                "Burn".to_string(),
-                "LockFee".to_string(),
-            ],
+            operation_statuses: MeshApiOperationStatus::iter()
+                .map(|s| models::OperationStatus::new(s.to_string(), s.into()))
```
dhedey, line 24: MINOR: an `Into<bool>` is a little indirect for my liking - it's not clear what the bool means (success or failure?).

Perhaps there should just be an `From< MeshApiOperationStatus> for models::OperationStatus` implementation?

core-rust/mesh-api-server/src/mesh_api/types.rs
```diff
@@ -0,0 +1,38 @@
+use crate::engine_prelude::*;
+use crate::prelude::*;
+
+#[derive(Debug, Clone, EnumIter, Display, FromRepr)]
+#[repr(i64)]
+pub(crate) enum MeshApiOperationTypes {
+    Withdraw,
+    Deposit,
+    // LockFee,
```
dhedey, line 9: In the design doc, I suggested the following. These allow us to ensure that the operations in a constructed transaction are a subset of the operations in a returned transaction.

In particular `UnusedLockFee` is refunded; and a non-contingent `LockFee` and `UnusedLockFee` would always have an operation success of success for a committed transaction.
```
Withdraw - against any Global Entity
Deposit - against any Global Entity
Mint (Because mint events by themselves don't have a clear destination; I suggest a mint event is converted to an equal “mint” to the resource itself, and then an equal “withdraw” from the resource)
Burn (Similarly, a burn event is converted to an equal “burn” and then a “deposit” against the resource itself)
LockFee (We split fee burn into the “lock fee” operation from the intent / event, and the unused lock fee refund)
UnusedLockFee
```

core-rust/mesh-api-server/src/mesh_api/handlers/block.rs
```diff
@@ -0,0 +1,190 @@
...
+    database: &StateManagerDatabase<impl ReadableRocks>,
+    state_version: StateVersion,
+    transaction_identifiers: &CommittedTransactionIdentifiers,
+    requested_transaction_identifier: Option<&str>,
+) -> Result<(Vec<models::Operation>, String), ResponseError> {
+    let (operations, transaction_identifier) = match transaction_identifiers
+        .transaction_hashes
+        .as_user()
+    {
+        // In case of non-user transaction we return empty transaction,
```
dhedey, line 128: Can you expand this comment - why does the `mesh-cli` return an error?
* Does it not expect to see transactions which weren't submitted?
* Does it not understand what we return? (i.e. it's an implementation bug)
* Does some other invariant get broken?

core-rust/mesh-api-server/src/mesh_api/handlers/block.rs
```diff
@@ -0,0 +1,190 @@
...
+    }))
+}
+
+fn get_operations_and_transaction_identifier(
+    mapping_context: &MappingContext,
+    database: &StateManagerDatabase<impl ReadableRocks>,
+    state_version: StateVersion,
+    transaction_identifiers: &CommittedTransactionIdentifiers,
+    requested_transaction_identifier: Option<&str>,
+) -> Result<(Vec<models::Operation>, String), ResponseError> {
```
dhedey, line 123: Instead of returning a string, let's return a `models::TransactionIdentifier` here. That way we can have a mapping function that creates a `models::TransactionIdentifier` from a state version and act as documentation that a transaction identifier concept is a state version.

core-rust/mesh-api-server/src/mesh_api/handlers/block.rs
```diff
@@ -0,0 +1,190 @@
...
+
+            let mut index = 0_i64;
+            let mut operations = vec![];
+            for (address, balance_changes) in local_transaction_execution
+                .global_balance_summary
+                .global_balance_changes
+            {
+                if address.is_account() {
+                    for (resource_address, balance_change) in balance_changes {
+                        if let BalanceChange::Fungible(amount) = balance_change {
```
dhedey, line 171: This is okay for a first pass, but we will want to fix this in due course to use a combination of:
* Balance changes & Fee lock data
* Mint/burn events

dhedey (approved):

core-rust/mesh-api-server/src/mesh_api/conversions/block.rs
lrubasze: Good point

core-rust/mesh-api-server/src/mesh_api/handlers/block.rs
lrubasze: Makes sense

core-rust/mesh-api-server/src/mesh_api/handlers/network_options.rs
lrubasze: Good idea

core-rust/mesh-api-server/src/mesh_api/conversions/hashes.rs
lrubasze: 👍

core-rust/mesh-api-server/src/mesh_api/handlers/block.rs
lrubasze: According to the description in https://docs.cdp.coinbase.com/mesh/docs/models#blockresponse blocks must form a canonical, CONNECTED chain of blocks.
So we have two options here:

1. In case of non-user transaction we return empty transaction, to preserve block continuity.

2. [Presumably] We could return None for non-user transactions, if `parent_block_identifier` for user transaction was pointing to the previous user transaction.
This one seems really nice, but we would need some mapping from block_identifier to state version. 

I would suggest going with 1) for now.

core-rust/mesh-api-server/src/mesh_api/handlers/block.rs
lrubasze: TODO added

core-rust/mesh-api-server/src/mesh_api/handlers/network_options.rs
lrubasze: Done

core-rust/mesh-api-server/src/mesh_api/conversions/block.rs
lrubasze: Done

core-rust/mesh-api-server/src/mesh_api/handlers/block.rs
dhedey: @lrubasze  the linked documentation is about blocks not being commited. In Radix, all blocks/state versions are committed...

Are either of these two approaches viable?

1. We return system transactions -- some of these will still have operations, and they would need to be returned to ensure that resource balance changes can be reconciled.
2. If we have to filter out system transactions, we return blocks for state versions with system transactions as blocks with 0 transactions.

PR #1027, opened 2024-11-15
Feature/mesh cleanups

lrubasze:
Some cleanups and TODO addressing

### DIFF
--- a/core-rust/mesh-api-server/mesh-cli-configs/default.json
+++ b/core-rust/mesh-api-server/mesh-cli-configs/default.json
@@ -46,7 +46,7 @@
         }
     },
     "construction": {
-        "offline_url": "http://127.0.0.1:3337/mesh",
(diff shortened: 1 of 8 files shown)


core-rust/mesh-api-server/mesh-cli-configs/default.json
```diff
@@ -46,7 +46,7 @@
         }
     },
     "construction": {
-        "offline_url": "http://127.0.0.1:3337/mesh",
+        "offline_url": "http://localhost:3337/mesh",
```
lrubasze, line 49: This is to cover the test requirement:
`offline_url should be different from offline_url and not empty`

iamyulong: Cheeky :-D

iamyulong (approved):

PR #1032, opened 2024-11-20
Add cuttlefish-part2

iamyulong:
## Summary

Add `cuttlefish-part2` protocol update.

### DIFF
--- a/core-rust/core-api-server/src/core_api/generated/models/system_version.rs
+++ b/core-rust/core-api-server/src/core_api/generated/models/system_version.rs
@@ -17,6 +17,8 @@ pub enum SystemVersion {
     V1,
     #[serde(rename = "V2")]
     V2,
+    #[serde(rename = "V3")]
+    V3,
 
 }
 
@@ -25,6 +27,7 @@ impl ToString for SystemVersion {
         match self {
             Self::V1 => String::from("V1"),
             Self::V2 => String::from("V2"),
+            Self::V3 => String::from("V3"),
         }
     }
 }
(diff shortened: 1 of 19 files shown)


core-rust/core-api-server/src/core_api/generated/models/system_version.rs
```diff
@@ -17,6 +17,8 @@ pub enum SystemVersion {
     V1,
     #[serde(rename = "V2")]
     V2,
+    #[serde(rename = "V3")]
+    V3,
```
dhedey, line 21: Side note: Slightly annoyingly, we'll also need a new Gateway API build to handle this :D

core-rust/state-manager/src/protocol/protocol_configs/dumunet_protocol_config.rs
```diff
@@ -44,5 +44,8 @@ pub fn dumunet_protocol_config() -> ProtocolConfig {
                 },
             ],
         },
+        ProtocolVersionName::cuttlefish_part2() => EnactImmediatelyAfterEndOfProtocolUpdate {
```
dhedey, line 47: This needs to be `EnactAtStartOfEpochIfValidatorsReady` as well to avoid wiping dumunet (same as stokenet)

dhedey (approved):
Looks pretty good, although Dumunet needs changing, and I think you're missing updating this bit:
* Add cuttlefish-part2 to  `public static ImmutableList<String> VERSION_NAMES` in Java tests.

core-rust/core-api-server/src/core_api/generated/models/system_version.rs
dhedey: If we need it today, give me a ping and I can try to do it.

PR #1043, opened 2024-12-18
Update `TransactionPreviewResponse` in Core API schema

lrubasze:
## Summary
Deprecate and mark as not-required `instruction_resource_changes` field in `TransactionPreviewResponse`.


core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -8077,17 +8076,24 @@ components:
...
+            This receipt is primarily intended for use with the toolkit and may contain information
+            that is already available in the receipt provided in the `receipt` field of this
             response.
 
-            A typical client of this API is not expected to use this receipt. The primary clients 
-            this receipt is intended for is the Radix wallet or any client that needs to perform 
+            A typical client of this API is not expected to use this receipt. The primary clients
+            this receipt is intended for is the Radix wallet or any client that needs to perform
             execution summaries on their transactions.
         instruction_resource_changes:
```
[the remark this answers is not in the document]
dhedey: Personally I'd rather not introduce a flag if we're deprecating it, because the flag would have to default to `true` anyway for backwards compatibility, so we don't get any real use out of it, it will just be a pain because we'll have to have a deprecated flag for another protocol update after the feature is removed :)

core-rust/core-api-server/core-api-schema.yaml
```diff
@@ -8077,17 +8076,24 @@ components:
...
           type: array
+          description: |
+            This object holds changes in resource balances for all vaults within affected
```
dhedey, line 8089: It's probably worth saying it does _not_ include recalls, and should not be relied on for a comprehensive view of balance changes. Instead, the receipt balance changes should be used.

core-rust/mesh-api-server/src/mesh_api/handlers/mempool_transaction.rs
```diff
@@ -28,10 +28,10 @@ pub(crate) async fn handle_mempool_transaction(
             )),
         );
     } else {
-        payload_hashes.get(0).unwrap()
+        payload_hashes.first().unwrap()
```
dhedey, line 31: It looks like you must have run `clippy` at a different (more recent?) version compared to the code base.

This isn't a problem, but in future, can we create a separate PR for reformatting changes, if we want to do them? This avoids too much noise in the PR :)

dhedey (approved):
Nice work, thanks

PR #1044, opened 2024-12-28
Adjust transaction retry from timestamp calculation

dhedey:
## Summary

* Adjusts the logic for calculation of how long to wait before allowing a recalculation of a transaction which wants to be in the mempool (or is already in the mempool, but got kicked out).
* Refactors the mempool and cached to ensure we evict from mempool on re-evaluation

## Testing

Test was added; related code was also refactored.


core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
```diff
@@ -339,6 +338,49 @@ pub enum RetryFrom {
...
 
+impl RetryFrom {
+    fn exponential_backoff(
+        from: SystemTime,
+        base_delay: Duration,
+        exponent: f32,
+        multiplier_range: core::ops::Range<f32>,
+    ) -> Self {
+        Self::calculated_or_never(|| {
+            let multiplier: f32 = 2.0_f32.powf(exponent).clamp(multiplier_range.start, multiplier_range.end);
```
talekhinezh, line 349: maybe better to clamp exponent instead?

talekhinezh (approved):

core-rust/state-manager/src/mempool/pending_transaction_result_cache.rs
dhedey: Out of interest, why do you think it would be better?

I actually originally had a clamp of the exponent, but I changed to clamping the multiplier as it felt more declarative / easier to understand; but maybe that assumption is wrong! I'm happy to change back though if you prefer?

For what it's worth, `(2^100) = +Inf`, which clamps to `multiplier_range.end` - will add some unit tests for this though.

PR #1059, opened 2025-02-14
Add dugong protocol update

dhedey:
## Summary

Adds support for the WIP dugong protocol update, with a tag pointing at the tip of https://github.com/radixdlt/radixdlt-scrypto/pull/2068

You'll need to wipe your dev database to run it, as it attempts to run Dugong right after Cuttlefish part 2.

## Testing

A test checks that Dugong enacts with default testing config.

## Changelog

Updated.


.github/workflows/ci.yml
```diff
@@ -253,9 +253,7 @@ jobs:
           path: ~/.gradle/caches
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
           restore-keys: ${{ runner.os }}-gradle
-      - name: Build Node
-        run: ./gradlew build
```
dhedey, line 257: This was also running all standard tests, which is a waste / duplication of CI resources.

dhedey: Instead, `./gradlew :core:assemble` only runs the build and not also the tests. (Thanks gradle for being confusing like that).

LukasGasior1 (approved):

iamyulong (approved):

